PR #1, opened 2019-09-24
Assignment 1 Test.

KennethWhite:
Test first assignment, ~15 mins from fork to PR. I like the null ref you stuck in there

### DIFF
--- a/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/Program.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.IO;
+using System.Linq;
 
 namespace PrincessBrideTrivia
 {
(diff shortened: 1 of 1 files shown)


PrincessBrideTrivia/Program.cs
```diff
@@ -55,9 +57,13 @@ public static bool DisplayResult(string userGuess, Question question)
...
         {
             Console.WriteLine("Question: " + question.Text);
-            for (int i = 0; i < question.Answers.Length; i++)
+            Random r = new Random();
+            int index = 1;
+            foreach (var answer in question.Answers.OrderBy(x => r.Next()))
             {
-                Console.WriteLine((i + 1) + ": " + question.Answers[i]);
+                Console.WriteLine((index) + ": " + answer);
+                question.CorrectAnswerIndex = index.ToString();
```
Keboo: Does this actually work? I would expect this to cause the correct answer to get moved around (which is not correct). The intent was for just the display of the options to be randomized.

PrincessBrideTrivia/Program.cs
```diff
@@ -19,12 +20,13 @@ public static void Main(string[] args)
                     numberCorrect++;
                 }
             }
-            Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
-        }
+
+            Console.WriteLine("You got " + $@"{GetPercentCorrect(numberCorrect, questions.Length)}" + " correct");
```
Keboo:
```suggestion
            Console.WriteLine("You got " + $"{GetPercentCorrect(numberCorrect, questions.Length)}" + " correct");
```
If you are going to switch it to use interpolation, it would probably be best to ditch all of the concatenations too.

PrincessBrideTrivia/Program.cs
KennethWhite: If you randomize the display of the options the number next to the option no longer corresponds with the `CorrectAnswerIndex`. E.g.
1. correct
2. wrong
=> shuffle would result in two options
1. wrong
2. correct
-------
2. wrong
1. correct

The first result looks correct but typing `1` actually results in the program reporting the answer as correct, despite the text for option one no longer corresponding to the correct index.

The second option bothered me too much..

PR #2, opened 2019-09-27
Some notes from lec 1

ashermancinelli:
Assuming this can stay open and other students can add to the PR till we have more notes from today.


Keboo:
Thanks for sharing your notes! 
Unfortunately, this wont allow for everyone to edit, I migrated the markdown to [the wiki](https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/wiki/Notes) where everyone has permission to add/edit.

PR #3, opened 2019-10-01
Fix issue 1

breaman:
I think I got the first part of the assignment done.


.gitignore
```diff
@@ -354,3 +354,8 @@ healthchecksdb
 MigrationBackup/
 
 # End of https://www.gitignore.io/api/csharp
+
+# Mac specific files
+.DS_Store
+
+*.sqlite
```
Keboo: Shouldn't these changes be on master?

Keboo:
This PR is going against master. Shouldn't it be targeting the Assignment 1 branch?

.gitignore
breaman: Yes you are correct, thank you for pointing out my idiocracy in front of the whole class. :)

Keboo: Sorry, PRs are like mtn dew; when I see one, it is hard to resist.

PR #6, opened 2019-10-01
AS1: IS2

ashermancinelli:
Casted to float avoiding int division


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((float)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
breaman: Only thing I may have added here would be a Math.Round before adding the % sign since if you ended up with something like a 6/9, the "grade" would be weird.

PR #8, opened 2019-10-02
Assignment #1 Second Issue

pnotequalnp:
Forced the percent calculation to use floating point division instead of integer division.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -55,6 +55,31 @@ namespace PrincessBrideTrivia.Tests
             Assert.IsTrue(File.Exists(filePath));
         }
 
+        [TestMethod]
+        public void LoadQuestions_ArrayContainsNonNull()
+        {
+            string filePath = Path.GetRandomFileName();
+            try
+            {
+                // Arrange
+                GenerateQuestionsFile(filePath, 2);
+
+                // Act
+                Question[] questions = Program.LoadQuestions(filePath);
+
+                // Assert
+                foreach(Question question in questions)
+                {
+                    Assert.IsNotNull(question);
+                }
+            }
+            finally
+            {
+                File.Delete(filePath);
+            }
+        }
+
+
         [DataTestMethod]
         [DataRow(1, 1, "100%")]
         [DataRow(5, 10, "50%")]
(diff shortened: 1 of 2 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,6 +89,7 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question;
```
ashermancinelli: Is the solution to issue 1 supposed to be in the solution to issue 2?

pnotequalnp: Not quite sure. You couldn't run it without the fix and there's no branch here with it already done. Because it's just tests on this one and the incorrect function isn't tested, it may work. But that wouldn't apply to issues where that isn't the case.

breaman: Since the second part of the assignment just had to have the unit tests passing, technically this didn't need to be here in order for the assignment to be correct.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -55,6 +55,31 @@ public void GetFilePath_ReturnsFileThatExists()
             Assert.IsTrue(File.Exists(filePath));
         }
 
+        [TestMethod]
+        public void LoadQuestions_ArrayContainsNonNull()
```
breaman: This test didn't need to be here either. Part 1 and 2 were actually independent of each other. One thing we didn't go over in class very thoroughly was that this would have all "resolved" itself if we would have merged your first and second assignments into the original branch.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (numberCorrectAnswers * 100.0 / numberOfQuestions) + "%";
```
breaman: Only thing I would have added here would have been Math.Round before adding the %.

Also, you could have just converted one of these to a float (which is what is happening under the covers when you multiplied by 100.0)

```suggestion
            return Math.Round(numberCorrectAnswers * 100.0 / numberOfQuestions) + "%";
```

PR #9, opened 2019-10-02
Assignment #1 Third Issue

pnotequalnp:
Randomize order of questions displayed each run and add unit test to ensure that none are null still.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -89,6 +89,7 @@ namespace PrincessBrideTrivia
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question;
             }
             return questions;
         }
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.IO;
 
 namespace PrincessBrideTrivia
@@ -8,7 +9,7 @@ namespace PrincessBrideTrivia
         public static void Main(string[] args)
         {
             string filePath = GetFilePath();
-            Question[] questions = LoadQuestions(filePath);
+            Question[] questions = RandomizeAnswers(LoadQuestions(filePath));
 
             int numberCorrect = 0;
             for (int i = 0; i < questions.Length; i++)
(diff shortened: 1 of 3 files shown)


mmwoodfo:
I do not believe this bit of code here does what the project asks for? That is, it is not randomizing the order of the _answers_ for each question, but rather randomizing the order the _questions_ that are displayed.
```C#        
public static Question[] RandomizeQuestions(Question[] questions)
        {
            Random rnd = new Random();
            return questions.OrderBy(question => rnd.Next()).ToArray();
        }
```

> **Issue 3 (extra credit)** Feature request: Currently the application displays all of the answers in the same order each time. Adjust the application so the **_answers_ are displayed in a random order.** Add appropriate unit tests and ensure the app continues to function properly.

ps: I like your use of Linq's `OrderBy` method, however.

pnotequalnp:
> I do not believe this bit of code here does what the project asks for? That is, it is not randomizing the order of the _answers_ for each question, but rather randomizing the order the _questions_ that are displayed.
> 
> ```cs
> public static Question[] RandomizeQuestions(Question[] questions)
>         {
>             Random rnd = new Random();
>             return questions.OrderBy(question => rnd.Next()).ToArray();
>         }
> ```
> 
> > **Issue 3 (extra credit)** Feature request: Currently the application displays all of the answers in the same order each time. Adjust the application so the **_answers_ are displayed in a random order.** Add appropriate unit tests and ensure the app continues to function properly.
> 
> ps: I like your use of Linq's `OrderBy` method, however.

Indeed I misread answers as questions.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -66,6 +67,20 @@ public static string GetFilePath()
             return "Trivia.txt";
         }
 
+        public static Question[] RandomizeAnswers(Question[] questions)
```
mmwoodfo: Looks good!

PrincessBrideTrivia/PrincessBrideTrivia/Question.cs
```diff
@@ -5,5 +5,8 @@ public class Question
         public string Text;
         public string[] Answers;
         public string CorrectAnswerIndex;
+        public string CorrectAnswer {
```
breaman: Looks like you added a property purely for the purpose of testing which is not a good idea. If you would have used this in the program to say "Sorry, but the correct answer is..." that would have made sense, but it isn't considered good practice to add properties purely for testability.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
breaman: Nice solution for randomizing the answers.

PR #11, opened 2019-10-03
Fixes Assignment 1 issue 1

dougdoner:
Fixes NullReferenceException and adds unit test


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -6,6 +6,20 @@ namespace PrincessBrideTrivia.Tests
     [TestClass]
     public class ProgramTests
     {
+        [TestMethod]
```
breaman: Nice work, good test

PR #16, opened 2019-10-03
Assignment1 issue 2

twofingerrightclick:
Tasks completed for issue 2.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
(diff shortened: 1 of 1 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (((float)numberCorrectAnswers / (float)numberOfQuestions * 100)).ToString("n0") + "%";
```
breaman: Not bad, I would have actually done something like

```suggestion
return Math.Round((float)numberCorrectAnswers / numberOfQuestions * 100) + "%";
```

Only need to convert one to a float since it till implicitly do float math from there on out.

PR #20, opened 2019-10-03
change the variable type to double temporarily

kingofsky1995:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
breaman: Not bad, I would have done something like

```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```
this way you don't end up with some weird 66.666...% and it rounds up instead of down, which I think a lot of people prefer.

PR #22, opened 2019-10-03
Issue2 Fixed

AaronSauther:


dnisyndrum:
Casting the given integers to floats seems to work just fine and fixes the issue.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((float)numberCorrectAnswers / (float)numberOfQuestions * 100) + "%";
```
breaman: Only thing I might have done would have been something like the following so that if it isn't a whole number when done, you don't end up with any weird decimal answers.

```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #25, opened 2019-10-03
fixed integer math "error"

Cybotive:
Real fix. do not delete

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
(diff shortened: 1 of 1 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";//my fix
```
breaman: Only suggestion I have around this would have been to round the answer so if the answer ended up being a decimal, things would look more like a "real" percentage.
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #27, opened 2019-10-03
Fixed issue 2

jake-mcg:


kingofsky1995:
nice  and easy solution that I never thought of... My solution is changing it to a double variable but i guess if this program is much complicate then it would not help.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (numberCorrectAnswers * 100 / numberOfQuestions) + "%";
```
breaman: Not a bad solution, but this also leads to a truncating instead of rounding answer if for some reason the final answer is a decimal. I would have done something like

```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #28, opened 2019-10-03
Fixed issue 2

GamerBah:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -19,12 +19,13 @@ public static void Main(string[] args)
...
                 }
             }
+
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (double) numberCorrectAnswers / numberOfQuestions * 100 + "%";
```
breaman: Only thing I may have done differently would have been to apply rounding in case the percentage did not come back as a whole number.
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #30, opened 2019-10-04
Assignment1 issue#2

dbosco051394:
The tests are running successfully. Since this branch only covers issue #2, the program will still crash when trying to run program.cs, but the tests are running successfully.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,6 +24,8 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
+            //DevNote: could be integer division issue here.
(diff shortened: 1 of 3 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,9 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            //Issue 2 Resolution:
+            //Cast the int parametes to doubles to avoid integer division. Tests are completing successfully.
+            return ((1.0 * numberCorrectAnswers) / (1.0 * numberOfQuestions) * 100) + "%";
```
breaman: You could have just cast one of the numbers to a float as well, which then would have made all following operations float instead of int. The only other change I would have made would have been to add a round in case the answer was not a whole number
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #31, opened 2019-10-04
Fix issue 1

RSXEnthusiast:


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +74,18 @@ public void GetFilePath_ReturnsFileThatExists()
...
+        [TestMethod]
+        public void Question_IsNotNull() {
+
+            // Arrange
+
+            // Act
+            Question[] questions = Program.LoadQuestions(Program.GetFilePath());
+
+            // Assert
+            Assert.IsNotNull(questions[0]);
```
mmwoodfo: I think this unit test would be more valuable if you were to check all questions in the array for null values, rather than just the first object in the array simply because it seems like an improvement in the test it's self to cover all possible questions in the array.

RSXEnthusiast: You're right, laziness happened. Fixed now.

PR #32, opened 2019-10-04
Fix issue 2

RSXEnthusiast:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +25,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
breaman: Only really needed to cast one of these, but nothing wrong with doing both. Only suggestion I would make would be to round the answer at the end in case you don't get a whole number in the result.
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -55,9 +56,18 @@ public static bool DisplayResult(string userGuess, Question question)
         public static void DisplayQuestion(Question question)
         {
             Console.WriteLine("Question: " + question.Text);
+            Random rand = new Random();
+            int[] a = new int[3];
             for (int i = 0; i < question.Answers.Length; i++)
             {
-                Console.WriteLine((i + 1) + ": " + question.Answers[i]);
+                while (true) {
```
breaman: This looks like it was supposed to be part of the other PR and not this one. In the real world, I would have rejected this PR and made you remove this code since this is not really part of the problem you were supposed to be solving here.

RSXEnthusiast: Gotcha, I wasn't entirely clear on how you wanted them submitted at the time of this assignment. I do understand now. Thanks!

PR #37, opened 2019-10-04
Fix issue 3

RSXEnthusiast:
The new unit test is not 100% accurate, since it simply creates the array of questions twice and tests to see if they're equal, throwing an exception if they are. It's extremely unlikely that they'll be equal, but they could be, in which case the unit test would fail.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -1,6 +1,7 @@
 
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System.IO;
+using PrincessBrideTrivia;
 
 namespace PrincessBrideTrivia.Tests
 {
@@ -71,6 +72,18 @@ namespace PrincessBrideTrivia.Tests
 
             // Assert
             Assert.AreEqual(expectedString, percentage);
+        }
+
+        [TestMethod]
+        public void Question_IsNotNull() {
+
+            // Arrange
+
+            // Act
+            Question[] questions = Program.LoadQuestions(Program.GetFilePath());
+
+            // Assert
+            Assert.IsNotNull(questions[0]);
         }
 
 
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -86,6 +86,19 @@ namespace PrincessBrideTrivia.Tests
             Assert.IsNotNull(questions[0]);
         }
 
+        [TestMethod]
+        public void Answers_AreRandomized() {
+
+            //Arrange
+
+            // Act
+            Question[] questions = Program.LoadQuestions(Program.GetFilePath());
+            Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());
+
+            //Assert
+            Assert.AreNotEqual(questions, questions2);
+        }
+
 
         private static void GenerateQuestionsFile(string filePath, int numberOfQuestions)
         {
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -96,7 +96,7 @@ namespace PrincessBrideTrivia.Tests
             Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());
 
             //Assert
-            Assert.AreNotEqual(questions, questions2);
+                Assert.AreNotEqual(Question.ToString(questions), Question.ToString(questions2));
         }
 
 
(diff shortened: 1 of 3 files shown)


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +74,31 @@ public void GetFilePath_ReturnsFileThatExists()
...
+        public void Answers_AreRandomized() {
+
+            //Arrange
+
+            // Act
+            Question[] questions = Program.LoadQuestions(Program.GetFilePath());
+            Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());
+
+            //Assert
+            Assert.AreNotEqual(questions, questions2);
```
mmwoodfo: I could be wrong, but because objects in c# are references, I think this only returns false because they are different objects. Also, I think you should specifically check for the questions array because the question titles and answer index are allowed to be different. 
``` C#
            // Act
            Question[] questions = Program.LoadQuestions(Program.GetFilePath());
            Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());

            //Assert
            Assert.AreNotEqual(questions, questions2);
```

RSXEnthusiast: Pretty sure you're right about the pointer thing, so I fixed that. What do you think I should check for in the questions array?

mmwoodfo: So it's returning true because questions and questions2 are different obejcts, if you were to get rid of your randomization function in program.cs and put the original function there you would still have this test case returning true. It looks like you've already fixed the issue, but I would suggest comparing using either 
```C# 
CollectionAssert.AreNotEqual(questions[i].Answers, questions2[i].Answers)
```
or doing what you did and create a ToString to compare the two answer arrays. The text for each question you compare should be the same during comparison, but their correct answer indexes may be different since they are randomized. 

You'd probably want to check that the answer index is correct in a different unit test, so in this unit test, I would specifically check that Answers array within the question object is different from the answers array in the question2 object and exclude the question text and question correct answer index from your ToString function. I haven't actually looked at your new solution, so you may be doing that already.

RSXEnthusiast: Yup, I did just make a toString() method, and compare the resulting strings from the arrays, to make sure they're different. For some reason I thought there was already a unit test to check that the answers were correct, I was wrong, so I just added one. Thanks for the advice/ideas!

PrincessBrideTrivia/PrincessBrideTrivia/Question.cs
```diff
@@ -1,9 +1,29 @@
...
+namespace PrincessBrideTrivia
 {
     public class Question
     {
         public string Text;
         public string[] Answers;
         public string CorrectAnswerIndex;
+
+        override
+        public string ToString() {
```
breaman: Not really ideal to put logic in your code just to support testing. This would have been better as a helper method in your test that took Question as a parameter instead.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +74,45 @@ public void GetFilePath_ReturnsFileThatExists()
...
+            // Act
+            Question[] questions = Program.LoadQuestionsNonRandom(Program.GetFilePath());
+            Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());
+
+            //Assert
+                Assert.AreNotEqual(Question.ToString(questions), Question.ToString(questions2));
+        }
+
+        [TestMethod]
+        public void CorrectAnswerIndex_IsCorrect() {
```
breaman: Nice test to make sure that the answer index got updated as well.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -72,6 +74,45 @@ public void GetFilePath_ReturnsFileThatExists()
...
+        public void Answers_AreRandomized() {
+
+            //Arrange
+
+            // Act
+            Question[] questions = Program.LoadQuestionsNonRandom(Program.GetFilePath());
+            Question[] questions2 = Program.LoadQuestions(Program.GetFilePath());
+
+            //Assert
+                Assert.AreNotEqual(Question.ToString(questions), Question.ToString(questions2));
```
breaman: Random is hard to test, I agree. I would say what might have been better here would have been just to make sure that at some point in time (like maybe randomize the questions 10 times) that at least once in that point in time question 1 of non-random was not the same as question 1 of random. That would at least prove they were "moving around" in order.

PrincessBrideTrivia/PrincessBrideTrivia/Question.cs
RSXEnthusiast: That makes a lot of sense. I'll make sure to do this in the future, this was my first foray into unit testing.

PR #38, opened 2019-10-04
Assignment1 issue 3

twofingerrightclick:
Issue 3 with some basic tests showing function.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -10,8 +10,7 @@ namespace PrincessBrideTrivia
         {
             string filePath = GetFilePath();
             //Question[] questions = LoadQuestions(filePath);
-            Question[] questions = LoadQuestions(filePath);
-            questions = RandomAnswerOrder(questions);
+            Question [] questions = RandomAnswerOrder(filePath);
 
             int numberCorrect = 0;
             for (int i = 0; i < questions.Length; i++)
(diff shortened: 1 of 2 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -92,5 +94,65 @@ public static Question[] LoadQuestions(string filePath)
...
+
+                for (int x = 0; x < numberofAnswersPerQuestion; x++)
+                {
+
+                    int newIndexForAnswer = random.Next(0, numberofAnswersPerQuestion);
+
+                    string answer = lines[lineIndex + x + 1];
+
+
+                    while (usedAnswerIndexes[newIndexForAnswer] == true)
```
twofingerrightclick: not efficient for large number of answers. A linked list of answers that haven't been placed would be more efficient. After the answer was placed remove it from the list, and then randomly select from the remaining number answers without having to check if an answer had been placed as done in this method.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,8 +91,68 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question;
             }
             return questions;
         }
+
+        public static Question[] LoadQuestionsRandomAnswerOrder(string filePath)
```
mmwoodfo: I think this function may benefit from some code refactoring to become two separate functions rather than one large function in the spirit of modularity.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,8 +91,68 @@ public static Question[] LoadQuestions(string filePath)
...
+
+            Question[] questions = new Question[lines.Length / 5];
+            for (int i = 0; i < questions.Length; i++)
+            {
+                Question question = new Question();
+
+                int lineIndex = i * 5;
+                string questionText = lines[lineIndex];
+                question.Text = questionText;
+                question.Answers = new string[3];
```
mmwoodfo: You also have this variable: 
``` C#
int numberofAnswersPerQuestion = 3;
```

But you do not utilize it on this line: 
``` C#
question.Answers = new string[3];
```

And only use it in your loop
``` C#
for (int x = 0; x < numberofAnswersPerQuestion; x++)
```

I point this out because it would be easy to accidentally initilize your array to a size of 2 instead of 3 then get an exception within your for loop.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
twofingerrightclick: Alright, I'll try that!

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
twofingerrightclick: Thank you! good catch!

PR #39, opened 2019-10-04
Fixed issue 2

AustinHall7798:
Casted numberOfQuestions as a double so that the division operator would produce a decimal. Ignore the updated package reference, somehow I updated it while fixing issue 2.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -89,6 +89,7 @@ namespace PrincessBrideTrivia
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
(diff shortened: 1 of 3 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (numberCorrectAnswers / (double) numberOfQuestions * 100) + "%";
```
breaman: Only thing I might change here would be to perform a rounding operation just in case the number didn't come out as a whole number after the division
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #41, opened 2019-10-05
Assignment1 - Issue 2 : mmwoodfo

mmwoodfo:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,9 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return Math.Floor((float)numberCorrectAnswers / (float)numberOfQuestions * 100) + "%";
```
breaman: Nothing wrong with this solution, but why not use Math.Round instead? I know if for some reason I had a 89.5 I would want the 90% instead of the 89% ;)

mmwoodfo: Hmm, I don't think I had a good reason for flooring, just a choice I made when coding this. I think the main reason I picked it is that I haven't used it before in C# so it stood out to me, haha. I'm fully aware of Math.Round, and I agree, I would be rather upset if my professors started flooring all my quiz grades. :) I think I will change it, as the expected behavior would be to round, not floor. 

Thank you!

GamerBah:
Interesting! Doing it this way never crossed my mind! A little bit complicated for my taste, but very cool.

PR #42, opened 2019-10-05
Assignment1 - Issue 3 : mmwoodfo

mmwoodfo:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,8 +89,33 @@ public static Question[] LoadQuestions(string filePath)
...
             }
             return questions;
         }
+
+        /*Bonus - Using Knuth shuffle*/
+        public static Question RandomizeQuestionAnswersOrder(Question question)
+        {
+            string correctAnswer = question.Answers[Convert.ToInt32(question.CorrectAnswerIndex) - 1];
+
+            Random rand = new Random();
```
Fervidusletum: This was a gotcha for me too.  Declaring a new random each time you call the randomize method can lead to some undesired behavior. Since the default behavior of new random is to use time for the seed, quick repeated calls to this function can result in the same seed and getting the same answer ordering for a few calls in a row.  One way to get around this is to only make one random and pass it as an arg to this method.

mmwoodfo: Ah! Thank you, I didn't think that would be an issue unless it was inside the for loop it's self, but I did notice a few bugs where the questions would not always become randomized. Couldn't figure it out.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -27,6 +28,62 @@ public void LoadQuestions_RetrievesQuestionsFromFile()
...
+        {
+            //Arrange
+            Question question = GenerateAQuestion();
+            Question question2 = GenerateAQuestion();
+
+            //Act
+            question = Program.RandomizeQuestionAnswersOrder(question);
+
+            //Assert
+            CollectionAssert.AreNotEqual(question.Answers, question2.Answers);
```
breaman: Not a bad solution, but just an fyi on how CollectionAssert works. On objects, this would have just done a reference equality, so if you had something like

```
List<Person> people1 = new List<Person>();
List<Person> people2 = new List<Person>();
people1.Add(new Person {FirstName = "Inigo", LastName = "Montoya"});
people2.Add(new Person {FirstName = "Inigo", LastName = "Montoya"});

CollectionAssert.AreNotEqual(people1, people2); // This would return true since they are 2 different object references in each collection even though the contents would be the same.
```

PR #46, opened 2019-10-05
Issue 2

Fervidusletum:
Fixed method causing unit test GetPercentCorrect_ReturnsExpectedPercentage to fail.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -24,7 +24,7 @@ namespace PrincessBrideTrivia
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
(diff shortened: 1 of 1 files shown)


crackerJackTech:
I didn't think of a float type cast for the return value. I changed the incoming parameters as a double. Great work.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((float) numberCorrectAnswers / (float) numberOfQuestions * 100) + "%";
```
breaman: nice answer, but i would make the following change just in case the number wasn't a whole number after the multiplication
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #48, opened 2019-10-05
Fixed first issue.

samrnixon:


dnisyndrum:
Simple fix, looks like it should work just fine.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +86,17 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
                 File.AppendAllLines(filePath, lines);
             }
         }
+
+        [TestMethod]
+        private void LoadQuestions_CheckIfNotNull()
+        {
+            string fPath = Program.GetFilePath();
+            Question[] q = Program.LoadQuestions(fPath);
```
breaman: I know it is more typing, but naming this "questions" would be clearer in the long run since q could have been question or questions if the code was longer. Nice test, though.

PR #50, opened 2019-10-05
Assignment 1 issue 2

dougdoner:
Fixes error so all tests pass.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -89,6 +89,8 @@ namespace PrincessBrideTrivia
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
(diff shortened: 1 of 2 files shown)


markdietzler:
Heh, your fix is a lot simpler than mine.

I HATE int division.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((numberCorrectAnswers / (double)numberOfQuestions) * 100) + "%";
```
ashermancinelli: Note that casting to double instead of float will take more memory

dougdoner: oh okay thanks for the heads up.

PR #51, opened 2019-10-05
Fixed second issue.

samrnixon:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return ((double)numberCorrectAnswers / (double)numberOfQuestions * 100) + "%";
```
JerettLatimer: I also did the type cast to double. To me that was the easiest/fastest way of solving the issue. Great job!

PR #53, opened 2019-10-06
Hw1 issue1

dsergio:
Fixed Issue 1


matHopg:
Correct fix and unit test looks good.

PR #54, opened 2019-10-06
Hw1 issue2

dsergio:


matHopg:
Nice solution with Math.round different solution to mine but still works.

PR #55, opened 2019-10-06
Hw1 issue3

dsergio:


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,6 +24,18 @@ public static void Main(string[] args)
...
+            Random r = new Random();
+            for (int i = 0; i < questions.Length - 1; i++)
+            {
+                int randomInt = r.Next(i + 1, questions.Length);
+                Question temp = questions[i];
+                questions[i] = questions[randomInt];
+                questions[randomInt] = temp;
+            }
+        }
+
```
pnotequalnp: I made this mistake too. The prompt asks to randomize the order of the _answers_ for each question, not the order that the questions are presented.

PR #56, opened 2019-10-06
Fix issue 3 with unit test

AustinHall7798:


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +86,66 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+                        count++;
+                    }
+                }
+                Assert.IsFalse(count == questions1.Length);
+
+
+            }
+            finally
+            {
+                File.Delete(filePath);
```
CDobbins1: This finally statement is not needed and can cause your tester to fail since you are not creating a new file, but instead using the same file that Program is.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -86,5 +86,66 @@ private static void GenerateQuestionsFile(string filePath, int numberOfQuestions
...
+                for (int i = 0; i < questions1.Length; i++)
+                {
+                    Program.RandomizeAnswers(questions1[i]);
+                }
+
+                // Assert 
+                int count = 0;
+                for (int i = 0; i < questions1.Length; i++)
+                {
+                    if ((questions2[i].Equals(questions1[i])))
```
CDobbins1: I believe this is not comparing the values within each question, but rather checking if the question objects are the same; resulting in a false return even if the objects have the same values.

AustinHall7798: You're right, thanks for pointing that out

PR #58, opened 2019-10-06
Changed arguments passed into the GetPercentCorrect function from int…

dnisyndrum:
Changed integers to floats for the two arguments passed into the GetPerfectCorrect function to prevent integer division.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
@@ -22,7 +22,7 @@ namespace PrincessBrideTrivia
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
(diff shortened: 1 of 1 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,7 +22,7 @@ public static void Main(string[] args)
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
-        public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+        public static string GetPercentCorrect(float numberCorrectAnswers, float numberOfQuestions)
```
breaman: Interesting solution to just change the parameter type. Did not think of this one. One change I would have made, though, would be to add a Math.Round in here in case after the division you still had a number that wasn't whole.
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #59, opened 2019-10-06
Fixed Issue one 

TatyanaHubbard:
Fixed by adding the question into the questions array to have no null questions.


Tpirtle:
Good, the code does the required task

breaman:
Nice work

PR #63, opened 2019-10-07
Issue 1

Tpirtle:
Issue 1 done


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -27,6 +27,32 @@ public void LoadQuestions_RetrievesQuestionsFromFile()
             }
         }
 
+        //[TestMethod]
```
JerettLatimer: Good job. This is very similar to what I did when writing my test method.

breaman: Why did you change this from [TestMethod] to [DataTestMethod]? You aren't doing any data driven tests here.

PR #64, opened 2019-10-07
Issue 2

Tpirtle:
Issue 2 complete


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,7 +22,7 @@ public static void Main(string[] args)
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
-        public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+        public static string GetPercentCorrect(double numberCorrectAnswers, double numberOfQuestions)
```
JerettLatimer: I approached this issues slightly different as I just performed a type cast to double in the return statement. However, your solution is likely more "proper." Good job!

PR #65, opened 2019-10-07
Issue 3

Tpirtle:
Was able to make the array of questions random, could not figure out how to ensure that it would be random for unit testing.
Also fixed a commit that was not supposed to be in there previously.


PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -27,6 +28,53 @@ public void LoadQuestions_RetrievesQuestionsFromFile()
...
+            {
+                // Arrange
+                GenerateQuestionsFileRandomTest(filePath, 15);
+
+                // Act
+                Question[] questionsNotRandom = Program.LoadQuestions(filePath);
+                Question[] questionRandom = Program.QuestionsRandomizer(questionsNotRandom);
+
+                // Assert 
+                Assert.AreNotEqual(questionsNotRandom, questionRandom);
```
breaman: This is not a good way to test for random order here since this will just be doing a reference equality check and each array is pointing to a different location in memory. Random is not easy to test for since even random can sometimes have the same result and you don't want to fail if that happens either. Usually when I test random, I will make sure to create about 5 instances of my random call (usually in a for loop) and then compare elements and using a boolean field, I make sure that at least one time through the loop of 5, something was in a different order than the original.

PR #69, opened 2019-10-07
Fixed Issue 1

matHopg:
Questions were null as in for loop they were not being put into array.


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -89,6 +88,7 @@ public static Question[] LoadQuestions(string filePath)
                 question.Answers[1] = answer2;
                 question.Answers[2] = answer3;
                 question.CorrectAnswerIndex = correctAnswerIndex;
+                questions[i] = question;
```
breaman: Where is the associated unit test? This fixes the problem, but should have had a unit test to make sure things didn't break again.

PR #70, opened 2019-10-07
Issue was in GetPercentCorrect, it was dividing with two ints and not…

matHopg:
… casting to a double. Once casted to a double the test was able to pass succesfully.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -66,7 +66,7 @@ namespace PrincessBrideTrivia.Tests
             // Arrange
 
             // Act
-            string percentage = Program.GetPercentCorrect(numberOfCorrectGuesses, numberOfQuestions);
+            string percentage = Program.GetPercentCorrect(numberOfCorrectGuesses, numberOfQuestions) + "";
 
             // Assert
             Assert.AreEqual(expectedString, percentage);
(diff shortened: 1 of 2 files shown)


PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,11 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            double percent = (double)numberCorrectAnswers / (double)numberOfQuestions;
+            percent = percent * 100;
+            string percentage = percent + "%";
+            //return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return percentage;
```
breaman: Not a bad solution here. I would have simplified it and also added a Math.Round in case the result still wasn't a whole number
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -66,7 +66,7 @@ public void GetFilePath_ReturnsFileThatExists()
             // Arrange
 
             // Act
-            string percentage = Program.GetPercentCorrect(numberOfCorrectGuesses, numberOfQuestions);
+            string percentage = Program.GetPercentCorrect(numberOfCorrectGuesses, numberOfQuestions) + "";
```
breaman: Could have just called .ToString() instead of + "" as well.

PR #75, opened 2019-10-08
Fixing second issue

kenny2892:


whereisdaniru:
I was curious about your use of Convert.ToDouble vs just casting with (double), which has now sent me on a StackOverflow crawl ...   I think it's valid, but what was your reasoning for using ToDouble?

kenny2892:
I went down the Google rabbit hole and saw that, from my understanding, casting was more like telling the compiler to treat the value as a double, whereas Convert would actually make it a double. So I went with that option.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -24,7 +24,7 @@ public static void Main(string[] args)
 
         public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
         {
-            return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
+            return (Convert.ToDouble(numberCorrectAnswers) / Convert.ToDouble(numberOfQuestions) * 100) + "%";
```
breaman: Could have just done a cast as well, but running a convert isn't bad either. Only change I would have made would have been to call Math.Round in case the result wasn't a whole number after doing the division
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #79, opened 2019-10-08
Assignment1 Issue2

crackerJackTech:
Fixed Issue2.

### DIFF
--- a/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
+++ b/PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
@@ -60,8 +60,8 @@ namespace PrincessBrideTrivia.Tests
         [DataRow(5, 10, "50%")]
         [DataRow(1, 10, "10%")]
         [DataRow(0, 10, "0%")]
-        public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuesses, 
-            int numberOfQuestions, string expectedString)
+        public void GetPercentCorrect_ReturnsExpectedPercentage(double numberOfCorrectGuesses, 
+            double numberOfQuestions, string expectedString)
         {
             // Arrange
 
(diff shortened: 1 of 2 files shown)


kenny2892:
I converted the ints into doubles at the return line, but this is much simpler. Nice job.

PrincessBrideTrivia/PrincessBrideTrivia.Tests/ProgramTests.cs
```diff
@@ -60,8 +60,8 @@ public void GetFilePath_ReturnsFileThatExists()
         [DataRow(5, 10, "50%")]
         [DataRow(1, 10, "10%")]
         [DataRow(0, 10, "0%")]
-        public void GetPercentCorrect_ReturnsExpectedPercentage(int numberOfCorrectGuesses, 
-            int numberOfQuestions, string expectedString)
+        public void GetPercentCorrect_ReturnsExpectedPercentage(double numberOfCorrectGuesses, 
+            double numberOfQuestions, string expectedString)
```
breaman: Would not need to change your input parameters here, should have just converted when calling your method instead.

PrincessBrideTrivia/PrincessBrideTrivia/Program.cs
```diff
@@ -22,7 +22,7 @@ public static void Main(string[] args)
             Console.WriteLine("You got " + GetPercentCorrect(numberCorrect, questions.Length) + " correct");
         }
 
-        public static string GetPercentCorrect(int numberCorrectAnswers, int numberOfQuestions)
+        public static string GetPercentCorrect(double numberCorrectAnswers, double numberOfQuestions)
         {
             return (numberCorrectAnswers / numberOfQuestions * 100) + "%";
```
breaman: Not a bad solution, but would have also added a Math.Round in case I still had a decimal after my division.
```suggestion
return Math.Round((float)numberOfCorrectAnswers / numberOfQuestions * 100) + “%”
```

PR #81, opened 2019-10-08
AS2: P1

ashermancinelli:


Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
...
+
+namespace Logger
+{
+    public class FileLogger: BaseLogger
+    {
+        private string _LoggerName; // Is this the correct naming convention?
+        private string fileName;
+        private StreamWriter writer;
+
+        public string LoggerName { get; set; }
```
mmwoodfo: Is there a reason you put this here instead of in the BaseLogger class?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger: BaseLogger
+    {
+        private string _LoggerName; // Is this the correct naming convention?
```
mmwoodfo: Also, for auto-properties, you shouldn't need this line. 
References:
Pg **265 & 266** of the book
[Microsoft: Auto-Implemented Properties](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/auto-implemented-properties)

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,76 @@
...
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void LogMessage_DidCreateFile_Success() {
+            string fileName = "TestFile";
+            FileLogger fileLogger = new FileLogger(fileName) { LoggerName = "TestLogger" };
+            fileLogger.Log(LogLevel.Warning, "Test Message");
+            Assert.IsTrue(File.Exists($"{fileName}.log"));
+            File.Delete($"{fileName}.log");
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,68 @@
...
+                this BaseLogger baseLogger,
+                string message,
+                params object[] args) {
+            if (baseLogger is null) {
+                throw new ArgumentNullException(nameof(baseLogger));
+            }
+            else {
+                baseLogger.Log(
+                    LogLevel.Error,
+                    String.Format(message, args)
```

```suggestion
                    string.Format(message, args)
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,37 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger: BaseLogger
+    {
+        private string fileName;
```
```suggestion
        private string _FileName;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,23 @@
 {
     public class LogFactory
     {
+        private string filePath { get; set; }
```
Keboo: Properties should be PascalCase 
```suggestion
        private string FilePath { get; set; }
```

PR #84, opened 2019-10-09
Assignment 2 - Logger

dnisyndrum:
Competed program and tests.

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,28 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using System.Linq;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_CreateFile_Qapla()
+        {
+            string path = "testPath.txt";
+            _ = new FileLogger(path);
+            Assert.IsNotNull(path);
+        }
+
+        [TestMethod]
+        public void FileLogger_LogToFile_Qapla()
+        {
+            string path = Path.GetTempFileName();
+            BaseLogger baseLogger = new FileLogger(path);
+            baseLogger.Log(LogLevel.Debug, "This is a test");
+            string line = File.ReadLines(path).First();
+            Assert.IsTrue(line.Contains("This is a test"));
+        }
     }
 }
--- a/Logger.Tests/LogFactoryTests.cs
+++ b/Logger.Tests/LogFactoryTests.cs
@@ -6,6 +6,25 @@ namespace Logger.Tests
     [TestClass]
     public class LogFactoryTests
     {
-        
+        [TestMethod]
+        public void LogFactory_WithNullFilePath_ReturnNull()
+        {
+            string className = "myLogger";
+            LogFactory logFactory = new LogFactory();
+            BaseLogger logger = logFactory.CreateLogger(className);
+            Assert.IsNull(logger);
+        }
+
+        [TestMethod]
+        public void LogFactory_WithConfiguredFilePath_ReturnFileLoggerObject()
+        {
+            string className = "myLogger";
+            LogFactory logFactory = new LogFactory();
+            string path = Path.GetTempFileName();
+            logFactory.ConfigureFileLogger(path);
+            BaseLogger logger = logFactory.CreateLogger(className);
+            Assert.IsNotNull(logger);
+        }
     }
 }
+
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -2,6 +2,8 @@
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; } 
+
         public abstract void Log(LogLevel logLevel, string message);
     }
 }
--- a/Logger/BaseLoggerMixins.cs
+++ b/Logger/BaseLoggerMixins.cs
@@ -1,7 +1,44 @@
 ﻿namespace Logger
 {
-    public static class BaseLoggerMixins
+    public static class BaseLoggerMixins 
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params string[] args)
+        { 
+            if(baseLogger is null)
+            {
+                throw new System.ArgumentNullException();
+            }
+            LogLevel logLevelError = LogLevel.Error;
+            baseLogger.Log(logLevelError, string.Format(message, args));
+        }
+
+        public static void Warning(this BaseLogger baseLogger, string message, params string[] args)
+        {
+            if (baseLogger is null)
+            {
+                throw new System.ArgumentNullException();
+            }
+            LogLevel logLevelWarning = LogLevel.Warning;
+            baseLogger.Log(logLevelWarning, string.Format(message, args));
+        }
+        public static void Information(this BaseLogger baseLogger, string message, params string[] args)
+        {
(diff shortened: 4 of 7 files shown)


Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+        {
+            if(!File.Exists(filePath))
+            {
+                sw = File.CreateText(filePath);
+            }   
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            DateTime now = DateTime.Now;
```
mmwoodfo: I think we are supposed to be writing one log per line, so if we had multiple logs our output would like as follows: 

> 10/7/2019 9:38:00 PM FileLoggerTests Deubg: Test message
> 10/8/2019 1:38:59 AM FileLoggerTests Warning: Test message
> 10/10/2019 10:38:509 AM FileLoggerTests Error: Test message

Also, a mistake I made, in the case of an exception I don't think your stream writer will close properly. So you should make sure to close it. There are a few ways to do it, I wrapped mine in a using statement as the using statement should automatically flush and dispose of the object.
``` C#
using (StreamWriter sw = new StreamWriter(filePath))
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,28 @@
...
+using System.Linq;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_CreateFile_Qapla()
```
Keboo: Had to look up `Qapla`very nice.
This test does not appear to match its name. The test really only guarantees that the FileLogger constructor does not throw an exception, and that the string is not null (which it will never be).

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,28 @@
...
+
+        [TestMethod]
+        public void FileLogger_LogToFile_Qapla()
+        {
+            string path = Path.GetTempFileName();
+            BaseLogger baseLogger = new FileLogger(path);
+            baseLogger.Log(LogLevel.Debug, "This is a test");
+            string line = File.ReadLines(path).First();
+            Assert.IsTrue(line.Contains("This is a test"));
+        }
```
Keboo: Given that one of the requirements was to ensure that the the file was appended to, it would have been nice to see a test that checks for that.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,25 @@ namespace Logger.Tests
...
+        public void LogFactory_WithNullFilePath_ReturnNull()
+        {
+            string className = "myLogger";
+            LogFactory logFactory = new LogFactory();
+            BaseLogger logger = logFactory.CreateLogger(className);
+            Assert.IsNull(logger);
+        }
+
+        [TestMethod]
+        public void LogFactory_WithConfiguredFilePath_ReturnFileLoggerObject()
```
Keboo: A minor thing here, but you can usually omit words like "Object" from your method names.

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,8 @@
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; }
```
Keboo: Properties should be named with PascalCase.
```suggestion
        public string ClassName { get; set; } 
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,44 @@
...
 {
-    public static class BaseLoggerMixins
+    public static class BaseLoggerMixins 
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params string[] args)
+        { 
+            if(baseLogger is null)
+            {
+                throw new System.ArgumentNullException();
```
Keboo: For ArgumentException (and the derived ArgumentNullException) you want to call the overload that passes in the parameter name.
```suggestion
                throw new System.ArgumentNullException(nameof(baseLogger));
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,26 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
Keboo: Fields should be named with underscore PascalCase.
```suggestion
        private string _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,26 @@
...
         public BaseLogger CreateLogger(string className)
         {
-            
-            return null;
+            if(this.filePath is null)
```
```suggestion
            if(filePath is null)
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,26 @@
...
+                { 
+                    className = className 
+                };
+                return newBaseLogger;
+            }           
+        }
+
+        public void ConfigureFileLogger(string newPath)
+        {
+            this.filePath = newPath;
```
```suggestion
            filePath = newPath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+            if (!File.Exists(filePath))
+            {
+                File.CreateText(filePath);
+            }
+            
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = new StreamWriter(path))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
+        public FileLogger(string filePath) 
+        {
+            this.path = filePath;
+            if (!File.Exists(filePath))
```
Keboo: This work is not needed since the Log method will create the file if it doesn't exist. You only need to store the file path here.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
+        public FileLogger(string filePath) 
+        {
+            this.path = filePath;
```
```suggestion
            path = filePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+            if (!File.Exists(filePath))
+            {
+                File.CreateText(filePath);
```
Keboo: This is actually a problematic call since this method returns a StreamWriter. As you used below, StreamReaders should be put in a using statement so they are properly cleaned up.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
+        public FileLogger(string filePath)
```
Keboo: Typically you would want to null check this parameter.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+                File.CreateText(filePath);
+            }
+            
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = new StreamWriter(path))
+            {
+                sw.WriteLine($"{DateTime.Now} {className} {logLevel.ToString() + ": "} {message}");
```
Keboo: With string interpolation this can be made simpler and remove the extra space before the message
```suggestion
                sw.WriteLine($"{DateTime.Now} {className} {logLevel}: {message}");
```

PR #85, opened 2019-10-10
Assignment 2 : mmwoodfo

mmwoodfo:


LoggerProj/Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,55 @@
+﻿using System;
+
+namespace Logger
+{
+    public static class BaseLoggerMixins
+    {
+        public static void Error(string message, params string[] args)
```
ashermancinelli: The extension methods take the first parameter with the `this` keyword. For example,
```c#
public static void Error(this BaseLogger baseLogger, string message, params string[] args)
```
so that the methods may be called on a BaseLogger object like this:
```c#
BaseLogger bl = new BaseLogger();
bl.Error("message {0}", 42);
```
and the BaseLogger object will know it can use the extension methods.

ashermancinelli: Also, taking the params argument as `params object[] args` allows any datatype to be formatted in the string.

mmwoodfo: Thank you! This PR is still a work in progress, but I was a bit stuck on this part so I greatly appreciate your input. I will go ahead and try implementing that. I have never worked with extension methods, so your explanation is quite helpful!

mmwoodfo: I will also change this. Thank you!

LoggerProj/Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                File.Create(filePath).Dispose();
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            string log = $"{DateTime.Now.ToString()} {ClassName} {logLevel} {message}";
+
+            using (StreamWriter sw = new StreamWriter(filePath, true))
```
twofingerrightclick: Does StreamWriter Writeline work to append the new log entry to the bottom of the file? Like each time a new log is added?

mmwoodfo: Yes, it appends to the bottom of the file each time. So an example of what a log file using this method would look like is shown below: 

> 10/14/2019 6:57:53 PM  Information Test Message 1
> 10/14/2019 6:57:53 PM  Warning Test Message 2
> 10/14/2019 6:57:53 PM  Error Test Message 3
> 10/14/2019 6:57:53 PM  Debug Test Message 4

twofingerrightclick: I looked it up. It does. Cool.

twofingerrightclick: Cool! Thanks

LoggerProj/Logger.Tests/FileLoggerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            //Arrange
+            string filePath = Path.GetRandomFileName();
+
+            //Act
+            FileLogger fileLogger = new FileLogger(filePath);
+
+            //Assert
+            Assert.IsTrue(File.Exists(filePath));
+
+            File.Delete(filePath);
```
Keboo: Though this is good, it is important to note that this will only get cleaned up in the case where your unit test passes. The `Assert` calls throw an exception on failure and the unit test stops there. To ensure cleanup a try/finally would be better.

LoggerProj/Logger.Tests/FileLoggerTests.cs
```diff
@@ -0,0 +1,45 @@
...
+
+            //Act
+            fileLogger.Log(LogLevel.Information, "Test Message 1");
+            fileLogger.Log(LogLevel.Warning, "Test Message 2");
+            fileLogger.Log(LogLevel.Error, "Test Message 3");
+            fileLogger.Log(LogLevel.Debug, "Test Message 4");
+            int lineCount = File.ReadLines(filePath).Count();
+
+            //Assert
+            Assert.AreEqual(4, lineCount);
```
Keboo: Would have been nice to also assert something about each of those lines to ensure they were actually logged in the right order.

LoggerProj/Logger.Tests/LogFactoryTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class LogFactoryTests
+    {
+        [TestMethod]
+        public void CreateLogger_NotConfigured_ReturnNull()
+        {
+            //Arrange
+            LogFactory logFactory = new LogFactory();
+
+            //Act
+            BaseLogger nullBaseLogger = logFactory.CreateLogger("Test");
```
Keboo: Minor thing, but usually you can usually omit your expectations from the variable names.

LoggerProj/Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,63 @@
...
+        {
+            LogLevel logLevel = LogLevel.Error;
+
+            if (baseLogger is null)
+            {
+                throw new ArgumentNullException(nameof(baseLogger));
+            }
+            else
+            {
+                baseLogger.Log(logLevel, String.Format(message, args));
```
Keboo: When possible, favor using C# keywords over the class name. 
```suggestion
                baseLogger.Log(logLevel, string.Format(message, args));
```

LoggerProj/Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
Keboo: Fields should be named with underscore PascalCase
```suggestion
        private string _FilePath;
```

LoggerProj/Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
```
Keboo:
```suggestion
            filePath = filePath;
```

LoggerProj/Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+
+            if (!File.Exists(filePath))
```
Keboo: You technically do not need to do this here as the StreamWriter will create the file if it does not exist.

LoggerProj/Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                File.Create(filePath).Dispose();
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            string log = $"{DateTime.Now.ToString()} {ClassName} {logLevel} {message}";
```
Keboo: ToString call is not needed when inside of an interpolated string.
```suggestion
            string log = $"{DateTime.Now} {ClassName} {logLevel} {message}";
```

LoggerProj/Logger/LogFactory.cs
```diff
@@ -0,0 +1,29 @@
+﻿namespace Logger
+{
+    public class LogFactory
+    {
+        private string filePath;
```
Keboo:
```suggestion
        private string _FilePath;
```

LoggerProj/Logger/LogFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+                    ClassName = className
+                };
+                return baseLogger;
+            }
+            
+        }
+
+        public void ConfigureFileLogger(string filePath)
+        {
+            this.filePath = filePath;
```
Keboo:
```suggestion
            filePath = filePath;
```

PR #86, opened 2019-10-11
finished logger/working on UnitTest

kingofsky1995:


Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,47 @@
...
+using System.Linq;
+
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
+        [TestMethod]
+        public void Getpath_Success()
```
Keboo: This is a minor thing, but in general you want you unit test name to reflect the method that is being tested. In this case something more like: 
```suggestion
        public void FileLogger_Constructor_CreatesFile()
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,47 @@
...
+
+            logContent = File.ReadLines(path).ElementAt(1);
+            Assert.IsTrue(logContent.Contains("Warning test."));
+
+            logContent = File.ReadLines(path).ElementAt(2);
+            Assert.IsTrue(logContent.Contains("Information test."));
+
+            logContent = File.ReadLines(path).ElementAt(3);
+            Assert.IsTrue(logContent.Contains("Debug test."));
+        }
```
Keboo: Good test, but for each of those line it is have to re-read the entire file. It would have been better to call `File.ReadLines()` once and store its return value in a local variable.

Logger.Tests/Logger.Tests.csproj
```diff
@@ -7,6 +7,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Microsoft.Build" Version="16.3.0" />
```
Keboo: Was this addition intentional?

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,11 @@
 {
     public abstract class BaseLogger
     {
+        public string className
```
```suggestion
        public string ClassName
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,63 @@
...
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params object[] vs)
+        {
+            LogLevel logLevel = LogLevel.Error;
+
+            if (baseLogger == null)
+                throw new ArgumentNullException(baseLogger.ToString());
```
Keboo: This will actually cause a null reference exception when the `baseLogger` is null. The appropriate thing to do is pass the name of the argument that was null.
```suggestion
                throw new ArgumentNullException(nameof(baseLogger));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
+
+        public FileLogger(string path)
```
Keboo: It would have also been good to null check the `path` argument

Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
```
```suggestion
        private string _Path;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
...
+            
+            //check if the path is valid
+
+            if (File.Exists(path)){
+                Console.WriteLine("the " + path + " is already existing");
+            }
+
+            else
+            {
+                Console.WriteLine("path doesn't exist, enter your path here: ");
```
Keboo: If the path does not exist it is not necessary to prompt the user for it. This also makes unit testing this code path very difficult. This is causing some unit tests to not complete.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,40 @@
...
+
+            //update new path...
+            this.path = path;
+
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StreamWriter streamWriter = new StreamWriter(this.path);
+            streamWriter.WriteLine( DateTime.Now.ToString() + " " + className + " " + logLevel + " " + message);
```
Keboo: This doesn't quite work. This will always only have a single line in the output file since this [overload of the StreamWriter](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=netcore-3.0#System_IO_StreamWriter__ctor_System_String_) will overwrite the existing file. There is an [overload that takes in bool](https://docs.microsoft.com/en-us/dotnet/api/system.io.streamwriter.-ctor?view=netcore-3.0#System_IO_StreamWriter__ctor_System_String_System_Boolean_) that will append to the file rather than overwriting it. This is the reason for your the unit test failure with the message: "System.IO.IOException: The process cannot access the file"
String interpolation would have also made this a lot cleaner.
```suggestion
            streamWriter.WriteLine( $"{DateTime.Now} {className} {logLevel} {message}");
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,29 @@
 {
     public class LogFactory
     {
+        private string path;
```
```suggestion
        private string _Path;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,29 @@
...
+
+
+
+            BaseLogger bl = new FileLogger(path)
+            {
+                className = className
+            };
+
+
+            if (path == null)
```
Keboo: Normally you would do this check before instantiating a new FileLogger.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,63 @@
...
+
+            else {
+                message = Convert.ToString(vs);
```
Keboo: This overwrites the passed in `message` value. The appropriate thing to do here is use the `string.Format` method.
```suggestion
                string newMessage = string.Format(message, vs); 
```

PR #87, opened 2019-10-11
Began Assignment 2

TatyanaHubbard:
Started fixing the code, will finish tests and extra credit later


Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+            {
+                File.Create(filePath);
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            string messageLog = $"{DateTime.Now.ToString()} {ClassName} {logLevel}: {message}";
+
+            using (StreamWriter streamWriter = File.CreateText(filePath))
```
Keboo: This is not quite right since [File.CreateText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.createtext?view=netcore-3.0) will overwrite the file if it already exists. The objective was to append to the file. There are existing methods like [File.AppendAllText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.appendalltext?view=netcore-3.0) that would do this.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
...
+
+        public void SetFilePath(string newFilePath)
+        {
+            filePath = newFilePath;
+
+            if (!File.Exists(filePath))
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,28 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
Keboo: Field should be underscore PascalCase
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,41 @@
...
+            logFactory.ConfigureFileLogger(className);
+
+            FileLogger fileLogger = (FileLogger)logFactory.CreateLogger(className);
+            fileLogger.SetFilePath(newFile);
+
+            string fileLogTest = $"{DateTime.Now.ToString()} {className} {LogLevel.Information}: {message}";
+            fileLogger.Information(message);
+
+            Assert.AreEqual(fileLogTest, File.ReadAllText(newFile).Trim());
+        }
```
Keboo: Given that one of the objectives was to append to the file. It would have been nice to see a unit test that covered that case.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,41 @@
...
+
+        [TestMethod]
+        public void Log_FileMessageIsCorrect_isTrue()
+        {
+            string newFile = Path.GetTempFileName();
+            string className = "FileLoggerTests";
+            string message = "FileLoggerTests are running";
+
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger(className);
```
Keboo: Shouldn't this be passing the file path?
```suggestion
            logFactory.ConfigureFileLogger(newFile);
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,41 @@
...
+
+            FileLogger fileLogger = (FileLogger)logFactory.CreateLogger(className);
+            fileLogger.SetFilePath(newFile);
```
Keboo: This cast to `FileLogger` and `SetFilePath` invocation should not have been needed.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,47 @@
...
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params string[] args)
+        {
+            if(baseLogger == null)
+            {
+                throw new ArgumentNullException();
+            }
+
+            baseLogger.Log(LogLevel.Error, message);
```
Keboo: This is dropping the args without formatting them as part of the message. This is why your BaseLoggerMixinTests are failing.
```suggestion
            baseLogger.Log(LogLevel.Error, string.Format(message, args));
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,41 @@
...
+
+            string fileLogTest = $"{DateTime.Now.ToString()} {className} {LogLevel.Information}: {message}";
```

PR #88, opened 2019-10-11
Assignment2

AaronSauther:
Code is complete with unit tests! There is a minor error with github in that this pull request is not automatically updating and showing my latest commit in the "Commits" tab, but the "Files Changed" section is up to date with my latest commit. So the code in "Files Changed" is indeed the final version of my assignment2.


Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,19 @@
...
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Null_Message_ThrowsException()
+        {
+            LogFactory myFactory = new LogFactory();
+            myFactory.ConfigureFileLogger("test.txt");
+            FileLogger myLogger = (FileLogger)myFactory.CreateLogger("test");
```
Keboo: Since the `FileLogger` class is what is being tested here. You can simply directly instantiate it rather than using the `LogFactory`. The goal of unit testing is to test the smallest possible unit of code.

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,8 @@
 {
     public abstract class BaseLogger
     {
+        public string name { get; set; }
```
```suggestion
        public string Name { get; set; }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,71 @@
...
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if (baseLogger is null)
+            {
+                throw new ArgumentNullException("baseLogger is null");
```
```suggestion
                throw new ArgumentNullException(nameof(baseLogger));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,71 @@
...
+            }
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                string toBeReplaced = "{" + i + "}";
+                message = message.Replace(toBeReplaced, args[i].ToString());
+            }
+
```
Keboo: There is a much simpler way to do this using `string.Format`

```suggestion
string formattedMessage = string.Format(message, args);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,71 @@
...
+            if (baseLogger is null)
+            {
+                throw new ArgumentNullException("baseLogger is null");
+            }
+
+            for (int i = 0; i < args.Length; i++)
+            {
+                string toBeReplaced = "{" + i + "}";
+                message = message.Replace(toBeReplaced, args[i].ToString());
+            }
```
Keboo: Rather than duplicating this logic it would have been better to refactor so this logic was in a common method that all of these methods could have then used.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public String filePath { get; }
```
Keboo: Properties should be PascalCase.
You want to favor using C# keywords rather than class names.
```suggestion
        public string FilePath { get; }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        internal FileLogger(String filePath)
+        {
+            this.filePath = filePath;
```
```suggestion
            filePath = filePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if (message is null)
+            {
+                throw new ArgumentNullException("Null message");
```
```suggestion
                throw new ArgumentNullException(nameof(message));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,29 @@
...
+            }
+            StringBuilder sb = new StringBuilder();
+            sb.Append(System.DateTime.Now + " ");
+            sb.Append(this.name + " ");
+            sb.Append(logLevel.ToString() + " ");
+            sb.Append(message + Environment.NewLine);
```
Keboo: This works, but there is also a convenience method on StringBuilder that will append the new line for you.
```suggestion
            sb.AppendLine(message);
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,28 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,19 @@
...
+            myLogger.Log(LogLevel.Error, null);
+        }
     }
-}
+}
```
Keboo: This is fine, but it does leave the important part of the `FileLogger.Log` method untested.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,26 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Null_Classname_ThrowsException()
+        {
+            LogFactory myFactory = new LogFactory();
+            myFactory.ConfigureFileLogger("test.txt");
+            FileLogger myLogger = (FileLogger)myFactory.CreateLogger(null);
+        }
     }
-}
+}
```
Keboo: These are good tests, but what about a test that ensures you can get a FileLogger, and one that tests the case when className is set, but filePath is null?

PR #89, opened 2019-10-12
Assignment 2

Fervidusletum:
Assignment 2 and extra credit

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,46 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        /*
+        [TestMethod]
+        public void BaseTest()
+        {
+            try
+            {
+                // Arrange
+                // Act
+                // Assert
+            }
+        }
+        */
+
+        [TestMethod]
+        public void FileIO_OpensAndLogs()
+        {
+            FileLogger logger;
+            string logFilePath = Path.GetRandomFileName();
+            try
+            {
+                // Arrange
+                LogLevel level = LogLevel.Information;
+                string message = "test";
+
+                // Act
+                logger = new FileLogger(logFilePath);
+                logger.Log(level, message);
+
+                // Assert
+                Assert.IsFalse( string.IsNullOrEmpty( File.ReadAllText(logFilePath) ));
+            }
+            finally
+            {
+                File.Delete(logFilePath);
+            }
+        }
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -20,13 +20,36 @@ namespace Logger.Tests
         */
 
         [TestMethod]
-        public void FileIO_OpensAndLogs()
+        public void FileIO_DefaultConstructor_OpensAndLogs()
         {
-            FileLogger logger;
+            FileLogger logger = new FileLogger();
+            string logFilePath = logger.FilePath;
+            try
+            {
+                // Arrange
+                LogLevel level = LogLevel.Information;
+                string message = "test";
+
+                // Act
+                logger.Log(level, message);
+
+                // Assert
+                Assert.IsFalse(string.IsNullOrEmpty(File.ReadAllText(logFilePath)));
+            }
+            finally
+            {
+                File.Delete(logFilePath);
+            }
+        }
+
+        [TestMethod]
+        public void FileIO_CustomPathConstructor_OpensAndLogs()
+        {
             string logFilePath = Path.GetRandomFileName();
             try
             {
                 // Arrange
+                FileLogger logger;
                 LogLevel level = LogLevel.Information;
                 string message = "test";
 
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -6,19 +6,6 @@ namespace Logger.Tests
     [TestClass]
     public class FileLoggerTests
     {
-        /*
-        [TestMethod]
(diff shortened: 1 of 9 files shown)


Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,56 @@
...
+                Assert.IsFalse(string.IsNullOrEmpty(File.ReadAllText(logFilePath)));
+            }
+            finally
+            {
+                File.Delete(logFilePath);
+            }
+        }
+
+        [TestMethod]
+        public void FileIO_CustomPathConstructor_OpensAndLogs()
```
twofingerrightclick: This is a decent test. I didn't think of padding my code with try and final. This is a good practice yeah?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Text;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public string FilePath { get; private set; }
+
+        private static string GetDefaultPath() => Path.Combine(Directory.GetCurrentDirectory(), DateTime.Now.ToString("yyyy-MM-dd_HH-mm-ss") + ".txt");
```
twofingerrightclick: Good idea to use date and time to create the log file name and to place it in the current directory for easy access when testing. Note taken

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public FileLogger(string logFilePath) => FilePath = string.IsNullOrEmpty(logFilePath) ? GetDefaultPath() : logFilePath;
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StringBuilder entry = new StringBuilder();
+            entry.AppendLine(DateTime.Now.ToString("g")).Append(" ").Append(ClassName ?? "Unknown Class").Append(" ")
+                 .Append(logLevel.ToString()).Append(": ").Append(message);
+
+            StreamWriter logFile = File.AppendText(FilePath);
```
twofingerrightclick: I have seen StreamWriter often used with a "Using statement", have a look at the first answer in this stackexchange https://stackoverflow.com/questions/15120254/when-is-file-close-necessary

Logger.Tests/FileLoggerTests.cs
Fervidusletum: I figure it's more graceful to wrap code that can reasonably throw, in a try block. Especially with file IO, the finally block lets you clean up in case there is a throw.

Looking at the `using()` statement you suggested, I wonder if that may be more appropriate here since I don't need to do any specific throw handling. Seems like it would be kind of unintuitive just to use it for a shorthand try block though.

Logger/FileLogger.cs
Fervidusletum: Thanks for the tip! `using()` looks pretty handy.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,56 @@
...
+                logger.Log(level, message);
+
+                // Assert
+                Assert.IsFalse( string.IsNullOrEmpty( File.ReadAllText(logFilePath) ));
+            }
+            finally
+            {
+                File.Delete(logFilePath);
+            }
+        }
```
Keboo: Given that one of the requirements was to append to the file. It would have been nice to see a unit test cover that case.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,34 @@ namespace Logger.Tests
...
+            // Arrange
+            string targetPath = Path.GetRandomFileName();
+            LogFactory factory = new LogFactory();
+
+            // Act
+            factory.ConfigureLogger(targetPath);
+            FileLogger logger = (FileLogger)factory.CreateLogger("test class");
+
+            // Assert
+            Assert.IsTrue(string.Equals(logger.FilePath, targetPath));
```
Keboo: A better assert would have been to use `Assert.AreEqual`. This way if the assertion fails you get a nice useful error message.
```suggestion
            Assert.AreEqual(targetPath, logger.FilePath);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,26 @@
...
+
+        public static void Debug(this BaseLogger logger, string message, params object[] args)
+            => WriteLog(logger, LogLevel.Debug, message, args);
+
+        private static void WriteLog(BaseLogger logger, LogLevel level, string message, params object[] args)
+        {
+            if (logger is null) throw new ArgumentNullException(nameof(logger));
+
+            logger.Log(level, string.Format(message, args));
+        }
```
Keboo: Nice work on this. Very well done.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public FileLogger() => FilePath = GetDefaultPath();
+
+        public FileLogger(string logFilePath) => FilePath = string.IsNullOrEmpty(logFilePath) ? GetDefaultPath() : logFilePath;
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StringBuilder entry = new StringBuilder();
+            entry.AppendLine(DateTime.Now.ToString("g")).Append(" ").Append(ClassName ?? "Unknown Class").Append(" ")
+                 .Append(logLevel.ToString()).Append(": ").Append(message);
```
Keboo: This certainly works. Though a simple interpolated string might have been simpler.

Logger/LogFactory.cs
```diff
@@ -1,11 +1,16 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public class LogFactory
     {
+        private string logFilePath = null;
```
Keboo: Fields should be underscore PascalCase. In C#, there is no need to initialize variables to their default values.
```suggestion
        private string _LogFilePath = null;
```

PR #90, opened 2019-10-12
Assignment2

Cybotive:
(Out Of Date)Currently working on making sure FileLogger does its thing (I have a feeling i need to do some more string formatting). Just finished passing the provided unit tests.

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -8,7 +8,7 @@ namespace Logger.Tests
     {
 
         [TestMethod]
-        public void AppendsMessage()
+        public void Log_AppendsMessage()
         {
             //Arrange
             string path = "AppendsMessage.txt";
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -6,5 +6,16 @@ namespace Logger.Tests
     public class FileLoggerTests
     {
         
+        [TestMethod]
+        public void AppendsCorrectMessage()
+        {
+            //Arrange
+
+            //Act
+
+            //Assert
+
+        }
+
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,20 +1,29 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+
         [TestMethod]
-        public void AppendsCorrectMessage()
+        public void AppendsMessage()
         {
             //Arrange
+            string path = "AppendsMessage.txt";
+            if (File.Exists(path) is false) { File.Create(path); }
+            string[] fileLines = File.ReadAllLines(path);
+            int lineCount = fileLines.Length;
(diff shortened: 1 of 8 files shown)


Cybotive:
I believe I have finished the assignment to spec.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,33 @@ namespace Logger.Tests
     [TestClass]
     public class LogFactoryTests
     {
-        
+        [TestMethod]
+        public void CreateLogger_BeforeConfiguring_ReturnsNull()
+        {
+            //Arrange
+            LogFactory lF = new LogFactory();
```
Keboo: We usually try to give our variables longer meaningful names. See the first few entries on the [Coding Guidelines](https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/wiki/Coding-Guidelines)

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,33 @@ namespace Logger.Tests
...
+            BaseLogger bL =  new FileLogger("placeholder.txt"); //Starting with non-null to ensure the action is what made bL null
```
Keboo: The new FileLogger created here is un-used. A better approach would be to simply declare and initialize `bL` in the Act portion of the test.
```C#
BaseLogger bL = lF.CreateLogger("LogFactoryTests");
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,41 @@
...
+        private static void LogLevelMessage(LogLevel LL, BaseLogger baseLogger, string message, params int[] v)
+        {
+            if (baseLogger is null) { throw new ArgumentNullException(); }
+
+            string[] parameters = new string[v.Length];
+
+            for (int i = 0; i < v.Length; i++)
+            {
+                parameters[i] = v[i].ToString();
+            }
```
Keboo: It is not actually necessary to convert all of the objects to strings. You can simply pass the object array directly into the `string.Format` call.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,30 @@
+﻿using System;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
```
```suggestion
        private string _Path;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,30 @@
...
     public class FileLoggerTests
     {
-        
+
+        [TestMethod]
+        public void Log_AppendsMessage()
+        {
+            //Arrange
+            string path = "AppendsMessage.txt";
+            if (File.Exists(path) is false) { File.Create(path); }
```
Keboo: This test fails the first time you run it with:
```
System.IO.IOException: The process cannot access the file
```
The reason is because [File.Create](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.create?view=netcore-3.0) opens the file and returns a `FileStream` object. This stream must be closed before the logger will be able to write to the stream. It would have probably been better to simply initialize `lineCount` to zero if the file did not exist rather than to create it.

Logger/BaseLoggerMixins.cs
Cybotive: I attempted that first, but was getting nothing but addresses until I added this step.

PR #91, opened 2019-10-12
Assignment_2

samrnixon:
Pretty much done. Possibly need more mixin tests? Also BaseLoggerMixinsTests.cs line 30, not sure yet how to deal with "Message {0}" as a parameter.

### DIFF
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -14,9 +14,10 @@ namespace Logger.Tests
             // Arrange
 
             // Act
-            //BaseLoggerMixins.Error(null, "");
+            BaseLoggerMixins.Error(null, "");
 
             // Assert
+
         }
 
         [TestMethod]
@@ -27,6 +28,8 @@ namespace Logger.Tests
 
             // Act
             //logger.Error("Message {0}", 42);
+            //not sure how to get from "Message {0}" to "Message"
+            logger.Error("Message", 42);
 
             // Assert
             Assert.AreEqual(1, logger.LoggedMessages.Count);
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,31 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using System.Linq;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_LogCreateFileCorrectly()
+        {
+            string fPath = Path.GetTempFileName();
+            FileLogger fLogger = new FileLogger(fPath);
+            Assert.IsTrue(File.Exists(fPath));
+            //File.Delete(fPath);
+        }
+
+        [TestMethod]
+        public void FileLogger_LogWrittenToFileCorrectly()
+        {
+            string fPath = Path.GetTempFileName();
+            FileLogger fLogger = new FileLogger(fPath);
+            fLogger.Log(LogLevel.Warning, "Testing Method");
+
+            int lineCount = File.ReadLines(fPath).Count();
+
+            Assert.AreEqual(1, lineCount);
+        }
     }
 }
--- a/Logger.Tests/LogFactoryTests.cs
+++ b/Logger.Tests/LogFactoryTests.cs
@@ -6,6 +6,27 @@ namespace Logger.Tests
     [TestClass]
     public class LogFactoryTests
     {
-        
+        [TestMethod]
+        public void CreateLogger_PathIsNull()
+        {
+            BaseLogger baseLogger;
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger(null);
+            baseLogger = logFactory.CreateLogger("Test Path");
+
+            Assert.IsNull(baseLogger);
+        }
+
+        [TestMethod]
+        public void CreateLogger_PathIsNotNull()
+        {
+            BaseLogger baseLogger;
+            string randoPath = Path.GetTempFileName();
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger(randoPath);
+            baseLogger = logFactory.CreateLogger("Test Path");
+
+            Assert.IsNotNull(baseLogger);
+        }
     }
 }
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -2,6 +2,7 @@
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; }
         public abstract void Log(LogLevel logLevel, string message);
     }
 }
--- a/Logger/BaseLoggerMixins.cs
+++ b/Logger/BaseLoggerMixins.cs
@@ -1,7 +1,81 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger bLog, string msg, params object[] msgArgs)
+        {
+            LogLevel error = LogLevel.Error;
+
+            if(bLog is null)
+            {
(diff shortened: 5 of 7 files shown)


Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -27,6 +28,8 @@ public void Error_WithData_LogsMessage()
 
             // Act
             //logger.Error("Message {0}", 42);
+            //not sure how to get from "Message {0}" to "Message"
+            logger.Error("Message", 42);
```
mmwoodfo: So you aren't going from `"Message {0}"` to `"Message"`. You're going from `"Message {0}"` to `"Message 42"`. This is because of the `{0}`, which is a string formatted that replaces the `{0}` with the first parameter passed into the formatted string. 
In order to convert it in the BaseLoggerMixins.cs you'll need to call `String.Format(message, msgArgs)` where the message is `"Message {0}"` (or whatever string you pass in) and msgArgs is the `42 `and whatever else you pass through.

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,7 @@
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; }
```
```suggestion
        public string ClassName { get; set; }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,81 @@
...
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger bLog, string msg, params object[] msgArgs)
+        {
+            LogLevel error = LogLevel.Error;
+
+            if(bLog is null)
+            {
+                throw new ArgumentNullException("Baselogger is null");
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,81 @@
...
+            }
+            else
+            {
+                foreach(object s in msgArgs)
```
Keboo: Though this works for simple examples, it work work for messages like "Hi {0}, how are you?" since it is blindly appending. The appropriate thing to do is use `msg = string.Format(msg, msgArgs);`

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string fPath;
+        StreamWriter sWriter;
```
Keboo: Specify access modifiers on members.
```suggestion
        private StreamWriter sWriter;
```

This field is also only used inside of the Log method and could have simply be a local member rather than field.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string fPath;
```
Keboo: Fields should be named with underscore PascalCase. 
```suggestion
        private string _Path;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+        {
+            fPath = pth;
+
+            if(File.Exists(fPath))
+            {
+
+            }
+            else
+            {
+                File.Create(fPath);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            DateTime dateTime = DateTime.Now;
+            string dTime = dateTime.ToString();
```
Keboo: From the coding guidelines

> DO NOT use abbreviations or contractions within identifier names.

```suggestion
            string time = dateTime.ToString();
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+            string lLevel = logLevel.ToString();
+
+            string logInput = $"{dTime} {className} {lLevel} {message}";
```
Keboo: Variable inside of string interpolation will automatically get converted to string for you. You do not need to explicitly call `.ToString()` on them.
```suggestion
            string logInput = $"{DateTime.Now} {className} {logLevel} {message}";
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,27 @@
 {
     public class LogFactory
     {
-        public BaseLogger CreateLogger(string className)
+        private string fPath;
```
```suggestion
        private string _Path;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,27 @@
...
+
+        public void ConfigureFileLogger(string pth)
```
```suggestion
        public void ConfigureFileLogger(string path)
```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,27 @@ namespace Logger.Tests
...
     public class LogFactoryTests
     {
-        
+        [TestMethod]
+        public void CreateLogger_PathIsNull()
+        {
+            BaseLogger baseLogger;
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger(null);
+            baseLogger = logFactory.CreateLogger("Test Path");
```
Keboo: A really minor thing here. The string "Test Path" here is pretty misleading since the argument is expected to be `className`

Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,27 @@ namespace Logger.Tests
...
+        [TestMethod]
+        public void CreateLogger_PathIsNotNull()
+        {
+            BaseLogger baseLogger;
+            string randoPath = Path.GetTempFileName();
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger(randoPath);
+            baseLogger = logFactory.CreateLogger("Test Path");
+
+            Assert.IsNotNull(baseLogger);
```
Keboo: Would have also been nice to assert that the `className` passed above was set on the logger.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,41 @@
...
+
+            using(sWriter = new StreamWriter(fPath))
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,31 @@
...
+        public void FileLogger_LogWrittenToFileCorrectly()
+        {
+            string fPath = Path.GetTempFileName();
+            FileLogger fLogger = new FileLogger(fPath);
+            fLogger.Log(LogLevel.Warning, "Testing Method");
+
+            int lineCount = File.ReadLines(fPath).Count();
+
+            Assert.AreEqual(1, lineCount);
+        }
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,31 @@
...
+        [TestMethod]
+        public void FileLogger_LogWrittenToFileCorrectly()
+        {
+            string fPath = Path.GetTempFileName();
+            FileLogger fLogger = new FileLogger(fPath);
+            fLogger.Log(LogLevel.Warning, "Testing Method");
+
+            int lineCount = File.ReadLines(fPath).Count();
+
+            Assert.AreEqual(1, lineCount);
```
Keboo: In addition to asserting that there was a single line in the created file, it would have been good to assert something about contents of the line. Perhaps that it contained "Warning" and "Testing Method" would have been good.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,31 @@
...
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_LogCreateFileCorrectly()
+        {
+            string fPath = Path.GetTempFileName();
+            FileLogger fLogger = new FileLogger(fPath);
+            Assert.IsTrue(File.Exists(fPath));
+            //File.Delete(fPath);
```
Keboo: I assume this line was commented out because the test was failing because the file was in use. The problem is that `File.Create(fPath);` call in `FileLogger` opens the file and returns an opened FileStream object. This stream would need to be closed before the file could be deleted.

PR #92, opened 2019-10-13
assignment 2 Doug Doner

dougdoner:
Adds FileLogger class and associated unit tests for FileLogger. Also adds auto-property to BaseLogger class.

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,46 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_Constructor_Sets_FilePath()
+        {
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("testfile.txt");
+                //Act
+
+                //Assert
+                Assert.AreNotEqual(null, testLogger.FilePath);
+            }
+            finally
+            {
+                File.Delete("testfile.txt");
+            }
+        }
+
+        [TestMethod]
+        public void FileLogger_creates_new_file()
+        {
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("newfile.txt");
+
+                //act
+
+                //Assert
+                Assert.AreEqual(true, File.Exists("newfile.txt"));
+            }
+            finally
+            {
+                File.Delete("newfile.txt");
+            }
+        }
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,4 +1,5 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 using System.IO;
 
 namespace Logger.Tests
@@ -40,6 +41,26 @@ namespace Logger.Tests
             finally
             {
                 File.Delete("newfile.txt");
+            }
+        }
+
+        [TestMethod]
+        public void FileLogger_Log_AppendsToFile()
+        {
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("logfile.txt");
+
+                //act
+                testLogger.Log(LogLevel.Error, "teststring");
+                string logFileConents = File.ReadAllText("logfile.txt");
+                //Assert
+                Assert.AreEqual($"{DateTime.Now} FileLoggerTests {LogLevel.Error}: teststring", logFileConents);
+            }
+            finally
+            {
+                File.Delete("logfile.txt");
             }
         }
     }
--- a/Logger.Tests/LogFactoryTests.cs
+++ b/Logger.Tests/LogFactoryTests.cs
@@ -15,6 +15,7 @@ namespace Logger.Tests
             LogFactory logFactory = new LogFactory();
 
             //Act
+            logFactory.ConfigureFileLogger("testfile.txt");
             var testLogger = logFactory.CreateLogger("FileLogger");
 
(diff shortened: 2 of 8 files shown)


skozubenko1:
You can change the main tittle to Assignment 2 instead of starts Assignment 2 if you want. Other then that, code looks great!
To simplify  CreateLogger(string className), I added the following line of code:
 return new FileLogger(FileName) { ClassName = className };

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,67 @@
...
+        {
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("newfile.txt");
+
+                //act
+
+                //Assert
+                Assert.AreEqual(true, File.Exists("newfile.txt"));
```
Keboo: This works, but a slightly better check is 
```suggestion
                Assert.IsTrue(File.Exists("newfile.txt"));
```
With this form if the assertion fails you will get a prettier error message.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,67 @@
...
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("logfile.txt");
+
+                //act
+                testLogger.Log(LogLevel.Error, "teststring");
+                string logFileConents = File.ReadAllText("logfile.txt");
+                //Assert
+                Assert.AreEqual($"{DateTime.Now} FileLoggerTests {LogLevel.Error}: teststring", logFileConents);
```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,66 @@
...
 using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class LogFactoryTests
     {
-        
+        [TestMethod]
+        public void LogFactory_Constructor_Returns_Logger()
```
Keboo: In general the information you want to include in unit test method name is the method being tested, and any pre-conditions, and expectations. In this case "Constructor" is problematic since it is not the constructor that is being tested.

```suggestion
        public void CreateLogger_Returns_Logger()
```

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,66 @@
...
+            //Act
+            logFactory.ConfigureFileLogger("testfile.txt");
+            var testLogger = logFactory.CreateLogger("FileLogger");
+
+            //Assert
+            Assert.IsTrue(testLogger is BaseLogger);
+        }
+
+        [TestMethod]
+        public void LogFactory_Constructor_Returns_FileLogger()
```
Keboo: This test is basically a better version of the test above. I would suggest that this test supersedes the previous one and the previous one could be removed.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,44 @@
...
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params object[] argArray)
+        {
+            if (baseLogger is null)
+                throw new ArgumentNullException();
+
+            baseLogger.Log(LogLevel.Error, String.Format(message, argArray));
```

```suggestion
            baseLogger.Log(LogLevel.Error, string.Format(message, argArray));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,31 @@
...
+            {
+                var file = File.Create(filePath);
+                file.Close();
+            }
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StreamWriter streamWriter = File.AppendText(this.FilePath);
+
+            streamWriter.Write($"{DateTime.Now} FileLoggerTests {LogLevel.Error}: {message}");
```
Keboo: Did you mean to pass in the log level rather than hard code `LogLevel.Error`

Logger/FileLogger.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public string FilePath { get; }
+
+        public FileLogger(string filePath)
+        {
+            FilePath = filePath;
+
+            if (!File.Exists(FilePath))
+            {
+                var file = File.Create(filePath);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,31 @@
...
+                file.Close();
+            }
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StreamWriter streamWriter = File.AppendText(this.FilePath);
```
Keboo: Typically we avoid using `this.` unless it is explicitly needed. However the naming guidelines ensure that it is almost never needed.

```suggestion
            StreamWriter streamWriter = File.AppendText(FilePath);
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,31 @@
...
+            else
+            {
+                return null;
+            }
+
+        }
+
+        public void ConfigureFileLogger(string filePath)
+        {
+            this.FilePath = filePath;
```
```suggestion
            FilePath = filePath;
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,31 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public class LogFactory
     {
+        private string FilePath;
```
Keboo: Fields should be named underscore PascalCase
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,67 @@
...
+
+        [TestMethod]
+        public void FileLogger_Log_AppendsToFile()
+        {
+            try
+            {
+                //Arrange
+                FileLogger testLogger = new FileLogger("logfile.txt");
+
+                //act
```
Keboo: Since this test claims to assert that contents were appended, it would have been good to call the Log method twice to assert that items were actually appended.

PR #93, opened 2019-10-13
Assignment2

jake-mcg:
WIP - Still need to do unit tests for FileLogger

### DIFF
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -7,6 +7,53 @@ namespace Logger.Tests
     [TestClass]
     public class BaseLoggerMixinsTests
     {
+
+        [DataTestMethod]
+        [DataRow("Message {0}", 12)]
+        [DataRow("Message {0}", "12")]
+        public void Debug_WithData_LogsMessage(string message, object messageContent)
+        {
+            // Arrange
+            var logger = new TestLogger();
+
+            // Act
+            logger.Debug(message, messageContent);
+
+            // Assert
+            Assert.AreEqual(1, logger.LoggedMessages.Count);
+            Assert.AreEqual(LogLevel.Debug, logger.LoggedMessages[0].LogLevel);
+            Assert.AreEqual("Message 12", logger.LoggedMessages[0].Message);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Debug_WithNullLogger_ThrowsException()
+        {
+            // Arrange
+
+            // Act
+            BaseLoggerMixins.Debug(null, "");
+
+            // Assert
+        }
+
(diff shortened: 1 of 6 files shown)


Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -7,33 +7,123 @@ namespace Logger.Tests
     [TestClass]
     public class BaseLoggerMixinsTests
     {
+
+        [DataTestMethod]
```
Keboo: Nice use of data driven unit tests

Logger/BaseLoggerMixins.cs
```diff
@@ -2,6 +2,34 @@
 {
     public static class BaseLoggerMixins
     {
-        
+
+        public static void Warning(this BaseLogger logger, string message, params object[] args)
+        {
+            if (logger is null)
+                throw new System.ArgumentNullException();
```
Keboo: Typically when throwing an `ArgumentNullException` you want to specify the argument name.
```suggestion
                throw new System.ArgumentNullException(nameof(logger));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    class FileLogger : BaseLogger
+    {
+        private string filepath;
```
Keboo: Fields should be named underscore PascalCase
```suggestion
        private string _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,15 @@
 {
     public class LogFactory
     {
+        private string filepath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+        public FileLogger(string filepath) 
+        {
+            this.filepath = filepath;
+            if (!File.Exists(filepath))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+                File.Create(filepath);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = new StreamWriter(filepath))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                sw.WriteLine($"{DateTime.Now.ToString()} {Name} {logLevel.ToString() + ": "} {message}");
```
Keboo: `.ToString()` is not needed on elements inside of string interpolation. It will be called for you.
```suggestion
                sw.WriteLine($"{DateTime.Now} {Name} {logLevel}: {message}");
```

PR #94, opened 2019-10-13
Assignment2

NikolasGraham:


Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,47 @@
...
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger logger, string message, params object[] args)
+        {
+           if(logger is null)
+            {
+                throw new ArgumentNullException("Logger is null");
```
Fervidusletum: Small suggestion here: you can use `throw new ArgumentNullException(nameof(logger));` to automagically insert the name of the specified argument for the exception, instead of hardcoding it in.

The rest of the PR looks good!

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,32 @@
...
+        public void FileLogger_Log_WritesToFile_SeparateLinesPerLog()
+        {
+            string filePath = Path.GetTempFileName();
+            FileLogger logger = new FileLogger(filePath);
+            logger.Log(LogLevel.Error, "Testing Error");
+            logger.Log(LogLevel.Warning, "Testing Warning");
+
+            int lineCount = File.ReadLines(filePath).Count();
+
+            Assert.AreEqual(2, lineCount);
```
Keboo: This is good, but it would have been even better to actually assert that the contents of each line contained the expected strings.

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,8 @@
-﻿namespace Logger
+﻿    namespace Logger
```
Keboo: A minor thing, but why the indentation here?

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,8 @@
...
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; }
```
```suggestion
        public string ClassName { get; set; }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
...
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
+
+            if (!File.Exists(filePath))
+            {
+                File.Create(filePath);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
+
+            if (!File.Exists(filePath))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,33 @@
...
+            {
+                File.Create(filePath);
+            }
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+           DateTime date = DateTime.Now;
+            String level = logLevel.ToString();
```
Keboo: From the coding guidelines. 

Logger/LogFactory.cs
```diff
@@ -2,10 +2,21 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,32 @@
...
+using System.Linq;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_Log_CreatesFile()
```
Keboo: A minor thing here. The name of this test seems to indicate it is testing the `Log` method, while it is actually testing the constructor.

PR #95, opened 2019-10-13
Hw2 solution Assignment2

dsergio:


kenny2892:
Good idea using a DataTestMethod in BaseLoggerMixinsTests instead of making multiple copies of the method, though you could have used a switch statement instead of so many if's and else if's.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -1,22 +1,67 @@
...
+            // Arrange
+
+            // Act
+            BaseLoggerMixins.Debug(null, "");
+
+            // Assert
+        }
+
         [TestMethod]
+        [ExcludeFromCodeCoverage]
```
Keboo: Why exclude your unit tests from code coverage?

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,58 @@
...
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+
+        [TestMethod]
+        public void CreateLogger_NotConfigured_ReturnsNull()
```
Keboo: This is not a FileLogger test and could be removed since it duplicates what you tested in the LogFactoryTests

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,58 @@
...
+            // Act
+            BaseLogger fileLogger = logFactory.CreateLogger(nameof(FileLoggerTests));
+
+            // Assert
+            Assert.IsNull(fileLogger);
+
+        }
+
+        [TestMethod]
+        public void CreateLogger_Configured_ReturnsNotNull()
```
Keboo: When testing the file logger the goal is to execute as little code as possible to run the test. So no need to use the `LogFactory` at all in these unit tests.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,94 @@
...
+            else
+            {
+                string logMessage;
+                if (message is null)
+                {
+                    logMessage = null;
+                } else
+                {
+                    logMessage = string.Format(message, args);
+                }
```
Keboo: This logic is duplicated accross all of these methods. It would have been a bit cleaner to refactor this logic into a shared method that they all could utilize.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,58 @@
...
+            BaseLogger fileLogger = logFactory.CreateLogger(nameof(FileLoggerTests));
+
+            // Act
+            fileLogger.Log(logLevel, message);
+            bool fileExists = File.Exists(path);
+
+            // Assert
+            Assert.IsTrue(fileExists);
+
+        }
```
Keboo: Though this all works it would have been nice to have a unit test that asserts on the contents written to the file.

Logger.Tests/ExtraCreditAdditionalLoggerTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        {
+            // Arrange
+            LogFactory logFactory = new LogFactory();
+            BaseLogger extraCreditAdditionalLogger = logFactory.CreateExtraCreditAdditionalLogger(nameof(ExtraCreditAdditionalLoggerTests));
+
+            // Act
+            extraCreditAdditionalLogger.Log(logLevel, message);
+
+            // Assert
+            Assert.IsNotNull(extraCreditAdditionalLogger);
```
Keboo: This assert doesn't seem to match the unit test name and is a little confusing on what is actually being tested here.

Logger/ExtraCreditAddtionalLogger.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Diagnostics;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class ExtraCreditAddtionalLogger : BaseLogger
+    {
+
+        public List<(LogLevel LogLevel, string Message)> LoggedMessages { get; } = new List<(LogLevel, string)>();
```
Keboo: Clever use of a tuple here. 👍

PR #96, opened 2019-10-13
Assignment2

skozubenko1:
+ extra credit

### DIFF
--- a/Logger/BaseLogger.cs
+++ b/Logger/BaseLogger.cs
@@ -1,7 +1,23 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        public string ClassName { get; set; }
         public abstract void Log(LogLevel logLevel, string message);
+
+        protected string FormatMsg(LogLevel logLevel, string message)
+        {
+            /*
+ The current date/time
+The name of the class that created the logger
+The log level
+The message
+The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message" 
+*/
+            return $"{DateTime.Now} {ClassName} {logLevel} {message}\n";
+        }
+
     }
 }
--- a/Logger/BaseLoggerMixins.cs
+++ b/Logger/BaseLoggerMixins.cs
@@ -1,7 +1,39 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        static void CheckArgument(Object obj)
+        {
+            if (obj == null)
+                throw new ArgumentNullException();
+        }
+
+        public static void Error(this BaseLogger This, string message)
+        {
+            CheckArgument(This);
+
+            This.Log(LogLevel.Error, message);
+        }
+
+        public static void Debug(this BaseLogger This, string message)
+        {
+            CheckArgument(This);
+
+            This.Log(LogLevel.Debug, message);
+        }
+        public static void Information(this BaseLogger This, string message)
+        {
+            CheckArgument(This);
+
+            This.Log(LogLevel.Information, message);
+        }
+        public static void Warning(this BaseLogger This, string message)
+        {
+            CheckArgument(This);
+
+            This.Log(LogLevel.Warning, message);
+        }
     }
 }
--- a/Logger/ConsoleLogger.cs
+++ b/Logger/ConsoleLogger.cs
@@ -1,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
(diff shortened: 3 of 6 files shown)


Logger/BaseLogger.cs
```diff
@@ -1,7 +1,23 @@
...
+﻿using System;
+
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        public string ClassName { get; set; }
         public abstract void Log(LogLevel logLevel, string message);
+
+        protected string FormatMsg(LogLevel logLevel, string message)
```
AustinHall7798: I like the way you created a method to format the message for you whenever you need it. Did you do this because of the ConsoleLogger class, or would you have done this without it as well? Code looks good to me

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,23 @@
...
+        {
+            /*
+ The current date/time
+The name of the class that created the logger
+The log level
+The message
+The format may vary, but an example might look like this "10/7/2019 12:38:59 AM FileLoggerTests Warning: Test message" 
+*/
+            return $"{DateTime.Now} {ClassName} {logLevel} {message}\n";
```
Keboo: When dealing with code that runs on both Windows and Mac we have to deal with the subtle differences in line endings (`\r\n` on Windows and `\n` on everything else). To account for this difference you can use `Environment.NewLine` which will give you the correct one depending on which operating system your code is running on.
```suggestion
            return $"{DateTime.Now} {ClassName} {logLevel} {message}{Environment.NewLine}";
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,39 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        static void CheckArgument(Object obj)
```
Keboo: Specify access modifiers. From the coding guidelines

```suggestion
        private static void CheckArgument(object obj)
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,39 @@
...
+        {
+            if (obj == null)
+                throw new ArgumentNullException();
+        }
+
+        public static void Error(this BaseLogger This, string message)
```
Keboo: Method parameters should be camelCase. Either rename `This` to a name that does not conflict with a keyword, or escape it with `@`

```C#
        public static void Error(this BaseLogger @this, string message)
        {
            CheckArgument(@this);

            @this.Log(LogLevel.Error, message);
        }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        public string FileName { get; private set; }
```
Keboo: Auto property can be made readonly.

```suggestion
        public string FileName { get; }
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,19 @@
 {
     public class LogFactory
     {
+        private string FileName;
```
Keboo: Fields should be underscore PascalCase.
```suggestion
        private string _FileName;
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,39 @@
...
+        {
+            CheckArgument(This);
+
+            This.Log(LogLevel.Error, message);
+        }
+
+        public static void Debug(this BaseLogger This, string message)
```
Keboo: These extension methods would also suppose to have "a parameter array of arguments for the message."

Logger/ConsoleLogger.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Logger
+{
+    public class ConsoleLogger : BaseLogger
```
Keboo: Where are the unit test for this class?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
```
Keboo: Where are the unit tests for this class?

Logger/LogFactory.cs
```diff
@@ -2,10 +2,19 @@
 {
     public class LogFactory
```
Keboo: Where are the unit tests for this class?

PR #97, opened 2019-10-13
Assignment 2

JerettLatimer:


dsergio:
One thing I forgot to mention is class properties should be PascalCase, not camelCase

JerettLatimer:
> One thing I forgot to mention is class properties should be PascalCase, not camelCase

Thank you for the feedback! I fixed the issue.

dnisyndrum:
Looks solid! good work.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,59 @@
...
+        public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if(baseLogger is null)
+            {
+                throw new ArgumentNullException("baseLogger is null");
+            }
+            else
+            {
+                LogLevel logLevel = LogLevel.Error;
+                baseLogger.Log(logLevel, String.Format(message, args));
```

```suggestion
                baseLogger.Log(logLevel, string.Format(message, args));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,59 @@
...
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if(baseLogger is null)
+            {
+                throw new ArgumentNullException("baseLogger is null");
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    class FileLogger : BaseLogger
```
Keboo: Specify access modifiers. The default access modifier for a class is `internal` meaning it will not be directly accessible outside of this project (and not directly available in the unit tests).

```suggestion
    public class FileLogger : BaseLogger
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private string path;
```
Keboo: Fields should be underscore PascalCase.
```suggestion
        private string _Path;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
...
+
+        public FileLogger(string filePath)
+        {
+            path = filePath;
+
+            if (!File.Exists(path))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
...
+            {
+                File.CreateText(path);
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = new StreamWriter(path))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private string path;
+
+        public FileLogger(string filePath)
+        {
+            path = filePath;
+
+            if (!File.Exists(path))
+            {
+                File.CreateText(path);
```
Keboo: This is actually problematic since [File.CreateText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.createtext?view=netcore-3.0) returns an opened StreamWriter object. You will not be able to do things like delete the file until this stream writer is closed.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,27 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,33 @@
...
+
+                // Assert
+                string line = File.ReadAllText(path);
+                Assert.IsTrue(line.Contains("Test Text"));
+            }
+            finally
+            {
+                File.Delete(path);
+            }
+        }
```
Keboo: This is good but part of the assignment was to append to the file, so it would have been to good to have a unit test that showed that calling `.Log` multiple times appended to the file.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,33 @@
...
     {
-        
+        [TestMethod]
+        public void Log_WroteToFile_Success()
+        {
+            string path = Path.GetTempFileName();
+            try
+            {
+                // Arrange
+                LogFactory testLogFactory = new LogFactory();
```
Keboo: When unit testing you want to execute as little code as possible to test the method. Ideally making the `FileLogger` public and instantiating it directly. `LogFactory` should not be needed in the `FileLoggerTests`

PR #98, opened 2019-10-13
Assignment 2

GamerBah:
Still need to complete Mixins and some more Unit Tests. Finding TDD for this to be a little difficult.

Rider defaults to adding blank lines between brace sets. I assume this extra whitespace is okay?


GamerBah:
> One thing that caught my eye is you're testing an 'append' functionality to a constant (1) in Log_EnsureCorrectWriteToFile(string). I looked into Path.GetTempFileName() and theoretically that will cover your back since it's supposed to provide an empty file, but what if it doesn't?

Good catch! Hadn't thought about that. I guess having a fallback would be a good idea lol.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,55 @@
...
     public class FileLoggerTests
     {
-        
+
+        [TestMethod]
+        [DataRow("path_test.txt")]
+        public void GivenPath_ReturnsValidLogger(string path)
+        {
+            var logger = new FileLogger(path);
+            Assert.IsNotNull(logger);
```
Keboo: This assert is not really providing any value since it simply asserts that constructors work.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,55 @@
...
+            Assert.AreEqual(path, logger.Path);
+
+            File.Delete(path);
```
Keboo: It is worth mentioning that the `Assert` methods throw an exception if they assertion fails. So if the above assert fails, you wont actually hit this line to delete the file.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,55 @@
...
+        }
+
+        [TestMethod]
+        [DataRow("Test Message")]
+        public void Log_EnsureCorrectWriteToFile(string message)
+        {
+            string path = Path.GetTempFileName();
+            var expected = 1;
+            if (File.Exists(path))
```
Keboo: By definition [Path.GetTempFileName](https://docs.microsoft.com/en-us/dotnet/api/system.io.path.gettempfilename?view=netcore-3.0) will create and return a unique file. So it will always exist, and it will always be empty.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,55 @@
...
+            {
+                expected += File.ReadAllLines(path).Length;
+            }
+
+            var logger = new FileLogger(path);
+            logger.Log(LogLevel.Information, message);
+
+            Assert.AreEqual(expected, File.ReadAllLines(path).Length);
```
Keboo: This is good, but it would have been nice to assert on the content of the new line to make sure it contained your message.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,55 @@
...
+
+            File.Delete(path);
+        }
+
```
Keboo: Would have been nice to have a test that asserts that multiple calls to the `.Log` method appended lines to the file.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+
+    public class FileLogger : BaseLogger
+    {
+
+        public string Path { get; }
+
+        public FileLogger(string path)
+        {
+            Path = path;
+            if (!File.Exists(path))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                File.Create(path);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if (!File.Exists(Path)) return;
+            using (StreamWriter writer = File.CreateText(Path))
```
Keboo: This is not quite right since [File.CreateText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.createtext?view=netcore-3.0) will overwrite the file if it already exists. The objective was to append to the file. There are existing methods like [File.AppendAllText](https://docs.microsoft.com/en-us/dotnet/api/system.io.file.appendalltext?view=netcore-3.0) that would do this.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                writer.WriteLine($"{DateTime.Now} {ClassName} {logLevel + ":"} {message}");
```
```suggestion
                writer.WriteLine($"{DateTime.Now} {ClassName} {logLevel}: {message}");
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,27 @@
 ﻿namespace Logger
 {
+
     public class LogFactory
     {
+
+        private string _path;
```
```suggestion
        private string _Path;
```

PR #99, opened 2019-10-13
Assignment 2

AustinHall7798:
I still have to do the extra credit and may have more stuff to be sure it all works, but any feedback on what I have now would be appreciated

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -29,11 +29,21 @@ namespace Logger.Tests
             string filePath = Path.GetRandomFileName();
 
             //Act
-            FileLogger fileLogger = new FileLogger(filePath);
+            FileLogger fileLogger = new FileLogger(filePath)
+            {
+                ClassName = nameof(FileLoggerTests)
+            };
+
             fileLogger.Log(LogLevel.Information, "Random message");
+
+            string[] result = File.ReadAllLines(filePath);
 
             //Assert
             Assert.AreEqual(1, File.ReadLines(filePath).Count());
+            Assert.IsTrue(result[0].Contains("Random message"));
+            Assert.IsTrue(result[0].Contains("Information"));
+
+            File.Delete(filePath);
         }
     }
 }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,39 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using System.Linq;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+
+        [TestMethod]
+        public void FileLogger_CreateFile_Success()
+        {
+            //Arrange
+            string filePath = Path.GetRandomFileName();
+
+            FileLogger fileLogger = new FileLogger(filePath);
+
+            //Assert
+            Assert.IsTrue(File.Exists(filePath));
+
+            File.Delete(filePath);
+        }
+
+        [TestMethod]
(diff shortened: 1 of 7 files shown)


Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,38 @@
-﻿namespace Logger
+﻿using System;
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void GenericMixin(this BaseLogger baseLogger, string message, int logLevel, params object[] args)
```
JerettLatimer: I like your approach to this method. It minimizes a lot of duplicate code. Good job!

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,49 @@
...
+        {
+            //Arrange
+            string filePath = Path.GetRandomFileName();
+
+            FileLogger fileLogger = new FileLogger(filePath);
+
+            //Assert
+            Assert.IsTrue(File.Exists(filePath));
+
+            File.Delete(filePath);
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,49 @@
...
+            {
+                ClassName = nameof(FileLoggerTests)
+            };
+
+            fileLogger.Log(LogLevel.Information, "Random message");
+
+            string[] result = File.ReadAllLines(filePath);
+
+            //Assert
+            Assert.AreEqual(1, File.ReadLines(filePath).Count());
```
Keboo:
```suggestion
            Assert.AreEqual(1, result.Length);
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,26 @@
...
```
Keboo: Is there a reason the `logLevel` parameter is an `int` and not a `LogLevel`?

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
+
+            if (!File.Exists(filePath))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                File.Create(filePath).Dispose();
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter writer = new StreamWriter(filePath))
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,49 @@
...
+            Assert.IsTrue(result[0].Contains("Random message"));
+            Assert.IsTrue(result[0].Contains("Information"));
+
+            File.Delete(filePath);
+        }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            {
+                writer.Write(DateTime.Now.ToString() + " ");
+                writer.Write(ClassName + " ");
+                writer.Write(logLevel + " : ");
+                writer.Write(message);
+                writer.WriteLine();
```
Keboo: This works but string interpolation could have made all of these lines simply:

```C#
writer.WriteLine($"{DateTime.Now} {ClassName} {logLevel} : {message}");
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,23 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

PR #100, opened 2019-10-13
Assignment 2

whereisdaniru:
Assignment 2.

### DIFF
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,48 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.IO;
+using System;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+
+        [TestMethod]
+        public void FileLogger_LogFileCreated()
+        {
+            // Arrange
+            string path = "testlog.txt";
+
+            // Act
+            FileLogger testLogger = new FileLogger(path);
+
+            // Assert
+            Assert.IsTrue(File.Exists(path));
+
+            //cleanup
+            File.Delete(path);
+
+        }
+
+        [TestMethod]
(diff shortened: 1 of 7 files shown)


Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,48 @@
...
+            string path = "testlog.txt";
+
+            // Act
+            FileLogger testLogger = new FileLogger(path);
+
+            // Assert
+            Assert.IsTrue(File.Exists(path));
+
+            //cleanup
+            File.Delete(path);
```

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,15 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        public string className
```
```suggestion
        public string ClassName 
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public FileLogger(string filePath)
+        {
+            this.filePath = filePath;
+
+            if(!File.Exists(filePath))
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,28 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,48 @@
...
+            testLogger.Log(LogLevel.Error, "Testing Error");
+
+            // Assert
+            Assert.IsTrue(File.ReadAllText(path).Contains("Testing Error"));
+
+            //cleanup
+            File.Delete(path);
+
+        }
+
```

PR #101, opened 2019-10-13
Assignment 2 Logger

CDobbins1:
Still new to unit testing.

### DIFF
--- a/Logger/LogFactory.cs
+++ b/Logger/LogFactory.cs
@@ -5,16 +5,9 @@
         private string filePath;
         public BaseLogger CreateLogger(string className)
         {
-            if (filePath is null)
-            {
-                return null;
-            }
+            if (filePath is null) return null;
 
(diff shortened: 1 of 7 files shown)


Logger/LogFactory.cs
```diff
@@ -2,10 +2,15 @@
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,15 @@
...
         public BaseLogger CreateLogger(string className)
         {
-            
-            return null;
+            if (filePath is null) return null;
+
+            var logger = new FileLogger(filePath) { ClassName = className ?? "CLASSNAME NOT GIVEN" };
```
Keboo: Nice use of null coalescing operator.

PR #102, opened 2019-10-13
Fixed loggers

jacob-berger:
Added all required methods. Don't know if my PR from yesterday went through to the right location.

### DIFF
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -34,6 +34,31 @@ namespace Logger.Tests
             Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
         }
 
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Error_WithNullMessage_ThrowsException()
+        {
+            //Arrange
+            var logger = new TestLogger();
+
+            //Act
+            logger.Error(null);
+
+            //Assert
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Error_WithEmptyMessage_ThrowsException()
+        {
+            //Arrange
+            var logger = new TestLogger();
+
+            //Act
+            logger.Error("");
+
+            //Assert
+        }
     }
 
     public class TestLogger : BaseLogger
--- a/Logger.Tests/BaseLoggerMixinsTests.cs
+++ b/Logger.Tests/BaseLoggerMixinsTests.cs
@@ -14,7 +14,7 @@ namespace Logger.Tests
             // Arrange
 
             // Act
-            //BaseLoggerMixins.Error(null, "");
+            BaseLoggerMixins.Error(null, "");
 
             // Assert
         }
@@ -26,10 +26,10 @@ namespace Logger.Tests
             var logger = new TestLogger();
 
             // Act
-            //logger.Error("Message {0}", 42);
+            logger.Error("Message {0}", 42);
 
-            // Assert
-            Assert.AreEqual(1, logger.LoggedMessages.Count);
+			// Assert
+			Assert.AreEqual(1, logger.LoggedMessages.Count);
             Assert.AreEqual(LogLevel.Error, logger.LoggedMessages[0].LogLevel);
             Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
         }
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,50 @@
-﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
 
 namespace Logger.Tests
 {
-    [TestClass]
-    public class FileLoggerTests
-    {
-        
-    }
+	[TestClass]
+	public class FileLoggerTests
+	{
+		public List<(LogLevel LogLevel, string Message)> LoggedMessages { get; } = new List<(LogLevel, string)>();
+
+
+		[TestMethod]
+		[ExpectedException(typeof(ArgumentNullException))]
+		public void Error_WithNullPath_ThrowsException()
+		{
+            //Arrange
+
+            //Act
+			FileLogger logger = new FileLogger(null);
+
+            //Assert
+		}
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Error_WithEmptyPath_ThrowsException()
+        {
+            //Arrange
+
+            //Act
+            FileLogger logger = new FileLogger("");
+
+            //Assert
+        }
+
+		[TestMethod]
+		public void PathDoesNotExist()
+		{
+            //Arrange
+
+            //Act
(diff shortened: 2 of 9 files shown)


dbosco051394:
It's looking good to me! Small suggestion on the CreateLogger method in LogFactory is that there could be a need to create different kinds of loggers based on the ClassName that is passed into the method. A simple "if (className == 'FileLogger')" block would allow you to parse what the intended class name the user wants to instantiate is.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -26,14 +26,39 @@ public void Error_WithData_LogsMessage()
             var logger = new TestLogger();
 
             // Act
-            //logger.Error("Message {0}", 42);
+            logger.Error("Message {0}", 42);
 
-            // Assert
-            Assert.AreEqual(1, logger.LoggedMessages.Count);
+			// Assert
+			Assert.AreEqual(1, logger.LoggedMessages.Count);
```
Keboo: Any reason for the extra indentation here? There is a hotkey in Visual Studio Ctrl+K,D that will automatically format the code for you

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,67 @@
...
+            LogFactory logFactory = new LogFactory();
+            logFactory.ConfigureFileLogger("Testing.txt");
+
+            //Act
+            BaseLogger logger = logFactory.CreateLogger("FileLogger");
+
+            //Assert
+            Assert.IsNotNull(logger);
+            Assert.IsTrue(File.Exists("Testing.txt"));
+            File.Delete("Testing.txt");
```
Keboo: It is worth mentioning that the Assert methods throw an exception if they assertion fails. So if the above assert fails, you wont actually hit this line to delete the file.

Logger.Tests/LogFactoryTests.cs
```diff
@@ -1,11 +1,67 @@
...
+            Assert.IsFalse(File.Exists("Testing.txt"));
```
Keboo: This assert is just testing that the above File.Delete worked. When unit testing, you only need to assert that your code worked correctly. You don't need to test anything in .NET.

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,58 @@
...
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if (baseLogger is null || message is null || message == "")
```
Keboo: There is a handy helper method for checking if a string is null or empty.
```suggestion
            if (baseLogger is null || string.IsNullOrEmpty(message))
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,58 @@
...
+                LogLevel level = LogLevel.Error;
+                baseLogger.Log(level, string.Format(message, args));
+            }
+        }
+
+        public static void Warning(this BaseLogger baseLogger, string message, params object[] args)
+        {
+            if (baseLogger is null || message is null || message == "")
+            {
+                throw new ArgumentNullException(nameof(baseLogger));
```
Keboo: Typically you want separate if statements so that you throw a different ArgumentNullException with the corresponding parameter name of the parameter that was null.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+    public class FileLogger : BaseLogger
+    {
+        public string _path { get; set; }
+
+        public FileLogger(string path)
+        {
+			if (!(path is null) && path != "")
+			{
+				_path = path;
+				if (!File.Exists(path))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+			}
+			else
+			{
+				throw new ArgumentNullException(path, "Path cannot be null");
+			}
+		}
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            StreamWriter streamWriter = new StreamWriter(_path);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,35 @@
...
+            string log = $"{ DateTime.Now.ToString()} { LoggerName} { logLevel} { message}";
```
Keboo: Inside of string interpolation you do not need to call `.ToString()` it will be called automatically for you.

```suggestion
            string log = $"{ DateTime.Now} { LoggerName} { logLevel} { message}";
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,36 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public class LogFactory
     {
+        private string path;
```
```suggestion
        private string _Path;
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,36 @@
...
+
+                return logger;
+            }
+        }
+
+        public void ConfigureFileLogger(string path)
+        {
+            if (path is null || path == "")
+            {
+                throw new ArgumentNullException();
```
Keboo: When throwing an ArgumentNullException pass the name of the argument that was null.
```suggestion
                throw new ArgumentNullException(nameof(path));
```

Logger.Tests/Logger.Tests.sln
```diff
@@ -0,0 +1,17 @@
...
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{8446387E-0958-44D6-8A0F-ECB0097809CB}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{8446387E-0958-44D6-8A0F-ECB0097809CB}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{8446387E-0958-44D6-8A0F-ECB0097809CB}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{8446387E-0958-44D6-8A0F-ECB0097809CB}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+EndGlobal
```
Keboo: It looks like you created two separate solution files. Typically this occurs if you double click on the .csproj files. 
When working with multiple projects look for an already created .sln (solution file) in a parent directory. You typically only ever need a single solution file.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,50 @@
...
+		{
+            //Arrange
+
+            //Act
+            FileLogger logger = new FileLogger("ThisDoesNotExist.txt");
+
+            //Assert
+			Assert.IsTrue(File.Exists("ThisDoesNotExist.txt"));
+		}
+	}
```
Keboo: The `FileLogger.Log` method appears untested.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -26,14 +26,39 @@ public void Error_WithData_LogsMessage()
...
+        public void Error_WithEmptyMessage_ThrowsException()
+        {
+            //Arrange
+            var logger = new TestLogger();
+
+            //Act
+            logger.Error("");
+
+            //Assert
+        }
```
Keboo: It only appears that the Error method has any tests. The other methods in BaseLoggerMixins should have unit tests as well.

PR #103, opened 2019-10-13
Assignment 2 - Logger

RSXEnthusiast:
Been having issues with Visual Studio and Git, not sure if user-error or actual issues, regardless it slowed progress.

I've tried running the included tests before I add more of my own, but Visual Studios hasn't been finding the two included tests. The test window is just blank. The same happens if I switch over to the last assignment branch, and try to run the tests that were previously fully functional. It also won't run the previously fully functional Program.cs from the last assignment, it just does nothing. Visual Studio is also having trouble formatting my document - ctrl+k, ctrl+d rarely works. I think it's due to errors, but Visual Studio isn't showing me those consistently either. I've done the basics, reinstalled and repaired Visual Studio, googled my issues, reset the settings to default, but haven't been able to fix it. Any ideas?

Edit: Scratch all of that. Somehow didn't know I needed to open the sln file. Stuff is working again. It's always the smallest things...

Edit: This is no longer a WIP.

### DIFF
--- a/Lectures/2019.10.08/Lecture2019.10.08.sln
+++ b/Lectures/2019.10.08/Lecture2019.10.08.sln
@@ -1,0 +1,25 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29326.143
+MinimumVisualStudioVersion = 10.0.40219.1
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Lecture2019.10.08", "Lecture2019.10.08\Lecture2019.10.08.csproj", "{A60215FD-5D82-4D8E-BAD5-F57C6CE19870}"
+EndProject
+Global
+	GlobalSection(SolutionConfigurationPlatforms) = preSolution
+		Debug|Any CPU = Debug|Any CPU
+		Release|Any CPU = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(ProjectConfigurationPlatforms) = postSolution
+		{A60215FD-5D82-4D8E-BAD5-F57C6CE19870}.Debug|Any CPU.ActiveCfg = Debug|Any CPU
+		{A60215FD-5D82-4D8E-BAD5-F57C6CE19870}.Debug|Any CPU.Build.0 = Debug|Any CPU
+		{A60215FD-5D82-4D8E-BAD5-F57C6CE19870}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{A60215FD-5D82-4D8E-BAD5-F57C6CE19870}.Release|Any CPU.Build.0 = Release|Any CPU
+	EndGlobalSection
+	GlobalSection(SolutionProperties) = preSolution
+		HideSolutionNode = FALSE
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {9E84C982-31D7-410D-A5AA-C763372A0CD7}
+	EndGlobalSection
+EndGlobal
--- a/Logger.Tests/FileLoggerTests.cs
+++ b/Logger.Tests/FileLoggerTests.cs
@@ -1,10 +1,48 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.IO;
 
 namespace Logger.Tests
 {
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_CreatedFile()
+        {
+            //arrange
+            string userName = Environment.UserName;
+            string path = "C:/Users/" + userName + "/Documents/Visual Studio Logs/logs.txt";
+
+            //act
+            FileLogger fLogger = new FileLogger(path);
+
+            //assert
+            Assert.IsTrue(File.Exists(path));
+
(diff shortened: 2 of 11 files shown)


Lectures/2019.10.08/Lecture2019.10.08.sln
```diff
@@ -0,0 +1,25 @@
+﻿
```
Keboo: It looks like this branch was not originally based on the assignment 2 branch. Which is the reason for all of these extra files. If you have questions on creating the assignment branches feel free to message me, but it looks like you have the assignment 3 branch done correctly.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,48 @@
...
+            string path = "C:/Users/" + userName + "/Documents/Visual Studio Logs/logs.txt";
+
+            //act
+            FileLogger fLogger = new FileLogger(path);
+
+            //assert
+            Assert.IsTrue(File.Exists(path));
+
+            //cleanup
+            File.Delete(path);
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,48 @@
...
     [TestClass]
     public class FileLoggerTests
     {
-        
+        [TestMethod]
+        public void FileLogger_CreatedFile()
+        {
+            //arrange
+            string userName = Environment.UserName;
+            string path = "C:/Users/" + userName + "/Documents/Visual Studio Logs/logs.txt";
```
Keboo: This test is failing for me because "Visual Studio Logs" does not exist. Typically you would not do a fully qualified path and instead just do a file name (which is a relative path from the current program).

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,48 @@
...
+            FileLogger fLogger = new FileLogger(path);
+
+            //act
+            fLogger.Log(LogLevel.Error, "Test 1: Testing Error");
+            fLogger.Log(LogLevel.Warning, "Test 2: Testing Warning");
+            fLogger.Log(LogLevel.Information, "Test 3: Testing Information");
+            fLogger.Log(LogLevel.Debug, "Test 4: Testing Debug");
+
+            //assert
+            Assert.AreEqual(212, File.ReadAllText(path).Length);
```
Keboo: This test is failing for me. I suspect the problem is that the length of the string that is created from `DateTime.Now` is variable. I would suggest that this test should be re-written to ignore that portion of output by simply asserting that the text contains the messages you expect.

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,8 @@
-﻿namespace Logger
+namespace Logger
 {
     public abstract class BaseLogger
     {
+        public string className { get; set; }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,51 @@
...
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string s, params object[] args)
+        {
+            LogLevel logLevel = LogLevel.Error;
+
+            if (baseLogger == null)
+            {
+                throw new ArgumentNullException("baseLogger is null!");
```
```suggestion
                throw new ArgumentNullException(nameof(baseLogger));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,51 @@
...
+            }
+            baseLogger.Log(logLevel, String.Format(s, args));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
+using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string outPath;
```
Keboo: Fields should be underscore PacalCase
```suggestion
        private string _OutPath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public FileLogger(string outPath)
+        {
+            this.outPath = outPath;
+            if (!File.Exists(outPath))
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                File.Create(outPath).Close();
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string s)
+        {
+            StreamWriter sw = new StreamWriter(outPath, true);
+            sw.WriteLine($"{DateTime.Now.ToString()}{className}{logLevel}{s}");
```
Keboo: You don't need to call .ToString inside of string interpolation.
```suggestion
            sw.WriteLine($"{DateTime.Now}{className}{logLevel}{s}");
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+        public FileLogger(string outPath)
+        {
+            this.outPath = outPath;
+            if (!File.Exists(outPath))
+            {
+                File.Create(outPath).Close();
+            }
+        }
+
+        public override void Log(LogLevel logLevel, string s)
```
Keboo: When overriding methods it is best to keep the parameters named the same.
```suggestion
        public override void Log(LogLevel logLevel, string message)
```

Logger.Tests/FileLoggerTests.cs
RSXEnthusiast: Good to know, that makes a lot of sense.

Also wanted to thank you for being so thorough when grading, it's very helpful!

PR #104, opened 2019-10-14
Assignment2

dbosco051394:
~~Finished 4/5 implementations, all unit tested. Still working on the last issue with BaseLoggerMixins. I will not be attempting the extra credit on this assignment.~~

EDIT: I have implemented and unit tested all issues for Assignment 2.


Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -0,0 +1,79 @@
...
+            logger.Error("Message {0}", 42);
+
+            // Assert
+            Assert.AreEqual(1, logger.LoggedMessages.Count);
+            Assert.AreEqual(LogLevel.Error, logger.LoggedMessages[0].LogLevel);
+            Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
+        }
+
+        [TestMethod]
+        public void Success_AllExtensionMethods()
```
Keboo: In general you want unit tests only testing one thing at a time. So rather than one big test you would end up with four smaller ones. This is one time when a little code duplication is a good thing.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -0,0 +1,29 @@
...
+            int lineCount = File.ReadLines(path).Count();
+
+            //Act
+            fileLogger.Log(LogLevel.Warning, "Test Warning");
+
+            //Assert
+            Assert.IsTrue(File.Exists(path)); //Check if the file still exists
+            Assert.IsTrue(new FileInfo(path).Length > 0); //Check if the file is non-empty
+            Assert.IsTrue(File.ReadLines(path).Count() == lineCount+1); //Check to see if it appended one line exactly
+        }
```

Logger/BaseLogger.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace Logger
+{
+    public abstract class BaseLogger
+    {
+        public abstract void Log(LogLevel logLevel, string message);
+        public abstract string ClassName {get; set;}
```
Keboo: This member does not need to be abstract.

Logger/BaseLoggerMixins.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace Logger
+{
+    public static class BaseLoggerMixins
+    {
+        public static void Error(this BaseLogger baseLogger, string message, params object[] messageArgs)
+        {
+            if (baseLogger == null) throw new ArgumentNullException("BaseLogger is null");
```
```suggestion
            if (baseLogger == null) throw new ArgumentNullException(nameof(baseLogger));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+	    public FileLogger()
```
Keboo: Empty default constructors are automatically generated in C# if not constructor is explicitly specified so this can be omitted.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,27 @@
...
+	    {
+	    }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            using (StreamWriter sw = File.AppendText(FilePath))
+            {
+                sw.WriteLine(DateTime.Now + " " + ClassName + " " + logLevel + " " + message);
```
Keboo: String interpolation would make this cleaner.
```suggestion
                sw.WriteLine($"{DateTime.Now} {ClassName} {logLevel} {message}");
```

Logger/LogFactory.cs
```diff
@@ -0,0 +1,33 @@
...
+                BlankLogger blankLogger = new BlankLogger();
+                blankLogger.ClassName = className;
+                return blankLogger;
+            }
+            else if(className == "FileLogger")
+            {
+                if(FilePath != null)
+                {
+                    FileLogger fileLogger = new FileLogger();
+                    fileLogger.ClassName = className;
```
Keboo: This property was suppose to be set in an object initializer.
```C#
FileLogger fileLogger = new FileLogger
{
    ClassName = className;
}
```

PR #105, opened 2019-10-15
Assignment2

kenny2892:


Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+        public FileLogger(string filePath)
+        {
+            path = filePath;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if(!(path is null) && File.Exists(path))
+            {
+                string output = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + ClassName + " " + logLevel + ": " + message;
```
CDobbins1: Using this format, consecutive logs will result in messages being appended on the same line; rather than appending messages on a new line . Consider adding "\n" to the end of the string or a similar method.

markdietzler:
Question: Why not use a List of type string in the file logger tester instead of a string array?

I didn't use one either, too much :effort: to get it working.

Logger/FileLogger.cs
kenny2892: Thanks, I added NewLine to the output.

kenny2892:
> Why not use a List of type string in the file logger tester instead of a string array?

I honestly wasn't really thinking about it. I just needed to add some kind of text to create the temp file.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,35 @@
...
-        
+        [TestMethod]
+        public void Log_Output()
+        {
+            string filePath = Path.GetRandomFileName();
+            try
+            {
+                // Arrange
+                string[] whiteSpace = new string[] { "" };
+                File.AppendAllLines(filePath, whiteSpace);
```
AaronSauther: Nice way to simply test that data is indeed being written to a file!

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string path;
```
Keboo: Fields should be named underscore PascalCase
```suggestion
        private string _Path;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public FileLogger(string filePath)
+        {
+            path = filePath;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if(!(path is null) && File.Exists(path))
```
Keboo: Generally if you are going to check `path` for null you would do it at the point it is passed it, not when it is used. 

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+            {
+                string output = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + ClassName + " " + logLevel + ": " + message + Environment.NewLine;
```
```suggestion
                string output = $"{DateTime.Now:MM/dd/yyyy hh:mm:ss tt} {ClassName} {logLevel}: {message}{Environment.NewLine}";
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,19 @@
 {
     public class LogFactory
     {
+        private string path;
```
```suggestion
        private string _Path;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,35 @@
...
+                string output = text[text.Length - 1];
+
+                // Assert 
+                Assert.AreEqual(DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + logger.ClassName + " Error: Hello World", output);
+            }
+            finally
+            {
+                File.Delete(filePath);
+            }
+        }
```

PR #106, opened 2019-10-15
Finished solution for assignment 2

markdietzler:


GamerBah:
I'm noticing references to a `TestLogger` class in your Mixins tests, but can't find a class for it anywhere in the PR... Forgot to add them maybe?

Noticing a few of your files are missing line feeds at the end of files, too.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,39 @@
...
+                FileLogger newFileLogger = new FileLogger(testFilePath) { 
+                    ClassName = this.GetType().Name
+                };
+                newFileLogger.Error("{0}", "Inconceivable!");
+                //List<string> parsedFile = new List<string>(); //simpler to just use a string array
+
+                string[] arrayOfTextFromFile = File.ReadAllLines(testFilePath);
+                string output = arrayOfTextFromFile[arrayOfTextFromFile.Length - 1];
+                                
+                Assert.AreEqual(DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + newFileLogger.ClassName + " Error: Inconceivable!", output);
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,39 @@
...
+            }
+            finally
+            {
+                File.Delete(testFilePath);
+            }
+        }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,45 @@
...
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger newBaseLogger, string message, params object[] args)
+        {
+            if (newBaseLogger is null)
+                throw new ArgumentNullException("Baselogger is null!");
```
```suggestion
                throw new ArgumentNullException(nameof(newBaseLogger));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using System.IO;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string pathToLogFile;
```
```suggestion
        private string _PathToLogFile;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+
+        public FileLogger(string newLogFilePath)
+        {
+            pathToLogFile = newLogFilePath;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if (!(pathToLogFile is null) && File.Exists(pathToLogFile))
```
Keboo: Generally if you are going to check `pathToLogFile ` for null you would do it at the point it is passed it, not when it is used. 

Logger/FileLogger.cs
```diff
@@ -0,0 +1,24 @@
...
+            {
+                string output = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + ClassName + " " + logLevel + ": " + message;
```
```suggestion
                string output = $"{DateTime.Now:MM/dd/yyyy hh:mm:ss tt} {ClassName} {logLevel}: {message}";
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,23 @@
 {
     public class LogFactory
     {
+        private string logFilePath;
```
```suggestion
        private string _LogFilePath;
```

PR #109, opened 2019-10-15
Fixes and tests complete

matHopg:


Logger/BaseLogger.cs
```diff
@@ -2,6 +2,9 @@
 {
     public abstract class BaseLogger
     {
+        //Auto property of className which is set in logFactory
+        public string className { get; set; }
```
```suggestion
        public string ClassName { get; set; }
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+
+        public FileLogger(string fileLogPath)
+        {
+            filePath = fileLogPath;
+        }
+        public override void Log(LogLevel logLevel, string message)
+        {
+            if(!(filePath is null) && File.Exists(filePath))
```
Keboo: Generally if you are going to check `path` for null you would do it at the point it is passed it, not when it is used. 

Logger/FileLogger.cs
```diff
@@ -0,0 +1,26 @@
...
+            {
+                string appendText = DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + className
+                    + " " + logLevel + ": " + message + Environment.NewLine;
```
```suggestion
                string appendText = $"{DateTime.Now:MM/dd/yyyy hh:mm:ss tt} {className} {logLevel}: {message}{Environment.NewLine}";
```

Logger/LogFactory.cs
```diff
@@ -1,11 +1,21 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public class LogFactory
     {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,30 @@
...
+                fileLogger.Error("{0} {1}", "Hello", "World");
+                string[] text = File.ReadAllLines(filePath);
+                string output = text[text.Length - 1];
+                Assert.AreEqual(DateTime.Now.ToString("MM/dd/yyyy hh:mm:ss tt") + " " + fileLogger.className + " Error: Hello World", output);
+            }
+            finally
+            {
+                File.Delete(filePath);
+            }
+        }
```

PR #110, opened 2019-10-15
Assignment2 required tasks

twofingerrightclick:


Logger.Tests/LogFactoryTests.cs
```diff
@@ -6,6 +6,36 @@ namespace Logger.Tests
...
+
+            factory.ConfigureFileLogger(path);
+            var fileLogger = factory.CreateLogger("FileLogger");
+            
+
+            // Assert
+            Assert.IsNotNull(fileLogger);
+
+            //cleanup
+            File.Delete(path);
```
Keboo: It is worth mentioning that the Assert methods throw an exception if they assertion fails. So if the above assert fails, you wont actually hit this line to delete the file.

Logger/BaseLogger.cs
```diff
@@ -1,7 +1,60 @@
...
+        }
         public abstract void Log(LogLevel logLevel, string message);
+
     }
+
+   
+
 }
+
+namespace Logger
```
Keboo: You don't need to repeat identical namespace declaration. Both classes can be inside the namespace above (though better would be to put each in their own file).

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,47 @@
...
     {
-        
+        public static void Error(this BaseLogger t, string message, params object [] args)
+        {
+            if (t==null)
+            {
+                throw new ArgumentNullException(nameof(t));
+            }
+
+            t.Log(LogLevel.Error, String.Format("{0} {1}", message, LogLevel.Error));
```

In addition you are dropping the `args` parameter.

```suggestion
            t.Log(LogLevel.Error, string.Format(message, args));
```

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -44,5 +63,20 @@ public override void Log(LogLevel logLevel, string message)
         {
             LoggedMessages.Add((logLevel, message));
         }
+
+       
+    }
+
+    static class ExtensionMethods
```
Keboo: You should not need these. 
```C#
var logger = new TestLogger();
logger.Error("Message {0}", 42);
```
Will invoke the BaseLoggerMixins.Error method.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -44,5 +63,20 @@ public override void Log(LogLevel logLevel, string message)
         {
             LoggedMessages.Add((logLevel, message));
         }
+
+       
+    }
```
Keboo: The Warning, Information, and Debug methods should have also been tested.

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -12,26 +13,44 @@ public class BaseLoggerMixinsTests
...
 
             // Act
-            //logger.Error("Message {0}", 42);
+            logger.Error("Message {0}", 42);
 
             // Assert
             Assert.AreEqual(1, logger.LoggedMessages.Count);
             Assert.AreEqual(LogLevel.Error, logger.LoggedMessages[0].LogLevel);
-            Assert.AreEqual("Message 42", logger.LoggedMessages[0].Message);
+            Assert.AreNotEqual("Message 42", logger.LoggedMessages[0].Message);
```
Keboo: This test should have passed without needing to invert the condition. See the change in the BaseLoggerMixins about fixing the message formatting.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,67 @@
...
+
+            string tempPath= Path.GetRandomFileName();
+
+            // Act
+            FileLogger testFileLogger = new FileLogger(tempPath);
+            testFileLogger.ClassName = "FileLogger";
+            testFileLogger.Log(LogLevel.Error, "there was an error");
+            testFileLogger.Log(LogLevel.Warning, "there was a warning");
+
+            using (StreamReader r = File.OpenText(tempPath))
```
Keboo: File.ReadAllLines or File.ReadAllText would have helped here.

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,67 @@
...
+            {
+                LogToConsole(r);
+            }
+
+
+            // Assert
+            Assert.IsTrue(File.Exists(tempPath));
```
Keboo: This good but it would have been nice to assert on the contents of the file. At least to assert that the messages you logged were in there.

Logger.Tests/LogFactoryTests.cs
twofingerrightclick: That is really good to know!

Logger/BaseLoggerMixins.cs
twofingerrightclick: understood.

Logger.Tests/BaseLoggerMixinsTests.cs
twofingerrightclick: Thansk, I now am aware of the C# String formatting options for working with "{0}"

PR #111, opened 2019-10-15
Assignment 2 Logger and Unit Testing Finished.

Tpirtle:


Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            DateTime localDate = DateTime.Now;
+            var culture = new CultureInfo("en-US");
+            streamWriter = new StreamWriter(filePath);
+            string toWrite = localDate.ToString(culture) + " " + cName + " " + logLevel + ": " + message;
+
+            streamWriter.WriteLine(toWrite);
```
NikolasGraham: There's an easier to do this without having StreamWrite be a class level variable and closing it. If you used using(StreamWriter sW = new StreamWriter(filePath)){.......}, it would automatically close it when finished, and make it a local variable. Just my 2 cents.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,21 @@
...
+            if (bLogger == null)
+                return null;
+            else
+                return bLogger;
+        }
+
+        public void ConfigureFileLogger(string path)
+        {
+            filePath = path;
         }
```
NikolasGraham: Everything else looks good. There are a couple formatting things I don't agree with (personally I like closing up even one line statements like with your if/else statements, keeps everything the same and easier to edit if needed), but they are still right so it's a personal preference. Good work!

Logger.Tests/BaseLoggerMixinsTests.cs
```diff
@@ -10,28 +10,89 @@ public class BaseLoggerMixinsTests
...
+            logger.Information("Message- Hello");
+
+            // Assert
+            Assert.AreEqual(1, logger.LoggedMessages.Count);
+            Assert.AreEqual(LogLevel.Information, logger.LoggedMessages[0].LogLevel);
+            Assert.AreEqual("Message- Hello", logger.LoggedMessages[0].Message);
+        }
+
+        [TestMethod]
+        public void Debig_WithData_LogsMessage()
```
Keboo:
```suggestion
        public void Debug_WithData_LogsMessage()
```

Logger.Tests/FileLoggerTests.cs
```diff
@@ -1,10 +1,39 @@
...
+                Assert.IsTrue(toTest.Contains("Debug"));
+                Assert.IsTrue(toTest.Contains("This is a message"));
+
+            }
+            finally
+            {
+                File.Delete(filePath);
+            }
+
+        }
```

Logger/BaseLogger.cs
```diff
@@ -2,6 +2,8 @@
 {
     public abstract class BaseLogger
     {
+        public string cName { get; set; }
```
```suggestion
        public string Name { get; set; }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -2,6 +2,25 @@
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger bLogger, string message,params string[] args)
+        {
+            if (bLogger == null) { throw new System.ArgumentNullException("The BaseLogger Parameter is null"); }
```
```suggestion
            if (bLogger == null) { throw new System.ArgumentNullException(nameof(bLogger)); }
```

Logger/BaseLoggerMixins.cs
```diff
@@ -2,6 +2,25 @@
...
+            bLogger.Log(LogLevel.Error, message);
+        }
+        public static void Debug(this BaseLogger bLogger, string message, params string[] args)
```

> DO NOT use abbreviations or contractions within identifier names.

```suggestion
        public static void Debug(this BaseLogger baseLogger, string message, params string[] args)
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
```
```suggestion
        private string _FilePath;
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+        private string cName;
```
Keboo: There is already a name in the base class. This should not be needed.

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+            cName = className;
+            filePath = path;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            DateTime localDate = DateTime.Now;
+            var culture = new CultureInfo("en-US");
+            streamWriter = new StreamWriter(filePath);
+            string toWrite = localDate.ToString(culture) + " " + cName + " " + logLevel + ": " + message;
```
Keboo: String interpolation can make this simpler.
```suggestion
            string toWrite = $"{localDate.ToString(culture)} {cName} {logLevel} {message}";
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+        {
+            cName = className;
+            filePath = path;
+        }
+
+        public override void Log(LogLevel logLevel, string message)
+        {
+            DateTime localDate = DateTime.Now;
+            var culture = new CultureInfo("en-US");
+            streamWriter = new StreamWriter(filePath);
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.IO;
+using System.Text;
+
+namespace Logger
+{
+    public class FileLogger : BaseLogger
+    {
+        private string filePath;
+        private string cName;
+        private StreamWriter streamWriter;
```
Keboo: This is only used inside of the Log method and could be a local variable.

Logger/LogFactory.cs
```diff
@@ -2,10 +2,21 @@
 {
     public class LogFactory
     {
+        public string filePath;
```
```suggestion
        public string _FilePath;
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,21 @@
...
+
         public BaseLogger CreateLogger(string className)
         {
-            
-            return null;
+            BaseLogger bLogger = new FileLogger(className,filePath);
```
Keboo: The name property was supposed to be set with an object initializer.
```suggestion
            BaseLogger bLogger = new FileLogger(filePath){ Name = className };
```

PR #112, opened 2019-10-15
Assignment 2

crackerJackTech:
[WIP] Have not completed the Unit Tests or the extension methods for BaseLoggerMixins.

### DIFF
--- a/Logger/BaseLoggerMixins.cs
+++ b/Logger/BaseLoggerMixins.cs
@@ -1,7 +1,82 @@
-﻿namespace Logger
+﻿using System;
+
+namespace Logger
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string logLevel, params object[] message)
+        {
+            
+            if(baseLogger is null)
+            {
+                throw new ArgumentNullException();
+            }
+
+            LogLevel log = LogLevel.Error;
+
+            foreach(object obs in message)
+            {
+                baseLogger.Log(log, obs.ToString());
+            }
+
+
+
+            
+
+        }
+
+        public static void Warning(this BaseLogger baseLogger, string logLevel, params object[] message)
+        {
+            if (baseLogger is null)
+            {
+                throw new ArgumentNullException();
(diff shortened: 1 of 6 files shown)


Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,82 @@
...
 {
     public static class BaseLoggerMixins
     {
-        
+        public static void Error(this BaseLogger baseLogger, string logLevel, params object[] message)
+        {
+            
+            if(baseLogger is null)
+            {
+                throw new ArgumentNullException();
```
Keboo: When throwing an ArgumentNullException you want to pass the name of the argument that was null.
```suggestion
                throw new ArgumentNullException(nameof(baseLogger));
```

Logger/BaseLoggerMixins.cs
```diff
@@ -1,7 +1,82 @@
...
+            }
+
+            LogLevel log = LogLevel.Error;
+
+            foreach(object obs in message)
+            {
+                baseLogger.Log(log, obs.ToString());
+            }
```
Keboo: I think what you wanted to do here was something like this to properly format the message.

```C#
baseLogger.Log(log, string.Fomat(logLevel, message));
```

Logger/FileLogger.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+
+namespace Logger
+{
+    class FileLogger : BaseLogger
+    {   
+        public override void Log(LogLevel logLevel, string message)
+        {
+            Console.WriteLine(DateTime.Now + " " + ClassName + " " + logLevel + " " + message);
```
Keboo: String interpolation can simplify this
```suggestion
            Console.WriteLine($"{DateTime.Now} {ClassName} {logLevel} {message}");
```

Logger/LogFactory.cs
```diff
@@ -2,10 +2,20 @@
 {
     public class LogFactory
     {
+        private string FilePath;
```
```suggestion
        private string _FilePath;
```

PR #115, opened 2019-10-17
Assignment 3

Fervidusletum:
See #154 for extra credit.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtension.cs
+++ b/Inheritance/src/Inheritance/ActorExtension.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static string Speak(this Actor actor, string phrase)
+        {
+            switch(actor)
+            {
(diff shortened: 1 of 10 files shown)


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public static string Speak(this Actor actor, string phrase)
+        {
+            switch(actor)
+            {
+                case null:
+                    throw new ArgumentNullException(nameof(actor));
+                case Raj raj when raj.WomenArePresent:
+                    return raj.Mumble();
+                case Raj raj when !raj.WomenArePresent:
```
mmwoodfo: I think that the `Speak` method is supposed to simply return a string, not have a switch statement that checks what actor is saying the phrase. I believe the switch statement to check if a woman is present is taken care of in the Raj class, since it is specific to Raj's character.

mmwoodfo: Oh, just kidding. I just finished the lecture so I guess this is correct. I was doing it another way.

PR #116, opened 2019-10-17
As3

ashermancinelli:


Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,18 @@
+using System;
+
+namespace Inheritance
+{
+    public static class ActorMixins
+    {
+        public static string Speak(this Actor actor) =>
```
breaman: Nice implementation of this logic.

Inheritance/test/Inheritance.Tests/FoodTests.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    [TestClass]
+    public class FoodTests
+    {
+        [TestMethod]
+        public void FoodPrintInfo_CorrectlyFormatted()
+        {
+            string upc = "8675309",
+                   brand = "SomeBrand";
+            var food = new Food() {
```
breaman: Only thing I would have done differently here would be to have used Item instead of var as the "datatype" since using Item would have proven you were overriding the PrintInfo method correctly. Would also prevent someone from accidentally removing the inheritance which would break things at runtime.

PR #117, opened 2019-10-17
Assignment 3

pnotequalnp:
Not completed yet.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtensions.cs
+++ b/Inheritance/src/Inheritance/ActorExtensions.cs
@@ -1,0 +1,17 @@
+using System;
+
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static string Speak(this Actor actor, bool womenArePresent) => actor switch
+        {
+            Penny penny                  => penny.Greet(),
+            Sheldon sheldon              => sheldon.Fact(),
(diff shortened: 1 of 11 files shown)


pnotequalnp:
It is now finished.

pnotequalnp:
Merged work from a separate branch for part 2.

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,17 @@
+using System;
+
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static string Speak(this Actor actor, bool womenArePresent) => actor switch
```
breaman: should probably have put "womenArePresent" on the Raj class instead of here since it is really only specific to Raj himself.

Inheritance/src/Inheritance/Food.cs
```diff
@@ -6,7 +6,14 @@ namespace Inheritance
 {
     public class Food : Item
     {
-        public string Upc { get; set; }
-        public string Brand { get; set; }
+        public string? Upc { get; set; }
+        public string? Brand { get; set; }
+
+        public override string PrintInfo() => (Upc, Brand) switch
```
breaman: I like what you did here for null checks, slick.

PR #119, opened 2019-10-17
Assignment3_Inheritance

RSXEnthusiast:
~~Gotta fix PersonGetsPrinted test, 'cause it isn't running at all, as in it returns "not run."~~

Edit: resolved.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtention.cs
+++ b/Inheritance/src/Inheritance/ActorExtention.cs
@@ -1,0 +1,23 @@
+﻿using System;
+
+namespace Inheritance {
+    public static class ActorExtention {
+        public static String Speak(Actor actor) {
+            string result = "";
+            if (actor is Raj) {
+                Raj raj = (Raj) actor;
+                if (raj.womenArePresent) {
+                    result = "*Incoherent mumbles*";
+                } else {
+                    result = "I'm so nerdy and awkward, it's hilarious!";
+                }
+            } else if (actor is Sheldon) {
+                result = "I really wish the programmer knew more about Big Bang Therory, so he could better make fun of me!";
+            } else if (actor is Penny) {
+                result = "If the programmer isn't mistaken, I'm just the token hot chick.";
+            }
+            result += " *canned laughter*";
+            return result;
+        }
+    }
+}
--- a/Inheritance/src/Inheritance/ActorExtention.cs
+++ b/Inheritance/src/Inheritance/ActorExtention.cs
@@ -2,7 +2,7 @@
 
 namespace Inheritance {
     public static class ActorExtention {
-        public static String Speak(Actor actor) {
+        public static String Speak(this Actor actor) {
             string result = "";
             switch (actor) {
                 case Sheldon sheldon:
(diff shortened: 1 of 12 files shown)


Inheritance/src/Inheritance/ActorExtention.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+
+namespace Inheritance {
+    public static class ActorExtention {
+        public static String Speak(Actor actor) {
```
kingofsky1995:
```suggestion
        public static String Speak(this Actor actor) {
``` 
the extention method should be called like this?

RSXEnthusiast: I don't believe so, since I pass in an actor. I tried changing it as suggested and it made no difference. Is there a reason to that I've forgotten?

CDobbins1: With how you currently have it, Speak() is not an extension method for the Actor class, only a static method in the ActorExtention static class. For example, you can call ActorExtention.Speak(), but an object of type Actor cannot call Speak() directly. If you put "this" at the beginning of the parameter as suggested, Speak() now becomes an extension method for the Actor class, and you should be able to call ActorExtention.Speak(), as well as call Speak() directly from any Actor object and all child classes of Actor.

RSXEnthusiast: Gotcha, that makes a lot of sense. Thanks for explaining!

Inheritance/src/Inheritance/ActorExtention.cs
```diff
@@ -0,0 +1,29 @@
...
+                    result = "I really wish the programmer knew more about Big Bang Therory, so he could better make fun of me!";
+
+                    break;
+                case Penny penny:
+                    result = "If the programmer isn't mistaken, I'm just the token hot chick.";
+
+                    break;
+                case Raj raj:
+                    raj = (Raj) actor;
+                    if (raj.womenArePresent) {
```
breaman: these if statements should have been handled in the case statement, not a separate call. Also, the cast from actor to Raj was not needed because that would already be done in the case statement.

Inheritance/src/Inheritance/Food.cs
```diff
@@ -1,12 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-
-namespace Inheritance
-{
-    public class Food : Item
-    {
+﻿namespace Inheritance {
```
breaman: c# coding standards says that { should be on a separate line, not inline.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -26,17 +19,66 @@ public void PersonGetsPrinted()
...
-                    {
+                    using (var reader = new StreamReader(stream)) {
                         var lineWritten = reader.ReadLine();
                         Assert.AreEqual("Test Item", lineWritten);
                     }
                 }
             }
         }
+
+        [TestMethod]
```
breaman: Nice work on these unit tests.

PR #121, opened 2019-10-17
Assignment #3

TatyanaHubbard:


Inheritance/src/Inheritance/ActorsSpeak.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+
+namespace Inheritance
+{
+    public static class ActorsSpeak
+    {
+        public static string Speak(this Actor actor)
+        {
+            switch (actor)
```
Fervidusletum: PR looks good!

Small suggestion here: include a null case in the switch block, so that you can throw a `ArgumentNullException`, which would be a bit clearer than a null `actor` falling through to the default case and throwing a `NotSupportedException`

breaman: Good advice here, glad you followed it.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -1,13 +1,59 @@
...
+using System;
 
 namespace Inheritance.Tests
 {
     [TestClass]
     public class ActorTests
     {
         [TestMethod]
-        public void TestMethod1()
+        public void Speak_RajWithWomenPresent_ReturnsMumble()
```
breaman: Nice work on the unit tests.

PR #122, opened 2019-10-17
Assignment 3

dnisyndrum:
Completed. Added PrintInfo() functions for food and television and child classes for actor. I included Howard as well. You can't have the Big Bang Theory gang without Howard! Tests added for food, television, and the actors using the Speak() extension method.

### DIFF
--- a/Inheritance/src/Inheritance/ActorMixins.cs
+++ b/Inheritance/src/Inheritance/ActorMixins.cs
@@ -1,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorMixins
+    {
+        public static string Speak(Actor actor) => actor switch
+        {
+            Penny penny => penny.Speak(),
+            Sheldon sheldon => sheldon.Speak(),
+            Howard howard => howard.Speak(),
+            Raj raj => raj.WomenPresent ? raj.Mumble() : raj.Speak(), 
+            { } => string.Empty,
+            null => throw new ArgumentNullException("Actor not specified", nameof(actor))
+        };
+
+    }
+}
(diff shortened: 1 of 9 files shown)


TatyanaHubbard:
Good work! I think you hit all of the possible tests and got the prompt right!

Inheritance/src/Inheritance/ActorMixins.cs
```diff
@@ -0,0 +1,20 @@
...
+namespace Inheritance
+{
+    public static class ActorMixins
+    {
+        public static string Speak(this Actor actor) => actor switch
+        {
+            Penny penny => penny.Speak(),
+            Sheldon sheldon => sheldon.Speak(),
+            Howard howard => howard.Speak(),
+            Raj raj => raj.WomenPresent ? raj.Mumble() : raj.Speak(),
```
breaman: Nice implementation here for the two cases with Raj.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +6,37 @@ namespace Inheritance.Tests
     public class ActorTests
     {
         [TestMethod]
-        public void TestMethod1()
+        public void Actor_Sheldon_Speak()
         {
+            Actor sheldon = new Sheldon();
+            Assert.AreEqual(sheldon.Speak(), "AAHHH!!! OH GOD, NOT EUCLID AVENUE!!!");
```
breaman: Not docking you for this, but just an fyi on the assertions, it is the expected value first and then the actual value, so you would have wanted to swap these parameters.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -7,14 +7,63 @@
 namespace Inheritance.Tests
 {
     [TestClass]
-    public class PersonPrinterTests
+    public class PrinterTests
```
breaman: Nice job on the unit tests, very thorough.

PR #123, opened 2019-10-18
Local assignment3

kingofsky1995:


Inheritance/src/Inheritance/ActorExtention.cs
```diff
@@ -0,0 +1,29 @@
...
+namespace Inheritance
+{
+    public static class ActorExtention
+    {
+        public static string Speak(this Actor actor, string content)
+        {
+            switch (actor) {
+
+                case Raj raj:
+                    if (raj.womenArePresent)
```
breaman: this if statement should have been in the case statement, not a separate call.

Inheritance/test/Inheritance.Tests/FoodTests.cs
```diff
@@ -0,0 +1,24 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+namespace Inheritance.Tests
+{
+    [TestClass]
+    class FoodTests
+    {
+        [TestMethod]
+        public void Food_PrintInfo()
+        {
+            Food food = new Food();
```
breaman: A better test here would have been to have an IItem food so that the test proves that the interface was implemented.

Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface  IItem
```
breaman: It looks like the PR that had the original work done was overwritten instead of creating a separate one that had the abstract class converted to an interface.

PR #125, opened 2019-10-18
Assignment3 Parts 1&2

Cybotive:
Part 1 finished. Came up with solution to unit test Raj's outputs, not sure how else to do it.

Both parts finished.

Edit: apparently the idea was that Speak() would return something? I based off the name of the method it should actually output something (aka speak), hence the auto property needed to unit test.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtension.cs
+++ b/Inheritance/src/Inheritance/ActorExtension.cs
@@ -1,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static void Speak(this Actor actor, bool WomenArePresent)
+        {
+            switch (actor)
+            {
+                case Sheldon s:
+                    s.Speak();
+                    break;
+                case Penny p:
+                    p.Speak();
+                    break;
+                case Raj r when WomenArePresent:
+                    r.Mumble();
+                    break;
+                case Raj r when WomenArePresent is false:
+                    r.Speak();
+                    break;
+            }
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static void Speak(this Actor actor, bool WomenArePresent)
```
breaman: WomenArePresent should be camelCase and should also have been a property on Raj instead of a parameter here since it isn't relevant to everyone, just him.

Inheritance/src/Inheritance/Sheldon.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Sheldon : Actor
+    {
+        public string lastSaid { get; private set; }
```
breaman: Properties should be PascalCase

Inheritance/src/Inheritance/Penny.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Penny : Actor
+    {
+        public string lastSaid { get; private set; }
```
breaman: I am not sure what the LastSaid property is for here.

Inheritance/src/Inheritance/Television.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Television : Item
```

PR #128, opened 2019-10-18
Assignment3

dougdoner:


Inheritance/src/Inheritance/Penny.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Penny : Actor
+    {
+    }
```
kingofsky1995: Do you need to implement something in here?

PR #129, opened 2019-10-18
Assignment 3 snixon

samrnixon:


Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -15,28 +15,29 @@ public void PersonGetsPrinted()
...
     }
 
-    public class TestItem : Item {
+    public class TestItem : Item 
+    {
         public string Name { get; set; }
+
+        public override string PrintInfo()
+        {
+            return Name;
```
GamerBah: What happens if Name is null?

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,31 @@
...
+                    return penny.Speak();
+                case Raj raj:
+                    if(raj.WomenAreAround is true)
+                    {
+                        return raj.Mumble();
+                    }
+                    else
+                    {
+                        return raj.Speak();
+                    }
```
GamerBah: Why not replace this `if-else` block with an `elvis operator` -> `?:`? That way you can shorten it down to a one-liner.

There are also [some handy pattern matching features](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/pattern-matching) that will help shorten this down even more (which I _think_ fits the specifications better?)

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -6,26 +6,12 @@ namespace Inheritance
...
-                        return raj.Speak();
-                    }
-                default:
-                    throw new ArgumentNullException("No correct actor");
-            }
-        }
+                Sheldon sheldon => sheldon.Speak(),
+                Penny penny => penny.Speak(),
+                Raj raj => (raj.WomenAreAround ? raj.Mumble() : raj.Speak())
+            };
```
GamerBah: This looks good, but something to also keep in mind is that this method could be passed an object that has no logic for it in the switch, or a null. Should be adding a case for `{ }` and `null`, where `{ }` matches any non-null object that didn't match an earlier case (per the MS docs).

Inheritance/src/Inheritance/Printer.cs
```diff
@@ -9,11 +9,10 @@ public class Program
     {
         public static void Main()
         {
-            using (var sw = new StreamWriter(Console.OpenStandardOutput()))
-            {
-                Printer.Print(new Food { Brand = "Frosted Mini Wheats", Upc = "123456789" }, sw);
-                sw.Flush();
-            }
+            using var sw = new StreamWriter(Console.OpenStandardOutput());
```
breaman: When doing a using statement like this, you do want to have it wrapped in braces so that dispose is called when things go out of scope.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +6,39 @@ namespace Inheritance.Tests
     public class ActorTests
     {
         [TestMethod]
-        public void TestMethod1()
+        public void ExpectedSpeech_Penny()
         {
+            Penny penny = new Penny();
+            Assert.AreEqual("Hello I am Penny", ActorExtensions.Speak(penny));
```
breaman: there isn't anything wrong with calling the extension method this way, but since it was written as an extension method, you could have just called penny.Speak() as well.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -15,28 +15,33 @@ public void PersonGetsPrinted()
             // Arrange
             var item = new TestItem { Name = "Test Item" };
 
-            using (var stream = new MemoryStream()) {
-                using (var writer = new StreamWriter(stream))
-                {
-                    // Act
-                    Printer.Print(item, writer);
-                    writer.Flush();
+            using var stream = new MemoryStream();
```
breaman: why did you unwrap the using statements?

Inheritance/src/Inheritance/Television.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Television : Item
```

PR #130, opened 2019-10-18
Assignment 3

jake-mcg:


Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+    {
+        public static string SayLine(this Actor actor) =>
+            actor switch
+        {
+            Sheldon sheldon => $"{actor.GetType().Name}: { sheldon.Script}",
+            Penny penny => $"{actor.GetType().Name}: { penny.Script}",
+            Raj raj when raj.AreWomenPresent => $"{actor.GetType().Name}: {raj.Script}",
+            Raj raj when !raj.AreWomenPresent => $"{actor.GetType().Name}: {raj.Script}",
+            { } => string.Empty
+        };
```
pnotequalnp: This will fail when actor is null.

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static string SayLine(this Actor actor) =>
```
pnotequalnp: The assignment specs wanted this method called "Speak".

Inheritance/test/Inheritance.Tests/FoodTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+                    // Assert
+                    using (var reader = new StreamReader(stream))
+                    {
+                        var lineWritten = reader.ReadLine();
+                        Assert.AreEqual("<Test Upc> - <Test Brand>", lineWritten);
+                    }
+                }
+            }
+        }
```
pnotequalnp: This tests both Food and Printer at the same time, better to isolate them and test each individually.

Inheritance/test/Inheritance.Tests/TelevisionTests.cs
```diff
@@ -0,0 +1,40 @@
...
+
+                    // Assert
+                    using (var reader = new StreamReader(stream))
+                    {
+                        var lineWritten = reader.ReadLine();
+                        Assert.AreEqual("<Test Manufacturer> - <Test Size>", lineWritten);
+                    }
+                }
+            }
+        }
```
pnotequalnp: Same as with FoodTests

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,18 @@
...
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static string Speak(this Actor actor) =>
+            actor switch
+        {
+            Sheldon sheldon                     => $"{actor.GetType().Name}: { sheldon.Script}",
+            Penny penny                         => $"{actor.GetType().Name}: { penny.Script}",
+            Raj raj when raj.AreWomenPresent    => $"{actor.GetType().Name}: {raj.Script}",
```
breaman: What is the difference between the two calls for Raj? They both call the same method, so you should have had them call different methods and then not have the if check in the Script method.

PR #131, opened 2019-10-19
Assignment 3 Extra Credit

pnotequalnp:
WIP

### DIFF
--- a/Inheritance/src/Inheritance/Inheritance.csproj
+++ b/Inheritance/src/Inheritance/Inheritance.csproj
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <OutputType>exe</OutputType>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
 </Project>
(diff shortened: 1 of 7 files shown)


pnotequalnp:
Completed

Inheritance/src/Inheritance/Inheritance.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <OutputType>exe</OutputType>
+    <Nullable>enable</Nullable>
```
twofingerrightclick: Good practice. I forgot to do this. Thanks

Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface Item
```
breaman: from a naming convention, this should have been IItem instead of just Item so it is easily identified elsewhere as an interface.

PR #132, opened 2019-10-19
Assignment 3

GamerBah:
Things WIP:
- Add better tests. Null tests are a big one I think?
- Some more features to the actor classes. Currently restricted to 1 hard-coded phrase.
- Null checking in other classes I think...


GamerBah:
> You might consider moving your TelevisionTests into the general PrinterTests. There's no need for a full separate test class, but I suppose it probably doesn't matter too much for the overall solution.

I had thought about that, I guess the only thing keeping me away from that was if there would ever be any "hypothetical" feature additions to Televisions. Obviously there won't be any for this assignment, so I may as well move them.

GamerBah:
Completed.

Inheritance/src/Inheritance/ActorMixins.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+
+namespace Inheritance
+{
+
+    public static class ActorMixins
+    {
+
+        public static string Speak(Actor actor, string message) => actor switch
```
breaman: nicely done here on the implementation, but in order for this to be an extension method, the signature should have been this Actor actor, instead of just Actor actor.

Inheritance/src/Inheritance/Printer.cs
```diff
@@ -9,11 +9,9 @@ public class Program
     {
         public static void Main()
         {
-            using (var sw = new StreamWriter(Console.OpenStandardOutput()))
-            {
-                Printer.Print(new Food { Brand = "Frosted Mini Wheats", Upc = "123456789" }, sw);
-                sw.Flush();
-            }
+            using var sw = new StreamWriter(Console.OpenStandardOutput());
```
breaman: the using statement should be wrapped with enclosing curly braces so that resources are disposed correctly when they go out of scope.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -1,13 +1,45 @@
...
     [TestClass]
     public class ActorTests
     {
+
+        [TestMethod]
+        public void Sheldon_SpeaksCorrectly()
+        {
+            var sheldon = new Sheldon();
+
+            Assert.AreEqual("Test", ActorMixins.Speak(sheldon, "Test"));
```
breaman: if the extension method was implemented correctly, you could have called this as sheldon.Speak("Test") instead.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -6,37 +6,78 @@
...
+        [ExpectedException(typeof(ArgumentException))]
+        public void Food_GivenNull_ThrowsException()
+        {
+            var food = new Food();
+            food.PrintInfo();
+        }
+
+        private static string TestPrint(Item item)
+        {
+            using var stream = new MemoryStream();
```
breaman: these using statements should be wrapped with curly braces.

PR #133, opened 2019-10-19
Assignment 3

jacob-berger:
Implemented all basic requirements.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtensions.cs
+++ b/Inheritance/src/Inheritance/ActorExtensions.cs
@@ -1,0 +1,40 @@
+﻿using System;
+
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static void Speak(this Actor actor, bool WomenArePresent)
+        {
+            string actorName = actor.GetType().ToString();
+            switch (actorName)
+            {
+                case ("Raj"):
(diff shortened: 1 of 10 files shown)


Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,47 @@ public void PersonGetsPrinted()
                 }
             }
```
jake-mcg: I didn't see any unit tests for food or television items. I tested the PrintInfo methods separately in their own test classes (FoodTests and TelevisionTests), and then I tested that Item can be set to a Television or Food object respectively. Also, good call on testing for null objects

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,50 @@
...
+            //} else if (actor.GetType() == typeof(Penny))
+            //{
+            //    Penny temp = new Penny();
+            //    temp.Speak();
+            //} else
+            //{
+            //    Sheldon temp = new Sheldon();
+            //    temp.Speak();
+            //}
+
```
matHopg: This is a very elegant solution to the normally cumbersome switch statements. This at least taught me some useful tricks within c#.

PR #134, opened 2019-10-19
Assignment3_Inheritance

CDobbins1:


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,20 @@
...
+    public static class ActorExtension
+    {
+        public static string Speak(this Actor actor) =>
+            actor switch
+            {
+                Penny p => p.Speach(),
+                Sheldon s => s.Speach(),
+                Raj { WomenArePresent: true } r => r.Mumble(),
+                Raj { WomenArePresent: false } r => r.Speach(),
+                { } => "Who am I?",
```
twofingerrightclick: I like these catch cases. I was meaning to look up how to do them. Thanks

PR #135, opened 2019-10-19
Assignment 3 Extra Credit : mmwoodfo

mmwoodfo:


dbosco051394:
Everything looks good to me here! Nice work.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -1,13 +1,83 @@
...
+        }
+
+        [TestMethod]
+        public void Sheldon_ActorTest()
+        {
+            //Arrange
+            Sheldon sheldon = new Sheldon();
+
+            sheldon.Says = "Interesting. You're afraid of insects and women. Ladybugs must render you catatonic.";
+            string expected = "Sheldon : Interesting. You're afraid of insects and women. Ladybugs must render you catatonic.";
```
twofingerrightclick: lol

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+    public static class ActorExtensions
+    {
+        public static string Speak(this Actor actor)
+        {
+            switch (actor)
+            {
+                case Raj raj when raj.WomenArePresent:
+                    return $"{actor.GetType().Name} : ";
+                case Raj raj when !raj.WomenArePresent:
+                    return $"{actor.GetType().Name} : {raj.Says}";
```
twofingerrightclick: I didn't find anything in your code with my c# abilities, but this actor.GetType(), is it helpful versus just using raj as the type because the case is raj?

mmwoodfo: I typically don't like to hard code things in, so I went with using `actor.GetType` instead, there really isn't an advantage in this project however I would assume that it would be good practice to use this over returning `$"Raj : {raj.Says}"` in case we changed Raj's name or something? _(Honestly, I mostly did this so I could copy-paste my code in each case because I'm lazy)._

Inheritance/test/Inheritance.Tests/ActorTests.cs
mmwoodfo: It seemed appropriate for the project context.

PR #137, opened 2019-10-20
Assignment3

NikolasGraham:


Inheritance/src/Inheritance/Penny.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Penny : Actor
+    {
+        public string getSpeak() => "HOLY CRAP ON A CRACKER!";
```
breaman: Methods and properties should be PascalCase, so if you wanted to call it this, it should have been GetSpeak()

PR #138, opened 2019-10-20
Assignment 3 Extra Credit

dsergio:


Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface Item
```
mmwoodfo: I think this should be called IItem as it is an interface.

Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface IItem
```
breaman: You made the change from the abstract class to an interface, but there are no unit tests.

PR #139, opened 2019-10-20
Hw3 solution

dsergio:


Inheritance/test/Inheritance.Tests/ItemTests.cs
```diff
@@ -0,0 +1,67 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance.Tests
+{
+    [TestClass]
+    public class ItemTests
```
breaman: Not sure what these unit tests are proving. Seems like you are just making sure that if you set a property, then the property actually gets set, but the properties are auto implemented, so you don't need to test that. There are not meaningful unit tests for Food and Manufacturer here.

PR #140, opened 2019-10-20
Assignment 3

crackerJackTech:
Need to complete ActorTests and Extra Credit.

### DIFF
--- a/Inheritance/src/Inheritance/ExtensionMethods.cs
+++ b/Inheritance/src/Inheritance/ExtensionMethods.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ExtensionMethods
+    {
+        public static string Speak(object actor) =>
+
+            actor switch
+            {
+                Raj { WomenArePresent: false } => "My name is Raj",
+                Raj r => "*mumble",
+                Sheldon s => "My name is Sheldon",
+                Penny p => "My name is Penny",
+
+
+                { } => throw new ArgumentException(message: "Not a known vehicle type", paramName: nameof(actor)),
+                _ => throw new ArgumentNullException(nameof(actor))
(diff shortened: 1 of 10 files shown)


Inheritance/src/Inheritance/ExtensionMethods.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ExtensionMethods
+    {
+        public static string Speak(this object actor) =>
```
breaman: We wouldn't want this extension method on all objects, just the Actor class, so this signature should have been "this Actor actor".

Inheritance/src/Inheritance/Printer.cs
```diff
@@ -20,6 +20,11 @@ public static class Printer
     {
         public static void Print(Item item, TextWriter writer)
         {
+            if(item is null || writer is null)
+            {
+                throw new ArgumentNullException();
```
breaman: If you are going to do this check, it is nice if you tell the user what argument was null.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -1,13 +1,52 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Inheritance.Tests
 {
     [TestClass]
-    public class ActorTests
+    public class ActorTests : Actor
```
breaman: Why does the test subclass the Actor class? This shouldn't be implementing any Actor functionality.

Inheritance/src/Inheritance/Television.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    class Television : Item
```

PR #141, opened 2019-10-20
Assignment3

skozubenko1:
Part 1 & 2
Test cases for Actors

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtension.cs
+++ b/Inheritance/src/Inheritance/ActorExtension.cs
@@ -1,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {     
+        public static string Speak(this Actor actor, bool WomenArePresent = false)
+        {
+            if (actor == null)
+                throw new ArgumentNullException();
+
+            switch (actor)
+            {
+                case Penny p:
+                    return p.Speaking();
+                case Sheldon s:
+                    return s.Speaking();                    
+                case Raj r:
+                    if (WomenArePresent)
+                        return r.Mumbling();
+                    else
+                        return r.Speaking();
+                default:
+                    throw new ArgumentNullException(nameof(actor));                   
+            }
+        }
+     }
+
+}
--- a/Inheritance/src/Inheritance/ActorExtension.cs
+++ b/Inheritance/src/Inheritance/ActorExtension.cs
@@ -8,7 +8,7 @@ namespace Inheritance
     {     
         public static string Speak(this Actor actor, bool WomenArePresent = false)
         {
-            if (actor == null)
+            if (actor is null)
                 throw new ArgumentNullException();
 
             switch (actor)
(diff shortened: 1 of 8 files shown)


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {     
+        public static string Speak(this Actor actor, bool WomenArePresent = false)
+        {
+            if (actor == null)
```
kenny2892: You should use "is null" instead of "== null" just in-case the equal operator has been overwritten.

skozubenko1: thank you for the suggestion, code corrected!

Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {     
+        public static string Speak(this Actor actor, bool WomenArePresent = false)
```
breaman: parameters should be camelCase and womenArePresent should really be a property on Raj, not a parameter into this method.

Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,31 @@
...
+                throw new ArgumentNullException();
+
+            switch (actor)
+            {
+                case Penny p:
+                    return p.Speaking();
+                case Sheldon s:
+                    return s.Speaking();                    
+                case Raj r:
+                    if (WomenArePresent)
```
breaman: this check should have been done on the case statement, not as a separate expression.

Inheritance/src/Inheritance/Penny.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Penny : Actor
+    {
+        public string Speaking() => "Pengny is Speaking";
```
breaman: my guess is Pengny is a typo, be careful of that.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -5,9 +5,32 @@ namespace Inheritance.Tests
...
+            var talk = actor.Speak(WomenArePresent);
+            Assert.IsNotNull(talk);
+            Assert.IsTrue(talk.Length > 0, $"{actor.GetType()} didn't say anything");
+            return talk;
+        }
+
+        [TestMethod]
+        public void TestMethodPenny()
+        {
+            TestActor(new Penny());
```
breaman: The test wrapper is nice to show that something is said, but you aren't verifying that what was said is what was expected.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +35,40 @@ public void PersonGetsPrinted()
                 }
             }
         }
+        [TestMethod]
```
breaman: Where is the test for the Food item?

PR #142, opened 2019-10-20
Assignment3 part2

dbosco051394:
I have everything completed for the part 2 requirements here. If I missed anything or misunderstood a requirement please let me know!

### DIFF
--- a/Inheritance/src/Inheritance/Situation.cs
+++ b/Inheritance/src/Inheritance/Situation.cs
@@ -1,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class Situation
+    {
+        public static bool WomenArePresent = false;
+        
+        public static string Speak(this Actor a)
+        {
+            string type = a.GetType().Name;
+            if (type == "Penny")
+            {
+                WomenArePresent = true;
+                return "Hello, I am Penny.";
+            }
+            else if (type == "Sheldon")
+            {
+                return "Hello, I am Sheldon.";
+            }
+            else if (type == "Raj")
+            {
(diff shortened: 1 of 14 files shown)


Inheritance/src/Inheritance/Situation.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace Inheritance
+{
+    public static class Situation
+    {
+        public static bool WomenArePresent = false;
+        
+        public static string Speak(this Actor a)
+        {
+            string type = a.GetType().Name;
```
breaman: This was suppose to use pattern matching (the switch statement stuff we went over in class) instead of a bunch of if statements.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            //Arrange
+            Actor sheldon = new Sheldon();
+            string spokenWords = "";
+            string expectedWords = "Hello, I am Sheldon.";
+
+            //Act
+            spokenWords = sheldon.Speak();
+
+            //Assert
+            Assert.AreEqual(spokenWords, expectedWords);
```
breaman: These parameters should have been swapped, the AreEqual takes (expected, actual).

Inheritance/src/Inheritance/Situation.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class Situation
+    {
+        public static bool WomenArePresent = false;
```
breaman: This should have only existed on Raj since it only effects him.

PR #144, opened 2019-10-20
Assignment3 extra credit

dbosco051394:
I've attempted the extra credit. If I didn't satisfy a requirement or have any errors please let me know!

### DIFF
--- a/Inheritance/src/Inheritance/Item.cs
+++ b/Inheritance/src/Inheritance/Item.cs
@@ -1,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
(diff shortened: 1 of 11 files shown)


Inheritance/src/Inheritance/Item.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public interface Item
```
mmwoodfo: I believe this should be named IItem since it is an interface.

dbosco051394: I forgot about that naming convention. I'll add that in. Thanks!

PR #145, opened 2019-10-20
Assignment 3 - Parts 1&2

JerettLatimer:


Inheritance/test/Inheritance.Tests/ActorExtension.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    static class ActorExtension
+    {
+        public static string Speak(Actor actor) => actor switch
```
breaman: to be an extension method, this signature should be (this Actor actor). I noticed your unit tests never actually call any methods here.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +6,55 @@ namespace Inheritance.Tests
...
+        public void PennySpeaks()
         {
+            // Arrange
+            var pennyTest = new Penny();
+
+            // Act
+            string phrase = "My name is Penny!";
+
+            // Assert
+            Assert.AreEqual(phrase, pennyTest.Phrase());
```
breaman: These tests call the class methods directly instead of using the extension method to figure out what should be called. These all should have called .Speak instead.

PR #146, opened 2019-10-20
Assignment 3- parts 1 and 2

AustinHall7798:
I think I have everything, but any feedback is appreciated


Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -7,36 +7,101 @@
 namespace Inheritance.Tests
 {
     [TestClass]
-    public class PersonPrinterTests
+    public class PrinterTests
     {
         [TestMethod]
-        public void PersonGetsPrinted()
+        public void ItemGetsPrinted()
```
JerettLatimer: Good job on your tests. They effectively test for expected and edge cases including null.

AustinHall7798:
After looking at others' assignments, I realize maybe I wasn't supposed to create a constructor for Food and Television so I'll go without that on the extra credit portion

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +7,67 @@ namespace Inheritance.Tests
...
+        }
+
+        [TestMethod]
+        public void PennySpeakTest_()
+        {
+            //Arrange
+            Penny penny = new Penny();
+
+            //Act
+            string speech = ExtensionClass.Speak(penny);
```
breaman: Since this was written as an extension method, you could have just called this as penny.Speak() as well.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +7,67 @@ namespace Inheritance.Tests
...
         [TestMethod]
-        public void TestMethod1()
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void NullValuePassedToSpeak_ExceptionExpected()
         {
+            ExtensionClass.Speak(null);
+        }
+
+        [TestMethod]
+        public void PennySpeakTest_()
```
breaman: you shouldn't end your method names with an underscore.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +7,67 @@ namespace Inheritance.Tests
...
+        public void SheldonSpeakTest_()
+        {
+            //Arrange
+            Sheldon sheldon = new Sheldon();
+
+            //Act
+            string speech = ExtensionClass.Speak(sheldon);
+
+            //Assert
+            Assert.AreEqual(speech, "Bazinga!");
```
breaman: the signature for AreEqual is (expected, actual) so these parameters should have been swapped.

PR #147, opened 2019-10-21
Assignment3

twofingerrightclick:


markdietzler:
Not testing for null actor or writer?

skozubenko1:
I like how you break all your work down with comments

Inheritance/src/Inheritance/BigBangCastMembers.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        
+        public void Speak(TextWriter writer)
+        {
+            writer.WriteLine(Name + " says \"Who do we love?\"");
+        }
+    }
+
+
+    public class Raj : Actor
```
breaman: Best practices has a separate file per class, these should not have been grouped into one big file.

Inheritance/src/Inheritance/BigBangCastMembers.cs
```diff
@@ -0,0 +1,69 @@
...
+        public string Name
+        {
+            get { return "Raj"; }
+
+        }
+
+
+        private bool womenPresent= false;
+
+        public bool WomenPresent
```
breaman: you could have just made this an auto-implemented property

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -1,13 +1,106 @@
...
+
+            string lineWritten = writerHelper(actor);
+            
+            Assert.AreEqual("Raj mumbles", lineWritten);
+                    
+                
+            
+        }
+
+        public string writerHelper(Actor actor)
```
breaman: methods should be defined as PascalCase.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,41 @@ public void PersonGetsPrinted()
                 }
             }
         }
+
+        [TestMethod]
```
breaman: Where is the test for Food?

PR #148, opened 2019-10-21
Assignment 3

whereisdaniru:
Apparently I never pushed this up....

Assignment 3 Part 1 and 2

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtension.cs
+++ b/Inheritance/src/Inheritance/ActorExtension.cs
@@ -1,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static String Speak(this Actor actor)
+        {
+            string line = "";
+
+            if(actor is Penny)
+            {
+                Penny penny = (Penny) actor;
+                line = penny.getSpeak();
+            }
+            if(actor is Sheldon)
+            {
+                Sheldon sheldon = (Sheldon) actor;
+                line = sheldon.getSpeak();
+            }
+            if(actor is Raj)
+            {
+                Raj raj = (Raj) actor;
+                if(raj.womenArePresent == true)
+                {
+                    line = raj.getSpeakWomen();
+                }
+                else if(raj.womenArePresent == false)
+                {
+                    line = raj.getSpeakNoWomen();
+                }
+            }
+
+            return line;
+        }
+    }
+}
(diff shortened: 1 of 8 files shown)


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static String Speak(this Actor actor)
```
samrnixon: The Part 2 instructions mention the logic should be expressed in a switch statement, might make this easier!

whereisdaniru: Oh! I completely missed that. Thank you!

Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+
+            string line = "";
+
+            if (actor is null)
+            {
+                throw new ArgumentNullException("Actor unknown");
+            }
+            else
```
breaman: nothing wrong with having this in an else statement, but it is also unnecessary since if actor is null, the exception would be thrown and the method would exit.

Inheritance/src/Inheritance/Penny.cs
```diff
@@ -0,0 +1,14 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Penny : Actor
+    {
+        public string getSpeak()
```
breaman: methods should be PascalCase

Inheritance/src/Inheritance/Raj.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Raj : Actor
+    {
+        public bool womenArePresent { get; set; }
```
breaman: properties should be PascalCase

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,70 @@ public void PersonGetsPrinted()
                 }
             }
         }
+
+        [TestMethod]
```
breaman: nice work on these unit tests.

PR #149, opened 2019-10-21
Assignment 3

kenny2892:


Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,33 @@
...
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static void Speak(this Actor actor, TextWriter writer)
+        {
+            if (actor is null || writer is null)
+                throw new ArgumentNullException();
+
+            switch (actor)
```
NikolasGraham: This looks pretty good, but if the actor is just an Actor object, it goes through without saying anything. This could be designed this way and be perfectly fine, but it could be good to include something for Actor, even if it's just a Console.log, since you don't have a default.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,68 @@ public void PersonGetsPrinted()
...
+        }
     }
 
     public class TestItem : Item {
         public string Name { get; set; }
+
+        public override string PrintInfo()
+        {
+            return Name;
+        }
```
NikolasGraham: All the rest looks good, nice job.

Inheritance/src/Inheritance/ActorExtension.cs
```diff
@@ -0,0 +1,33 @@
+﻿using System;
+using System.IO;
+
+namespace Inheritance
+{
+    public static class ActorExtension
+    {
+        public static void Speak(this Actor actor, TextWriter writer)
```
breaman: Wrapping everything they say in a TextWriter wasn't needed, but not a bad touch. Could have left this just as a string, though, since the TextWriter was only needed for the other part of the assignment.

PR #150, opened 2019-10-21
Assignment 3 Extra Credit

kenny2892:


Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface Item
```
breaman: by convention, an interface should be defined as IItem, not just Item.

Inheritance/src/Inheritance/Television.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+
+namespace Inheritance
+{
+    public class Television : Item
```
breaman: Where are the unit tests to show that Printer.Print still works?

PR #152, opened 2019-10-22
Created Assignment 3 solution

markdietzler:
Solution WITHOUT extra credit

### DIFF
--- a/Inheritance/src/Inheritance/Raj.cs
+++ b/Inheritance/src/Inheritance/Raj.cs
@@ -1,0 +1,32 @@
+﻿using System;
+using System.IO;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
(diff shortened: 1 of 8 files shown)


crackerJackTech:
In your Speak method, a pattern expression could be used with the switch statement but either way it works! Besides that everything looks great.

Inheritance/src/Inheritance/Raj.cs
```diff
@@ -0,0 +1,32 @@
...
+    public class Raj : Actor
+    {
+        public bool WomanIsInView { get; set; }
+
+        public Raj()
+        {
+            WomanIsInView = false;
+        }
+
+        public Raj(bool thereIsAWomanPresent)
```
ashermancinelli: Nice idea for the constructor!

PR #153, opened 2019-10-22
Assignment 3, part 1 and 2

Tpirtle:
Part 1 and 2 of Assignment 3 completed

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtensionMethods.cs
+++ b/Inheritance/src/Inheritance/ActorExtensionMethods.cs
@@ -1,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Inheritance
+{
+    public class ActorExtensionMethods
+    {
+        public string Speak(string castMem, bool womenPresent)
+        {
+            Penny penny=new Penny();
+            Raj raj=new Raj();
+            Sheldon sheldon=new Sheldon();
+
+            string wordsSpoken="";
+
+            switch (castMem)
+            {
+                case "Raj":
+                    if (womenPresent)
+                        wordsSpoken = raj.Speak();
+                    else
+                        wordsSpoken = raj.SpeakWomenPresent();
+                    break;
+                case "Sheldon":
+                    wordsSpoken = sheldon.Speak();
+                    break;
+                case "Penny":
+                    wordsSpoken = penny.Speak();
+                    break;
+
+            }
+
+            return wordsSpoken;
+        }
+    }
+}
--- a/Inheritance/src/Inheritance/Raj.cs
+++ b/Inheritance/src/Inheritance/Raj.cs
@@ -1,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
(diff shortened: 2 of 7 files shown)


Inheritance/src/Inheritance/ActorExtensionMethods.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Inheritance
+{
+    public class ActorExtensionMethods
+    {
+        public string Speak(string castMem, bool womenPresent)
```
dougdoner: Instead of passing in the name of the cast member, we can use extension methods by using (this _classname_ castMem).
This way when we instantiate an actor object, we can call the Speak() method without having to instantiate an ActorExtensionMethods object

Inheritance/src/Inheritance/ActorExtensionMethods.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            Penny penny=new Penny();
+            Raj raj=new Raj();
+            Sheldon sheldon=new Sheldon();
+
+            string wordsSpoken="";
+
+            switch (castMem)
+            {
+                case "Raj":
```
dougdoner: You can also use pattern matching here. If you used an extension method and called it from the object, you can check the type of the object inside of the extension method e.g.:
`object is Penny p`
then you can act on the object p by invoking p.Speak(), which will call the Penny.Speak() method.

Inheritance/src/Inheritance/Raj.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Inheritance
+{
+    public class Raj : Actor
+    {
+
```
dougdoner: I would recommend putting a property here for "WomenArePresent", since it's unique to the Raj class and you can pass a boolean into the constructor that way.

Inheritance/test/Inheritance.Tests/ActorTests.cs
```diff
@@ -6,8 +6,35 @@ namespace Inheritance.Tests
     public class ActorTests
     {
         [TestMethod]
-        public void TestMethod1()
+        public void TestSheldon()
         {
+            ActorExtensionMethods sheldon = new ActorExtensionMethods();
```
dougdoner: If you created an extension method for the _actor_ class, you can instantiate the sheldon variable as type _Actor_, then call Actor.Speak() (which you can then use pattern matching logic inside of the extension method to return expected behavior).

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,44 @@ public void PersonGetsPrinted()
                 }
             }
         }
+
+        [TestMethod]
+        public void Test_ItemUpc()
```
breaman: These tests here don't actually test anything except for the fact that you can set an auto-implemented property (which isn't logic you should have tested). Needed to copy the test that was already written for the actual Printer.Print and used the Food and Television object in it.

PR #155, opened 2019-10-22
Finished Part 1 and 2

matHopg:
Finished Part 1 and 2 of assignment.

### DIFF
--- a/Inheritance/src/Inheritance/ActorExtensions.cs
+++ b/Inheritance/src/Inheritance/ActorExtensions.cs
@@ -1,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static void Speak(this Actor actor, bool WomenArePresent)
+        {
+            //Had old lame switch statement but found elegant solution doing code review for jacob-berger
+            string spoke;
+            spoke = (actor switch
+            {
+                Penny penny => spoke = (new Penny().Speak()),
+                Sheldon sheldon => spoke = (new Sheldon().Speak()),
+                Raj raj when (WomenArePresent) => spoke = (new Raj().SpeakWithWomenPresent()),
+                Raj raj when (!WomenArePresent) => spoke = (new Raj().SpeakWithoutWomenPresent()),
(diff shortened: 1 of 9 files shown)


Inheritance/src/Inheritance/Food.cs
```diff
@@ -8,5 +8,18 @@ public class Food : Item
...
+        {
+            if(Upc is null || Upc == "")
+            {
+                throw new ArgumentNullException("Upc is not valid", new ArgumentNullException());
+            }
+            if(Brand is null || Brand == "")
+            {
+                throw new ArgumentNullException("Brand is not valid", new ArgumentNullException());
+            }
+            return $"{Upc} - {Brand}";
```
AustinHall7798: I believe this should be returning < Upc > - < Brand >, with the angle brackets around the words.

Inheritance/src/Inheritance/ActorExtensions.cs
```diff
@@ -0,0 +1,25 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Inheritance
+{
+    public static class ActorExtensions
+    {
+        public static void Speak(this Actor actor, bool WomenArePresent)
```
breaman: parameters should be camelCase and in this situation, the womenArePresent should have actually just been a property on Raj since it only pertains to him.

Inheritance/src/Inheritance/Television.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Inheritance
+{
+    public class Television : Item
+    {
+        public string Size { get; set; }
+        public string Manufacturer { get; set; }
+
+        public override string PrintInfo()
+        {
+            if(Size is null || Size == "")
```
breaman: not docking you, but this could have been implemented as if (string.IsNullOrWhitespace(Size)) as well? Just an fyi.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +34,213 @@ public void PersonGetsPrinted()
                 }
             }
         }
+
+        [TestMethod]
```
breaman: Excellent work on the tests here.

PR #156, opened 2019-10-22
Assignment3 extra credit

AustinHall7798:


dnisyndrum:
Looks solid to me!

Inheritance/src/Inheritance/Item.cs
```diff
@@ -4,8 +4,8 @@
 
 namespace Inheritance
 {
-    public abstract class Item
+    public interface Item
```
breaman: naming conventions says this should be IItem since it is an interface.

Inheritance/test/Inheritance.Tests/PrinterTests.cs
```diff
@@ -34,9 +35,80 @@ public void PersonGetsPrinted()
                 }
             }
         }
-    }
 
-    public class TestItem : Item {
-        public string Name { get; set; }
+        [TestMethod]
```
breaman: These tests don't prove that things are going through the Printer.Print method. while the interface was implemented correctly by the classes, these tests aren't showing that behavior.

PR #157, opened 2019-10-23
As4

ashermancinelli:
WIP


Configuration/FileConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+                        value = _value;
+                        return true;
+                    }
+                }
+                value = null;
+                return false;
+            }
+        }
+
+        #pragma warning disable CA1303
```
pnotequalnp: This should be in a a suppression file since per the readme it applies to all projects moving forward.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+                throw new ArgumentException($"Got multiple '=' characters for {nameof(input)}.");
+
+            string name = values[0], value = values[1];
+
+            SanitizeInput(name, value);
+            
+            return (name, value);
+        }
+
+        private static void SanitizeInput(string name, string? value)
```
pnotequalnp: Interesting choice to move validation to a separate method. 👍 

However I wouldn't call it sanitizing since that would seem to imply that it is modifying the input into something valid, rather than simply checking if it is valid.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,79 @@
...
+        {
+            if ((name.Any(c => c == ' ')) || (value.Any(c => c == ' ')))
```
pnotequalnp: the `.Contains` LINQ extension method seems more appropriate for this use case.

MockConfig.Tests/MockConfigTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    [TestClass]
+    public class MockConfigTests
+    {
+        [TestMethod]
+        public void GetConfigValue_UnsetName_ReturnsNull()
+        {
+            string name="SomeName";
+            string? value;
+            var sut = new MockConfig();
+            Assert.IsFalse(sut.GetConfigValue(name, out value));
```
pnotequalnp: Though not necessary, it is possible to declare a variable in the method call like `sut.GetConfigValue(name, out string? value)`

MockConfig.Tests/MockConfigTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [DataRow("SomeName", "SomeValue ")]
+        public void SetConfigValue_BadConfigValue_ThrowsException(string name, string value)
+        {
+            var sut = new MockConfig();
+            sut.SetConfigValue(name, value);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        [DataRow("SomeName", "")]
```
pnotequalnp: Passing an empty string should not throw an `ArgumentNullException` because the argument is not null.

MockConfig.Tests/MockConfigTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [DataRow("SomeName", null)]
+        public void SetConfigValue_NullConfigValue_ThrowsException(string name, string? value)
```
pnotequalnp: Since you're using the same name in every test, it's not necessary to have it as a parameter.

MockConfig/MockConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+
+namespace Mock
+{
+    public class MockConfig : IConfig
+    {
+        private Dictionary<string, string?> ConfigDict;
+
+        public MockConfig()
+        {
+            ConfigDict = new Dictionary<string, string?>();
```
pnotequalnp: Perhaps consider doing this initialization outside the constructor instead.

SampleApp/Program.cs
```diff
@@ -1,12 +1,103 @@
...
 {
-    class Program
+    public class Program
     {
-        static void Main()
+        public IConfig? Config { get; set; }
+
+        public Program()
+        {
+            if (Config is null) Config = new MockConfig();
```
pnotequalnp: The constructor happens before the object initializer. In your constructor, `Config` will always be null, and you will always assign a new `MockConfig` to it. After that, the caller may or may not assign a new `IConfig` to the field.

SampleApp/Program.cs
```diff
@@ -1,12 +1,103 @@
...
+                Console.WriteLine(
+                    $"Setting {name} to value {value} in {nameof(Config)} config.");
+
+                /*
+                 * Null/empty is checked in the setting function,
+                 * so we don't have to check here
+                 */
+                #pragma warning disable CS8602
+                Config.SetConfigValue(name, value);
+                #pragma warning restore
```
pnotequalnp: I believe `#nullable enable/disable` is preferable to manually changing the warning.

SampleApp/Program.cs
```diff
@@ -1,12 +1,103 @@
...
+                 * SanitizeValue will do the null checks for us
+                 * here, so we can disable the warning and
+                 * restore immediately
+                 */
+                #pragma warning disable CS8602
+                MockConfig.SanitizeValue(name);
+                Config.GetConfigValue(name, out outValue);
+                #pragma warning restore
+                if (outValue is null)
+                    throw new NullReferenceException($"Got null value from {nameof(Config)}");
```
pnotequalnp: You should never throw a `NullReferenceException`, that is reserved for the runtime's use.

SampleApp/Program.cs
```diff
@@ -1,12 +1,103 @@
...
+                    break;
+                case "file":
+                    sampleApp.Config = new FileConfig();
+                    break;
+                case "env":
+                    sampleApp.Config = new EnvironmentConfig();
+                    break;
+                default:
+                    throw new ArgumentException($"{nameof(configType)} cannot take passed value.");
+            }
```
pnotequalnp: This would be a great use case for a switch expression rather than a switch statement.

SampleApp.Tests/SampleAppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            // assert
+            foreach (var (name, value) in Program.SettingValues)
+            {
+                string? outValue;
+                Assert.IsTrue(sut.Config.GetConfigValue(name, out outValue));
+                Assert.IsNotNull(outValue);
+                Assert.AreEqual(value, outValue);
+            }
+        }
```
pnotequalnp: Testing `Program` and `EnvironmentConfig` together is redundant since both are tested separately.

MockConfig/MockConfig.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                value = ConfigDict[name];
+                return true;
+            }
+            else
+            {
+                value = null;
+                return false;
+            }
+        }
```
kingofsky1995: are you trying to add them again after checking on them? I'm not sure if this is appropriate but it is supposed to return true or false right? add or remove the value will be a different method?

Configuration/Configuration.csproj
```diff
@@ -1,8 +1,9 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netstandard2.1</TargetFramework>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
```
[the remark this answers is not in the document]
ashermancinelli: compatibility with my dotnet version - homebrew install needed this.

SampleApp/Program.cs
ashermancinelli: Which is the intended behavior, though you are right that I do not have to check for null.

SampleApp.Tests/SampleAppTests.cs
ashermancinelli: Good call!

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,46 @@
...
+        public void Cleanup()
+        {
+            foreach (string key in UsedKeys)
+            {
+                Console.WriteLine($"Key: {key}");
+                Environment.SetEnvironmentVariable(key, null);
+            }
+        }
+
+        ~EnvironmentConfig() => Cleanup();
```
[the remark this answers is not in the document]
ashermancinelli: Yes, although I may want a cleanup function anyways and I am also trying to be more explicit than I may have to to avoid getting a zero on the assignment! but thanks for the comment.

MockConfig.Tests/MockConfigTests.cs
ashermancinelli: Thanks!

PR #159, opened 2019-10-24
Assignment4 : mmwoodfo

mmwoodfo:


mmwoodfo:
Changes in all files still being made to this PR, but feel free to look & comment on what is currently up.

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,38 @@
...
+        {
+            if(Environment.GetEnvironmentVariable(name) is null)
+            {
+                return false;
+            }
+            else
+            {
+                return true;
+            }
+        }
```
ashermancinelli: This could be a lot more concise with

```c#
public bool GetConfigValue(string name, string? value) =>
    !(Environment.GetEnvironmentVariable(name) is null)
```

if that seems clearer to you. No need to sacrifice clarity for brevity though.

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,38 @@
...
+            {
+                Environment.SetEnvironmentVariable(name, value);
+                return true;
+            }
+        }
+
+        //~EnvironmentConfig()
+        //{
+        //    //some kind of cleanup?
+        //}
```
ashermancinelli: Could keep a list of all vars that were set by this instance, and unset them before finalizing.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,74 @@
...
+        }
+
+#pragma warning disable CA1822 // Mark members as static
+        public bool ValidateNameAndValue(string name, string? value)
+#pragma warning restore CA1822 // Mark members as static
+        {
+            if (name is null || value is null)
+            {
+                return false;
+            }
```
kingofsky1995: is this a repeat process? i thought you did check this null condition before you call this method?

mmwoodfo: It is yes! thank you for pointing that out. I'm going to remove it in my method that writes to the config file and keep it in my string validation method.

MarkMichaelis:
For more comments, look for "MMM" in the following commit: https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/pull/159/commits/07af2afc3b6fa4a0c3f55b2b49b4a7caf7a2130e

PR #160, opened 2019-10-24
Assignment 4

pnotequalnp:
WIP


Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        public bool GetConfigValue(string name, out string? value)
+        {
+            value = Environment.GetEnvironmentVariable(name);
+            return !(value is null);
+        }
+
+        public bool SetConfigValue(string name, string? value)
+        {
+            if (name is null) throw new ArgumentNullException(nameof(name));
```
mmwoodfo: I might be confused about the nullable types, but can't name not be passed as null? it would only be able to be passed as null if it was declared as `string? name` in the param list. That could be an inaccurate assumption, but I think you should check if the value is null or not, as I think value is what is being set to the specified name?

pnotequalnp: Nullable reference types are a new opt-in C# 8 feature. Someone calling the function from a project without nullable reference types enabled could pass null to it with no warning. FxCop will actually give you a warning if you don't check for null with a public method.

mmwoodfo: Okay, thank you! I will update my project with that information then! I think I recall that in the lecture now that you've reiterated it.

MarkMichaelis:
For more comments, look for "MMM" in the following commit: https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/pull/160/commits/2da827be3ce6784132216aecbd3dafe60f11629b

MarkMichaelis:
For more comments, look for "MMM" in the following commit: https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/commit/2da827be3ce6784132216aecbd3dafe60f11629b

PR #164, opened 2019-10-26
Assignment4 snixon

samrnixon:


crackerJackTech:
Great stubbed out methods for the GetConfigValue and SetConfig value in your EnvironmentsConfig file.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        [TestMethod]
+        public void FileConfig_ValidConstructor()
+        {
+            _ = new FileConfig("config.settings");
+        }
+
+        [TestMethod]
+        public void FileConfig_InvalidConstructor_CallGetValue_ReturnsTrue()
+        {
+            string filePath = Directory.GetCurrentDirectory() + "\\configTEST.settings";
```
MarkMichaelis: Use `System.IO.Path` to concat file names.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            var file = File.Create(filePath);
```
MarkMichaelis: Preferably use the explicit type name here, since it isn't obvious what `File.Create()` returns.

File.Create returns a stream which supports IDisposable so a using statement should be used.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            file.Close();
```
MarkMichaelis: Why do you close and the re-open the Stream writer?

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            using (StreamWriter sr = new StreamWriter(filePath))
+            {
+                sr.Write("testNAME=testVALUE");
+                sr.Close();
+            }
+            
+            FileConfig test = new FileConfig(filePath);
+            Assert.IsTrue(test.GetConfigValue("testNAME", out string? value));
+            File.Delete(filePath);
```
MarkMichaelis: What happens if you are debugging an then call stop before the delete runs?  Use delete at the start of the test (potentially doing both).

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+        }
+
+        [TestMethod]
+        public void FileConfig_SetConfig_SameNameNewValue()
+        {
+            string filePath = Directory.GetCurrentDirectory() + "\\configTEST.settings";
+            var file = File.Create(filePath);
+            file.Close();
+            FileConfig fileConfig = new FileConfig(filePath);
+            fileConfig.SetConfigValue("nameTEST", null!);
```
MarkMichaelis: What is this line for?  You don't seem to test if it works.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,98 @@
...
+            fileConfig.SetConfigValue("nameTEST", "valueTEST1");
+
+            Assert.IsTrue(fileConfig.SetConfigValue("nameTEST2", "valueTEST2"));
```
MarkMichaelis: You aren't testing to see if the set works?  Why not call `GetConfigValue()` to verify the set worked.

MarkMichaelis:
## <Full Name Missing> (samrnixon)

## Assignment4

| Assignment  | **15** |
| Create a *`EnvironmentConfig`* configuration class that implements `IConfig` and settings using `Environment.SetEnvironment()` and `Environment.GetEnvironment()` | ✔ |
| **Settings should not persist across process settings (if this occurs you will get a 0)** | ✔ |
| Implement an `FileConfig` that reads/writes settings to a config.settings file in the same directory as the application.  Settings are stored as <name>=<value>.  (You can use String.Split to retrieve individual values). | ✔ |
| Do not allow null, empty string, or names with spaces or '='.  Feel free to document other assumptions using unit tests. | ✔ |
| Choose simplicity over than complexity. | ✔ |
| Implement **SampleApp** to print out all the configuration settings based on hard coded values for config (iterating over the values is not supported by the interface). | ✔ |
| Have the **SampleApp** print out a hard coded list of settings to the console.  You do not need to unit test invocations of `System.Console` methods. | ✔ |
| Unit test the **SampleApp** using a `MockConfig` class (implementing `IConfig`) that uses settings stored in memory (such as an array or a `List<T>`) | ✔ |
| Use data driven testing to work with multiple values against the config API. | ❌ |
| Add code analysis to **SampleApp** address all warnings.  It is okay to appropriately disable rules such as  localization/culture rules. **Note: This is true for projects in all assignments going forward.*** | ❌ |
| Nullablility should be enabled at the project level for all projects and all warnings should be handled without disabling them.  Also, be sure to check for null when appropriate.  **Note: This is true for all assignments going forward.*** | ❌ |

| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |

| Configuration | Enable | Assignment4.sln |  | CS1701, CS1702, CS2008 | 4 |
| Configuration.Tests | Enable | Assignment4.sln |  | CS1701, CS1702 | 4 |
| SampleApp | Disable | Assignment4.sln |  | CS1701, CS1702 | 4 |

| Testing | **15** |
| Test Run Succeeded: Total tests=10, Passed=10, Failed=0, Skipped=0 | **65%** |


| Extra Credit | **0** |
| Provide a means to iterate over all the settings on a provider rather than only pulling out specific ones by name. |  ❌ |
| Refactor the test classes that test through `IConfig` into a base class so that the same tests can be used to test multiple. | ❌ |
| Create an IConfigGroup that returns a set of setting based on wildcard filters (rather than just a single setting) | ❌ |
| Provide a mock instance for `System.Console`.  (You do not need to test method invocations of `System.Console`.) | ❌ |
| Create and implement an `IConfigAsync` interface (we are unlikely to cover async this quarter). | ❌ |

PR #165, opened 2019-10-26
Assignment4

dnisyndrum:
WIP


Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,44 @@
...
+        public bool SetConfigValue(string name, string? value)
+        {
+            if(value is null)
+            {
+                return false;
+            }
+            else
+            {
+                ConfigNames.Add(name);
+                Environment.SetEnvironmentVariable(name, value);
```
skozubenko1: You need to ensure that your setting Enviroment variable for this process only.

  Environment.SetEnvironmentVariable(name, val, **EnvironmentVariableTarget.Process**);

twofingerrightclick: I believe process is the default of SetEnvironmentVariable(String, String)
"SetEnvironmentVariable(String, String) | Creates, modifies, or deletes an environment variable stored in the current process."
https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=netframework-4.8

MarkMichaelis:
## Assignment 4 (PR 165): **40** (94% Code Coverage but `Program` not tested)

| Implement IConfig in FileConfig and MockConfig | ❌ |
| Code Review Done | ✔ |

| Testing |  |
| SampleApp Program Tested from SampleApp.Tests | ❌ |

| Extra Credit |  |

PR #166, opened 2019-10-27
AS4 Doug Doner

dougdoner:


Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            bool settingWasSet = fileconfig.SetConfigValue("Test", "Hello");
+            bool settingExists = fileconfig.GetConfigValue("Test", out string? settingValue);
+
+            //Assert
+            Assert.IsTrue(settingExists);
+            Assert.IsNotNull(settingValue);
+        }
+
+        [TestMethod]
+        public void FileConfig_SetConfigValue_SetsValue()
```
AustinHall7798: It could be useful to write some tests to make sure that invalid names or values don’t get written to the file

dougdoner: Oh yeah good idea, I will implement that.

dougdoner: Fixed in 711a184

dougdoner: Fixed in 19b3a16

MarkMichaelis:
## Assignment 4 (PR 166): **30** (97% Code Coverage but no SampleApp.Tests testing SampleApp)

| The assignment misses several required elements. The solution contains errors. | 10 |
| Code Review Done | ✔ |

| Testing |  |

| Extra Credit |  |

PR #167, opened 2019-10-27
Assignment4

CDobbins1:
Still need to clean up current code, as well as implement sample app.


MarkMichaelis:
## Assignment 4 (PR 167): (Code Coverage 84.3%)

| Code Review Done | ✔ |

| Testing |  |

| Extra Credit |  |
| Given at the instructor’s discretion for good work beyond the minimum assigned work. | 5 |
|- Provide a means to iterate over all the settings on a provider rather than only pulling out specific ones by name.  | ❌✔ |
|- Refactor the test classes that test through `IConfig` into a base class so that the same tests can be used to test multiple. | ❌✔ |
|- Create an IConfigGroup that returns a set of setting based on wildcard filters (rather than just a single setting) | ❌✔ |
|- Provide a mock instance for `System.Console`.  (You do not need to test method invocations of `System.Console`.) | ❌✔ |
|- Create and implement an `IConfigAsync` interface (we are unlikely to cover async this quarter). | ❌✔ |

PR #168, opened 2019-10-27
WIP

jacob-berger:
Trying to figure out what I'm doing before work. Have added FileConfig and EnvironmentConfig, working on SampleApp


Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,56 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Configuration
+{
+    public class EnvironmentConfig : IConfig
+    {
+        private List<string> valueNames;
```
MarkMichaelis: Fields should be specified in _PascalCase, or switched to use an auto property

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,56 @@
...
+        public EnvironmentConfig()
+        {
+            valueNames = new List<string>();
+        }
+
+        public bool GetConfigValue(string name, out string? value)
+        {
+            if (name is null)
+            {
+                throw new ArgumentNullException("Name of variable cannot be null");
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,56 @@
...
+            {
+                return true;
+            }
+        }
+
+        public bool SetConfigValue(string name, string? value)
+        {
+            if (name is null)
+            {
+                throw new ArgumentNullException("Name of variable cannot be null");
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Configuration
+{
+    public class FileConfig
+    {
+        private string path;
```
MarkMichaelis: Fields should be specified in _PascalCase, or switched to use an auto property

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,56 @@
...
+                return false;
+            } else
+            {
+                valueNames.Add(name);
+                Environment.SetEnvironmentVariable(name, value);
+                return true;
+            }
+        }
+
+        ~EnvironmentConfig()
```
MarkMichaelis: If you are going to have a finalizer, you need to support `IDisposable`.

MarkMichaelis: If you are going to have a finalizer, you need to support `IDisposable`.

In this case, I don't believe it is necessary because SetEnvironmentVariable does not persist across runs of your process.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,70 @@
...
+        public FileConfig(string path)
+        {
+            if (string.IsNullOrEmpty(path))
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+            this.path = path;
+        }
+
+        public string GetPath()
```
MarkMichaelis: Use a property, a read only property in this case, rather than a method in C#.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,70 @@
...
+                throw new ArgumentException("Value invalid", nameof(value));
+            }
+
+            using (StreamWriter sw = new StreamWriter(path, true))
+            {
+                sw.WriteLine($"{name}={value}");
+            }
+        }
+
+        public List<Tuple<string, string?>> ReadConfig()
```
MarkMichaelis: Why use Tuple rather than ValueTuple and the associated C# tuple syntax?

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,70 @@
...
+                throw new ArgumentNullException(nameof(name));
+            } else if (string.IsNullOrEmpty(value))
+            {
+                throw new ArgumentNullException(nameof(value));
+            } else if (!CheckValidArguments(name, value))
+            {
+                throw new ArgumentException("Value invalid", nameof(value));
+            }
+
+            using (StreamWriter sw = new StreamWriter(path, true))
```
MarkMichaelis: Glad to see the using statement here.

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,70 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+
+namespace Configuration
+{
+    public class FileConfig
```
MarkMichaelis: You aren't supporting the IConfig interface?

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        public bool GetConfigValue(string name, out string? value)
+        {
+            if (name is null)
+            {
+                throw new ArgumentNullException("Name of variable cannot be null");
+            }
+            value = Environment.GetEnvironmentVariable(name);
+            if (value is null)
```
MarkMichaelis: Can you just return `value is object` ?

Configuration/EnvironmentConfig.cs
MarkMichaelis: Finalizers must be implemented along side IDisposable.

SampleApp/SampleApp.csproj
```diff
@@ -3,6 +3,17 @@
   <PropertyGroup>
```
MarkMichaelis: Where is the associated test project for this assembly?

ConfigurationTests/ConfigurationTests.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
MarkMichaelis: What is this file for?

ConfigurationTests1/ConfigurationTests1.csproj
```diff
@@ -0,0 +1,16 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
MarkMichaelis: What is this file for?

Configuration.Tests/EnvironmentConfigTests.cs
```diff
@@ -0,0 +1,93 @@
...
+            EnvironmentConfig config = new EnvironmentConfig();
+            string? value;
+            //Act
+            config.GetConfigValue(null, out value);
+
+            //Assert
+        }
+
+        [TestMethod]
+        public void GetConfigValueGoodNameNullValue()
```
MarkMichaelis: Glad to see you using your set to test your get and vice versa.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            //Act
+            fileConfig.WriteConfig("", "Value");
+
+            //Assert
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void WriteConfigAllBadParameters()
```
MarkMichaelis: Why do you want this to throw an ArgumentNullException - there are no null parameters that I can see.

Configuration.Tests/FileConfigTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+
+namespace Configuration.Tests
+{
+    [TestClass]
+    public class FileConfigTests
+    {
```
MarkMichaelis: Where to you test for good and bad path values?

MarkMichaelis:
##  (jacob-berger)

## Assignment4

| Build/Compile (Build Succeeded: 0 Error(s), 11 Warning(s))| **10** |


| Assignment  | **10** |
| Create a *`EnvironmentConfig`* configuration class that implements `IConfig` and settings using `Environment.SetEnvironment()` and `Environment.GetEnvironment()` | ✔ |
| **Settings should not persist across process settings (if this occurs you will get a 0)** | ✔ |
| Implement an `FileConfig` that reads/writes settings to a config.settings file in the same directory as the application.  Settings are stored as <name>=<value>.  (You can use String.Split to retrieve individual values). | ❌✔ |
| Do not allow null, empty string, or names with spaces or '='.  Feel free to document other assumptions using unit tests. | ✔ |
| Choose simplicity over than complexity. | ✔ |
| Implement **SampleApp** to print out all the configuration settings based on hard coded values for config (iterating over the values is not supported by the interface). | ✔ |
| Have the **SampleApp** print out a hard coded list of settings to the console.  You do not need to unit test invocations of `System.Console` methods. | ✔ |
| Unit test the **SampleApp** using a `MockConfig` class (implementing `IConfig`) that uses settings stored in memory (such as an array or a `List<T>`) | ❌ |
| Use data driven testing to work with multiple values against the config API. | ❌ |
| Add code analysis to **SampleApp** address all warnings.  It is okay to appropriately disable rules such as  localization/culture rules. **Note: This is true for projects in all assignments going forward.*** | ❌ |
| Nullablility should be enabled at the project level for all projects and all warnings should be handled without disabling them.  Also, be sure to check for null when appropriate.  **Note: This is true for all assignments going forward.*** | ❌ |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |
| CodeReview Performed | **-5** |


| Configuration | Enable | Assignment4.sln |  | CS1701, CS1702, CS2008 | 4 |
| Configuration.Tests | Enable | Assignment4.sln |  | CS1701, CS1702 | 4 |
| ConfigurationTests | Disable |  |  | CS1701, CS1702, CS2008 | 4 |
| ConfigurationTests1 | Disable |  |  | CS1701, CS1702, CS2008 | 4 |
| SampleApp | Enable | Assignment4.sln |  | CS1701, CS1702 | 4 |


| Testing | **10** |
| Test Run FAILED: Total tests=12, Passed=11, Failed=1, Skipped=0 | **48%** |
| Most assigned work is tested or contain some errors. | **0** |


| Extra Credit | **0** |
| | Provide a means to iterate over all the settings on a provider rather than only pulling out specific ones by name. |  ❌ |
| Refactor the test classes that test through `IConfig` into a base class so that the same tests can be used to test multiple. | ❌ |
| Create an IConfigGroup that returns a set of setting based on wildcard filters (rather than just a single setting) | ❌ |
| Provide a mock instance for `System.Console`.  (You do not need to test method invocations of `System.Console`.) | ❌ |
| Create and implement an `IConfigAsync` interface (we are unlikely to cover async this quarter). | ❌ |

PR #169, opened 2019-10-27
Assignment4 WIP

jake-mcg:


Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace Configuration
+{
+    public class EnvironmentConfig : IConfig
+    {
+
+        private List<string> keys;
```
AaronSauther: I like the idea of making a list of keys! Though i'm not sure the directions specified that we should do something like that. I only say this because the environment variables already exist and have set names, so the list may only serve to let you know what settings you have changed. Nice code though!

MarkMichaelis:
## Assignment 4 (PR 169): (92% code coverage)

| Code Review Done | ✔ |

| Testing |  |

| Extra Credit |  |

PR #173, opened 2019-10-27
HWAssign4

TatyanaHubbard:
Finished config files

### DIFF
--- a/Configuration.Tests/EnvironmentConfigTests.cs
+++ b/Configuration.Tests/EnvironmentConfigTests.cs
@@ -1,0 +1,46 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+using System.Text;
+
+namespace Configuration.Tests
+{
+    [TestClass]
+    public class EnvironmentConfigTests
+    {
+        [DataTestMethod]
+        [DataRow("myName", "value")]
+        public void EnvironmentConfig_SetConfigValue_ReturnsTrue(string name, string value)
+        {
+            IConfig environmentConfig = new EnvironmentConfig();
+            
+            Assert.IsTrue(environmentConfig.SetConfigValue(name, value));
+        }
+
+        
+        [DataTestMethod]
+        [DataRow("myName", "value")]
+        [DataRow("Name","ThisValue")]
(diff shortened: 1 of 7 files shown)


Configuration/FileConfig.cs
```diff
@@ -0,0 +1,66 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Configuration
+{
+    public class FileConfig
```
Fervidusletum: I think you should implement the `IConfig` interface on this class, so that it follows the same conventions of the other config classes in this assignment.

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,59 @@
...
+                Environment.SetEnvironmentVariable(name, value);
+                return true;
+            }
+        
+            return false;
+        }
+
+        public static bool IsValidInput(string name, string? value)
+        {
+            if (string.IsNullOrEmpty(name) || string.IsNullOrEmpty(value))
```
Fervidusletum: Passing a null value to `Environment.SetEnvironmentVariable(name, value)` will delete that key from the environment variables, so it may be desirable to allow a null value (assuming the key is valid) to be passed to the function.

Configuration.Tests/EnvironmentConfigTests.cs
```diff
@@ -0,0 +1,45 @@
...
+            Assert.AreEqual(value, myValue);
+        }
+
+        [DataTestMethod]
+        [DataRow(null, "value")]
+        [DataRow("Name", null)]
+        [DataRow(" ", "value")]
+        [DataRow("Name", "=")]
+        [DataRow("=", "value")]
+        public void EnvironmentConfig_SetConfigValue_ReturnFalse(string name, string value)
```
Fervidusletum: you can use `[DataRow("=", "value", DisplayName = "Invalid key: =")]` to name the particular test being run, which makes it a lot easier to tell what's failing at a glance.

MarkMichaelis:
## Assignment 4 (PR 173): (57% code coverage with no SamleApp.Tests project)

| The assignment misses several required elements. The solution contains errors. | 5 |
| Implement IConfig in FileConfig and MockConfig | ❌ |
| Code Review Done | ✔ |

| Testing |  |

| Extra Credit |  |

PR #176, opened 2019-10-28
Assignment 4

JerettLatimer:
WIP


Configuration/FileConfig.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Configuration
+{
+    public class FileConfig : IConfig
```
samrnixon: Does FileConfig have to implement IConfig? Is it just so EnvironmentConfig and FileConfig follow the same "guidelines"?

twofingerrightclick: Yes FileConfig should be implemented as a IConfig user. So yes, EnvironmentConfig and FileConfig will behave the same when using them because they have the same "guidelines" or "API".

Configuration/FileConfig.cs
```diff
@@ -0,0 +1,47 @@
...
+                }
+            }
+            value = null;
+            return false;
+        }
+
+        public bool SetConfigValue(string name, string? value)
+        {
+            if (string.IsNullOrEmpty(name) || string.IsNullOrEmpty(value) || name.Contains(" ") || value.Contains(" ")
+                || name.Contains("=") || value.Contains("="))
```
twofingerrightclick: These logic checks make sense here, but do not reveal what was wrong. Perhaps use the same exceptions and rules as Enviroment.SetEnvironmentVariable. And not required but a "null "value" is used to clear the variable. https://docs.microsoft.com/en-us/dotnet/api/system.environment.setenvironmentvariable?view=netframework-4.8

SampleApp/Program.cs
```diff
@@ -1,12 +1,29 @@
...
+
+            List<String> configSettings = new List<String>
+            {
+                "USERPROFILE",
+                "PATH",
+                "TEMP",
+                "HOME"
+            };
+
+            foreach(string setting in configSettings)
```
twofingerrightclick: Thank you. I was so confused as to what Environment Variables were meant to be printed. Just print some from Windows!

Configuration/FileConfig.cs
twofingerrightclick: It says "print out all the configuration settings"... does that mean we need to get all the Windows Environment Variables? I mean the few you used are reasonable enough.

SampleApp/Program.cs
twofingerrightclick: It says "print out all the configuration settings"... does that mean we need to get all the Windows Environment Variables? I mean the few you used are reasonable enough to demonstrate it works.

JerettLatimer: To be honest, I'm not sure. I just did what I think was being asked.

MarkMichaelis:
## Assignment 4 (PR 176): (86% code coverage but not SampleApp.Tests project)

| The assignment misses several required elements. The solution contains errors. | 5 |
| EnvironmentConfig  (with vallation of null, empty string, or names with spaces or '=') | ❌ |
| Code Review Done | ✔ |

| Testing |  |

| Extra Credit |  |

PR #177, opened 2019-10-28
Assignment 4

Fervidusletum:
Finished, includes extra credit part 2 as `BaseConfig`.


Fervidusletum:
@MarkMichaelis Thanks, you made a lot of good points.

[EnvironmentConfig.cs#L21](https://github.com/IntelliTect-Samples/EWU-CSCD371-2019-Fall/blob/7316700405ba24e5692ff852f22a72c7cae2e87d/Configuration/EnvironmentConfig.cs#L21)  The reason I'm using the `DeleteAllConfigs()` method and the finalizer, is that I wanted to be absolutely sure that I didn't leave any environment variables behind, resulting in a zero on the assignment.

MarkMichaelis:
## Assignment 4 (PR 177): (75% code coverage)

| Code Review Done | ✔ |

| Testing |  |
| All assigned work is tested | 2 |

| Extra Credit |  |
| Given at the instructor’s discretion for good work beyond the minimum assigned work. | 5 |
|- Refactor the test classes that test through `IConfig` into a base class so that the same tests can be used to test multiple. | ✔ |

PR #179, opened 2019-10-28
Assignment4

dbosco051394:
Had some trouble and I'm not totally sure I did it right. A little light on tests at the moment, will add more soon. My tests at least confirm that the config classes are doing their job.


dnisyndrum:
I believe you are missing several test cases for EnvironmentConfig and FileConfig.

MarkMichaelis:
## Assignment 4 (PR 179): (78% Code Coverage)

| Add code analysis to **SampleApp** address all warnings.  It is okay to appropriately disable rules such as  localization/culture rules. | ❌ |
| Nullable enabled for all projects | ❌ |
| Code Review Done | -5 |

| Testing | 15  |

| Extra Credit | 0 |
| Given at the instructor’s discretion for good work beyond the minimum assigned work. | 10 |

PR #180, opened 2019-10-28
Assignment4

twofingerrightclick:
Halfway through the fileConfig


Configuration/MockConfig.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                return false;
+            }
+            else
+                return true;
+        }
+
+        public bool SetConfigValue(string name, string? value)
+        {
+            IConfigUtilities.CheckValidConfigInput(name, value);
```
twofingerrightclick: set is always a true operation unless input is bad? Trying to figure out how to use the bool.

SampleApp/Program.cs
```diff
@@ -1,12 +1,32 @@
...
 
 namespace SampleApp
 {
-    class Program
+    class Program 
     {
         static void Main()
         {
-            Console.WriteLine("Hello World!");
+            Console.WriteLine("Current Enviroment Config Values: ");
```
twofingerrightclick: Is this what we are supposed to do here? This is code I copied from someone else to see how it worked.

kenny2892:
On the FileConfigTests, you could make the testFilePath into a static variable since it never changes. Other than that, looks good.

MarkMichaelis:
## Assignment 4 (PR 180): (85.5 % code coverage but no SampleApp.Tests project)

| The assignment misses several required elements. The solution contains errors. | 20 |
| Nullable enabled for all projects | ❌ |
| Code Review Done | ✔ |

| Testing | 15 |

| Extra Credit |  |
|- Provide a means to iterate over all the settings on a provider rather than only pulling out specific ones by name.  | ✔ |
|- Refactor the test classes that test through `IConfig` into a base class so that the same tests can be used to test multiple. | ✔ |
|- Create an IConfigGroup that returns a set of setting based on wildcard filters (rather than just a single setting) | ✔ |
|- Provide a mock instance for `System.Console`.  (You do not need to test method invocations of `System.Console`.) | ✔ |
|- Create and implement an `IConfigAsync` interface (we are unlikely to cover async this quarter). | ✔ |

PR #181, opened 2019-10-28
Assignment 4

GamerBah:
WIP. Struggling to understand exactly what I'm supposed to be doing for the assignment, and how... but slowly making my way through it I think.


Configuration.Tests/ConfigTests.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+
+    [TestClass]
+    public class EnvironmentConfigTests
+    {
+
+        [DataTestMethod]
+        [DataRow("key", "value")]
+        [DataRow("another-key", "another-value")]
+        public void EnvironmentConfig_GivenKvp_ReturnsCorrectGetSetValue(string key, string value)
```
dougdoner: I know this is a WIP, but as a suggestion I would separate out the tests for getting and setting configs and then use data-driven tests for each. It's my understanding that we want to keep tests simple and only test one piece of logic at a time.

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+
+    public class EnvironmentConfig : IConfig
+    {
+
+        public bool GetConfigValue(string name, out string value)
+        {
+            value = Environment.GetEnvironmentVariable(name);
+
+            return !(value is null);
```
dougdoner: This works fine, I handled it a little differently using (value is string)

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,24 @@
...
+        }
+
+        public bool SetConfigValue(string name, string value)
```
dougdoner: value should be nullable

Configuration.Tests/ConfigTests.cs
GamerBah: That's a good point. Was struggling to find ways to make the tests seem useful because it's so straightforward but I guess I was thinking into it too much.

Configuration/EnvironmentConfig.cs
GamerBah: Good catch! Rider's auto-generated method seems to miss nullables.

GamerBah:
Finished. Not sure if I did the `MockConfig.cs` and `Program.cs` stuff correctly, the assignment specifications seemed a little vague to me with what I was supposed to do.

MarkMichaelis:
## Assignment 4 (PR 181): (94% code coverage but no SampleApp.Tests project.)

| The assignment misses several required elements. The solution contains errors. | 10 |
| EnvironmentConfig  (with vallation of null, empty string, or names with spaces or '=') | ❌ |
| Nullable enabled for all projects | ✔ |
| Code Review Done | ✔ |

| Testing | 15 |

| Extra Credit | 0 |

PR #188, opened 2019-10-29
Assignment 4 solution

markdietzler:
Took me a while to figure out how to rebase Assignment 4.

### DIFF
--- a/Configuration.Tests/MockConfigTests.cs
+++ b/Configuration.Tests/MockConfigTests.cs
@@ -1,0 +1,131 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Configuration.Tests
+{
+    [TestClass]
+    public class MockConfigTests
+    {
+        [DataTestMethod]
+        [DataRow("")]
+        [DataRow(" ")]
+        [DataRow("       ")]
+        [DataRow("=")]
+        [DataRow(null)]
+        [ExpectedException(typeof(ArgumentException))]
+        public void Get_Invalid_Name_Thrown_Exception(string name)
+        {
+            IConfig mock = new MockConfig();
+            SetValidDataForConfig(mock);
+
+            string? value = "";
+            mock.GetConfigValue(name, out value);
+        }
+
+        [DataTestMethod]
+        [DataRow("", " ")]
+        [DataRow(" ", " ")]
+        [DataRow("       ", " ")]
+        [DataRow("=", " ")]
+        [DataRow(null, " ")]
+        [ExpectedException(typeof(ArgumentException))]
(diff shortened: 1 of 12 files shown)


Configuration.Tests/MockConfigTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        public void Get_Correct_Return()
+        {
+            IConfig mock = new MockConfig();
+            SetValidDataForConfig(mock);
+
+            string? value = "";
+            Assert.IsTrue(mock.GetConfigValue("Example0", out value));
+            Assert.IsTrue(mock.GetConfigValue("Example1", out value));
+            Assert.IsTrue(mock.GetConfigValue("Example2", out value));
+        }
```
NikolasGraham: Instead of having several Asserts in the test, it might be better to simply put those values in a DataRow.

Configuration.Tests/MockConfigTests.cs
```diff
@@ -0,0 +1,131 @@
...
+            string? value = "";
+            mock.GetConfigValue("Example3", out value);
+            Assert.AreEqual("", value);
+
+            mock.GetConfigValue("Example6", out value);
+            Assert.AreEqual("", value);
+
+            mock.GetConfigValue("E", out value);
+            Assert.AreEqual("", value);
+        }
```
NikolasGraham: Same thing down here and the test below this, DataRows are very useful.

Configuration/EnvironmentConfig.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            if (key is null)
+                throw new ArgumentException($"The name: {key} is null!");
+
+            else if (string.IsNullOrEmpty(key))
+                throw new ArgumentException($"The name: {key} is empty and therefore invalid.");
+
+            else if (string.IsNullOrWhiteSpace(key))
+                throw new ArgumentException($"The name: {key} is just whitespace therefore invalid.");
+
```
NikolasGraham: For right here, when you check for null or whitespace, you throw the same exception and only return "is just whitespace". I feel like this check would never help you figure out if there is a null somewhere, as there isn't an indication that it is in fact null.

Configuration.Tests/MockConfigTests.cs
markdietzler: I briefly considered creating a utility class with static string constants for all the testing, but then I thought about all the typing, and reconsidered. :D

I will do that next time.

MarkMichaelis:
## Assignment 4 (PR 188): (87% code coverage)

| Nullable enabled for all projects | ✔ |
| Code Review Done | ✔ |

| Testing | 20 |

| Extra Credit | 0 |

PR #191, opened 2019-10-29
Began assignment 5

TatyanaHubbard:
Began Assignment 5

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,68 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using Newtonsoft.Json;
+using System.IO;
+using System.Collections.Generic;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoader_CreatWithNullStream()
+        {
+            DataLoader data = new DataLoader(null);
+            data.Dispose();
+        }
+
+        [TestMethod]
+        public void DataLoader_Save_Success()
+        {
+            MemoryStream memoryStream = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            List<Mailbox> mailboxes = new List<Mailbox>();
+
+            mailboxes.Add(new Mailbox(Size.Small, (1, 1), new Person("Hubbard", "Tatyana")));
+            dataLoader.Save(mailboxes);
+            dataLoader.Dispose();
+        }
+
+
+        // Something weird is going on with this test method, but everything else works
+        [TestMethod]
+        public void DataLoader_Load_Success()
+        {
+            MemoryStream memoryStream = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            List<Mailbox> mailboxes = new List<Mailbox>();
+            List<Mailbox> returnMailboxes;
+
+            mailboxes.Add(new Mailbox(Size.Small, (1, 1), new Person("Hubbard", "Tatyana")));
+            mailboxes.Add(new Mailbox(Size.Medium, (1, 3), new Person("Hubbard", "Tatyana")));
+            mailboxes.Add(new Mailbox(Size.Large, (1, 5), new Person("Hubbard", "Tatyana")));
+            mailboxes.Add(new Mailbox(Size.Undeclared, (1, 7), new Person("Hubbard", "Tatyana")));
+
+            dataLoader.Save(mailboxes);
+            returnMailboxes = dataLoader.Load();
+
+            Assert.AreEqual(mailboxes.Count, returnMailboxes.Count);
+
+            dataLoader.Dispose();
+        }
+
+        [TestMethod]
+        public void DataLoader_Load_ThrowsException()
+        {
+            MemoryStream memoryStream = new MemoryStream();
+            using (StreamWriter sw = new StreamWriter(memoryStream, leaveOpen: true))
+            {
+                sw.Write(new Mailbox(Size.Medium, new ValueTuple<int,int>(1, 1), new Person("Hubbard", "Tatyana")));
+            }
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            Assert.IsNull(dataLoader.Load());
+            dataLoader.Dispose();
+        }
+    }
+}
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -7,6 +7,7 @@ namespace Mailbox
     {
         public DataLoader(Stream source)
         {
+
         }
 
         public List<Mailbox> Load()
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -8,6 +8,7 @@ namespace Mailbox
     public class DataLoader : IDisposable
     {
         private Stream Source { get; set; }
+        private bool StreamDisposed = false;
         public DataLoader(Stream source)
         {
             if(!source.Equals(null))
@@ -22,30 +23,68 @@ namespace Mailbox
 
         public List<Mailbox> Load()
         {
+            Source.Position = 0;
             List<Mailbox> mailboxes = new List<Mailbox>();
-            //Source.Seek
 
-            using (StreamReader sr = new StreamReader(Source))
+            try
             {
-                //Check if JSON
+                using (StreamReader sr = new StreamReader(Source))
+                {
+                    while(!sr.EndOfStream)
+                    {
+                       string? line = sr.ReadLine();
+                       Mailbox newMailbox = JsonConvert.DeserializeObject<Mailbox>(line);
+                       mailboxes.Add(newMailbox);
+                    }
 
-                
-            };
-
-            return mailboxes;
+                    return mailboxes;
+                };
+            } 
+            catch(JsonReaderException)
+            {
+                return null;
+            }
         }
 
         public void Save(List<Mailbox> mailboxes)
         {
-            using (StreamWriter sw = new StreamWriter(Source))
-            {
+            Source.Position = 0;
 
+            try
+            {
+                using (StreamWriter sw = new StreamWriter(Source))
+                {
+                    foreach(Mailbox mailbox in mailboxes)
+                    {
+                        sw.WriteLine(JsonConvert.SerializeObject(mailbox));
+                    }
+                }
+            }
+            catch (JsonReaderException)
+            {
+                return;
             }
         }
 
         public void Dispose()
         {
-            Source.Dispose();
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if(!StreamDisposed)
+            {
+                Source.Dispose();
+            }
+
+            StreamDisposed = true;
+        }
+
+        ~DataLoader()
+        {
+            Dispose(false);
         }
     }
 }
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -1,22 +1,51 @@
-﻿using System.Collections.Generic;
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private Stream Source { get; set; }
         public DataLoader(Stream source)
         {
-
+            if(!source.Equals(null))
+            {
+                Source = source;
+            }
+            else
+            {
+                throw new ArgumentNullException("DataLoader's Stream cannot be null");
+            }
         }
 
         public List<Mailbox> Load()
         {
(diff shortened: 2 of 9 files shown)


Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,28 @@
...
+            }
+
+            Size = size;
+            Owner = owner;
+            Location = location;
+        }
 
+        public override string ToString()
+        {
+            return $"{Owner.FirstName} {Owner.LastName}'s {Size} Mailbox is located at {Location.Item1} and {Location.Item2}";
```
mmwoodfo: Why not use owner.ToString() here?

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
         public DataLoader(Stream source)
         {
+            if(!(source is null))
+            {
+                Source = source;
+            }
+            else
+            {
+                throw new ArgumentNullException("DataLoader's Stream cannot be null");
+            }
```
dnisyndrum: You could probably put this all into one line. 

Source = source ?? throw new ArgumentNullException(nameof(source));

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
         {
+            Source.Position = 0;
+
+            List<Mailbox> mailboxes = new List<Mailbox>();
+
+            try
+            {
+                using (StreamReader sr = new StreamReader(Source))
+                {
+                    while(!sr.EndOfStream)
```
AustinHall7798:
```suggestion
                    return JsonConvert.DeserializeObject<List<MailBox>>(reader.ReadToEnd());
```
The try block can be significantly shortened by reading the lines straight into a MailBox list and returning. It will also fix the build errors

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream Source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
+        private bool StreamDisposed = false;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
         public DataLoader(Stream source)
         {
+            if(!(source is null))
+            {
+                Source = source;
+            }
+            else
+            {
+                throw new ArgumentNullException("DataLoader's Stream cannot be null");
```
Keboo: This constructor takes in the parameter name, not a message. 
```suggestion
                throw new ArgumentNullException(nameof(source));
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,40 @@
...
+    {
+        public Size Size { get; set; }
+        public ValueTuple<int, int> Location { get; set; }
+        public Person Owner { get; set; }
+
+        public Mailbox(Size size, ValueTuple<int, int> location, Person owner)
+        {
+            if(owner is null)
+            {
+                throw new ArgumentNullException("The owner cannot be null");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+    public class Person : IEquatable<Person>
+    {
+        public string LastName { get; set; }
+        public string FirstName { get; set; }
+
+        public Person(string lastName, string firstName)
+        {
+            if(lastName.Equals(null) || firstName.Equals(null))
+            {
+                throw new ArgumentNullException("The name can not be null");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+namespace Mailbox
 {
+    public class Person : IEquatable<Person>
+    {
+        public string LastName { get; set; }
+        public string FirstName { get; set; }
+
+        public Person(string lastName, string firstName)
+        {
+            if(lastName.Equals(null) || firstName.Equals(null))
```
Keboo: This is not the correct way to check for null. If `lastName` is null, then this will throw a NullReferenceException not an ArgumentException.
```suggestion
            if(lastName is null || firstName is null)
```

Mailbox/Program.cs
```diff
@@ -89,17 +89,51 @@ static void Main(string[] args)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Collections.Generic;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoader_CreatWithNullStream()
```
Keboo:
```suggestion
        public void DataLoader_CreateWithNullStream()
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        {
+            DataLoader data = new DataLoader(null);
+            data.Dispose();
```
Keboo: It appears you are testing the constructor here, so the Dispose call is not needed.

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        [TestMethod]
+        public void DataLoader_Save_Success()
+        {
+            MemoryStream memoryStream = new MemoryStream();
```
Keboo: MemoryStream is `IDisposable` and should be properly cleaned up.
```suggestion
            using MemoryStream memoryStream = new MemoryStream();
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            List<Mailbox> mailboxes = new List<Mailbox>();
+
+            mailboxes.Add(new Mailbox(Size.Small, (1, 1), new Person("Hubbard", "Tatyana")));
+            dataLoader.Save(mailboxes);
+            dataLoader.Dispose();
```
Keboo: This works as long as nothing throws an exception. It would have been better to mark the data with a using statement to ensure it always gets cleaned up; even in the case of an exception being thrown.
```
DataLoader dataLoader = new DataLoader(memoryStream);
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
 
         public List<Mailbox> Load()
         {
+            Source.Position = 0;
+
+            List<Mailbox> mailboxes = new List<Mailbox>();
+
+            try
+            {
+                using (StreamReader sr = new StreamReader(Source))
```
Keboo: This will close the stream. I believe this is the causes of the failing unit tests.
```suggestion
                using (StreamReader sr = new StreamReader(Source, leaveOpen:true))
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
+                return null;
+            }
         }
 
         public void Save(List<Mailbox> mailboxes)
         {
-            
+            Source.Position = 0;
+
+            using (StreamWriter sw = new StreamWriter(Source))
```
Keboo: This will close the stream. I believe this is the causes of the failing unit tests.
```suggestion
            using (StreamWriter sw = new StreamWriter(Source, leaveOpen:true))
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,84 @@
...
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if(!StreamDisposed)
+            {
+                Source.Dispose();
+            }
+
+            StreamDisposed = true;
```
Keboo: This is the correct implementation, but the name of this is a little off. This boolean flag is designed to indicate if _this_ object has been disposed, and is usually named something more like `IsDisposed`.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,40 @@
...
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public ValueTuple<int, int> Location { get; set; }
+        public Person Owner { get; set; }
+
+        public Mailbox(Size size, ValueTuple<int, int> location, Person owner)
```
Keboo: This works but value tuples have a simpler syntax:
```suggestion
        public Mailbox(Size size, (int, int) location, Person owner)
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public class Person : IEquatable<Person>
+    {
```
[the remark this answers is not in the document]
Keboo: Correct, this should be a `struct`

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+        {
+            if (ReferenceEquals(other, null))
+            {
+                return false;
+            }
+
+            return FirstName == other.FirstName && LastName == other.LastName;
+        }
+
+        public static bool operator ==(Person a, Person b) => (a == b);
```
Keboo: This method is recursive on all paths and will cause a stack overflow if you use it.
```
Person a = ...
Person b = ...
bool stackOverflow = a == b;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+                throw new ArgumentNullException("The name can not be null");
+            }
+
+            LastName = lastName;
+            FirstName = firstName;
+        }
+
+        public override bool Equals(object obj)
+        {
+            return obj is Person && Equals((Person)obj);
```
Keboo: You can use pattern matching here to avoid the cast.
```suggestion
            return obj is Person person && Equals(person);
```

Mailbox/Program.cs
```diff
@@ -89,17 +89,51 @@ static void Main(string[] args)
...
+
+            return listNames;
         }
 
         public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y)
         {
             
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                if (mailbox.Location.Equals(new ValueTuple<int,int>(x,y)))
```
Keboo: Simplify tuple syntax
```suggestion
                if (mailbox.Location.Equals((x,y)))
```

Mailbox/Program.cs
```diff
@@ -89,17 +89,51 @@ static void Main(string[] args)
...
         {
-            
+            string listNames = "";
+
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                listNames += mailbox.Owner + ", ";
+            }
+
+            return listNames;
```
Keboo: This method was expected to return a distinct list of owners not just a list of all owners.

Mailbox/Program.cs
```diff
@@ -89,17 +89,51 @@ static void Main(string[] args)
...
+            for(int xx = 1; xx <= 30; xx++)
+            {
+                for(int yy = 1; yy <= 10; yy++)
+                {
+                    
+                    bool isOccupied = mailboxes.GetAdjacentPeople(xx, yy, out HashSet<Person> adjacentPeople);
+
+                    if(!isOccupied && !adjacentPeople.Contains(person))
+                    {
+                        newMailbox = new Mailbox(size, ValueTuple.Create(xx, yy), person);
```
Keboo:
```suggestion
                        newMailbox = new Mailbox(size, (xx, yy), person);
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,40 @@
...
+            Size = size;
+            Owner = owner;
+            Location = location;
+        }
+
+        public override string ToString()
+        {
+            string mySize; 
+
+            if(Size == Size.Undeclared || Size == Size.Premium)
```
Keboo: For the assignment the expectation was that when it was a premium size it appended " Premium" to the string. Though I realize that bullet point was poorly worded.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,40 @@
...
+            {
+                mySize = "";
+            }
+            else
+            {
+                mySize = Size.ToString() + " ";
+            }
+          
 
+            return $"{Owner.ToString()}'s {mySize}Mailbox is located at {Location.Item1} and {Location.Item2}";
```
Keboo:
```suggestion
            return $"{Owner}'s {mySize}Mailbox is located at {Location.Item1} and {Location.Item2}";
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,40 @@
-﻿namespace Mailbox
+﻿using System;
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public ValueTuple<int, int> Location { get; set; }
```
Keboo: If you use the simpler value tuple syntax you can given names to the items in the tuple rather than just using Item1 and Item2.
```suggestion
        public (int X, int Y) Location { get; set; }
```

PR #192, opened 2019-10-30
Assignment5 : mmwoodfo

mmwoodfo:


mmwoodfo:
@Keboo tell me how bad it is in front of everyone, please.

Mailbox/DataLoader.cs
```diff
@@ -1,28 +1,68 @@
...
+                }
+                catch (JsonReaderException)
+                {
+                    return null;
+                }
+                catch (ArgumentNullException)
+                {
+                    return null;
+                }
+            }
```
ashermancinelli: Would it be better to have the `using` inside the try block do you think?

Mailbox/DataLoader.cs
```diff
@@ -1,28 +1,68 @@
...
         {
-            throw new NotImplementedException();
+            List<MailBox> mailboxes = new List<MailBox>();
+            source.Position = 0;
+
+            using (StreamReader sr = new StreamReader(source))
+            {
+                try
+                {
+                    string? jsonMailBox = sr.ReadLine();
```
ashermancinelli: This is just reading one line, correct? I believe you want to iterate over all the lines in the source rather than just deserializing the first line.

Keboo:
Pull request should be against Assignment5 branch not master.

Mailbox/Program.cs
```diff
@@ -69,7 +68,7 @@ static void Main(string[] args)
                     case 4:
                         Console.WriteLine("Enter box number as x,y");
                         string boxNumber = Console.ReadLine();
-                        string[] parts = boxNumber?.Split(',');
+                        string[]? parts = boxNumber?.Split(',');
```
Cybotive: should parts be nullable? does .Split() ever return null?

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,68 @@
...
 
-        public void Save(List<Mailbox> mailboxes)
+        public void Save(List<MailBox> mailboxes)
         {
-            
+            source.Position = 0;
+
+            using (StreamWriter sw = new StreamWriter(source))
+            {
+                foreach (MailBox mailbox in mailboxes)
```
Keboo:
```suggestion
string json = JsonConvert.SerializeObject(mailboxes);
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,68 @@
...
+                    string? jsonMailBox = sr.ReadLine();
+                    MailBox mailbox = JsonConvert.DeserializeObject<MailBox>(jsonMailBox);
+                    mailboxes.Add(mailbox);
+                    return mailboxes;
+                }
+                catch (JsonReaderException)
+                {
+                    return null;
+                }
+                catch (ArgumentNullException)
```
Keboo: In general you should never catch `ArgumentNullException`

Mailbox/Program.cs
```diff
@@ -1,21 +1,20 @@
...
 namespace Mailbox
 {
-    class Program
+    internal class Program
     {
         private const int Width = 50;
         private const int Height = 10;
 
-        static void Main(string[] args)
+        private static void Main(string[] args)
```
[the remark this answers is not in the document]
mmwoodfo: It runs as a private method, I think it was changed to private from a code cleaner I have.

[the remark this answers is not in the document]
Keboo: Once the class videos finish uploading I reference this in the second part when i touch on access modifiers :)

Mailbox/Program.cs
mmwoodfo: No, .Split() never returns null. I put the '?' there because I was getting a warning, but I think the better option would be to just remove the '?' from boxNumber, because readline should not return null either. So that is probably what I will do to fix the warning.

Mailbox/Program.cs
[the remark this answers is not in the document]
twofingerrightclick: I will always remember this c# fact. Haha

Mailbox/Person.cs
```diff
@@ -1,4 +1,38 @@
...
+            LastName = lastName ?? throw new ArgumentNullException(nameof(lastName));
+        }
+
+        public bool Equals([AllowNull] Person other)
+        {
+            return FirstName == other.FirstName &&
+            LastName == other.LastName;
+        }
+
+        public override bool Equals(object? obj)
```
twofingerrightclick: This method is recursive. If it gets called with an Object it will call itself until it stack overflows. This is the code from the lecture on Operator overrides, so I tried it as well and discovered that.

mmwoodfo: Oh, that is good to know. I need to read up on this more as I have had some difficulty understanding exactly how it works. Hence why it's implemented incorreclty. :) Thank you for letting me know! I'll try to fix this whenever I get home from work.

twofingerrightclick: Cool. Same here. Kevin (Student Kevin) showed me that "obj as Person?" is the same expression as "Nullable<Person> p = new Nullable<Person>();"

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,83 @@
...
+
+            using (StreamWriter sw = new StreamWriter(Source, leaveOpen: true))
+            {
+                foreach (MailBox mailbox in mailboxes)
+                    sw.WriteLine(JsonConvert.SerializeObject(mailbox));
+            }
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,38 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
```
Keboo: Person should override the equality (==) and inequality (!=) operators.

Mailbox/Program.cs
```diff
@@ -1,21 +1,20 @@
...
 
 namespace Mailbox
 {
-    class Program
+    internal class Program
     {
         private const int Width = 50;
         private const int Height = 10;
 
-        static void Main(string[] args)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,119 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace Mailbox.Tests
+{
+    [TestClass()]
+    public class DataLoaderTests
+    {
+        private List<MailBox> testMailboxes = new List<MailBox>()
```
Keboo: Fields should be specified in _PascalCase, or switched to use an auto property

Mailbox.Tests/BogusMailBoxTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            Assert.AreEqual(expectedString, receivedString);
+        }
+
+        [TestMethod()]
+        public void ToStringTest_PremiumSize()
+        {
+            //Arrange
+            Person owner = new Person(new Bogus.DataSets.Name().FirstName(), new Bogus.DataSets.Name().LastName());
+            MailBox mailBox = new MailBox(Sizes.MediumPremium, (1, 3), owner);
+            string expectedString = $"Mailbox Owner: {owner.ToString()}, Location: x = 1, y = 3, BoxSize: MediumPremium";
```
Keboo: `.ToString()` is not needed inside of string interpolation.
```suggestion
            string expectedString = $"Mailbox Owner: {owner}, Location: x = 1, y = 3, BoxSize: MediumPremium";
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            Assert.ThrowsException<ArgumentNullException>(() => new DataLoader(null));
+#pragma warning restore CS8625 // Cannot convert null literal to non-nullable reference type.
+
+        }
+
+        [TestMethod()]
+        public void LoadTest_ExceptionReturnsNull()
+        {
+            //Arrange
+            var ms = new MemoryStream();
```
Keboo: Memory stream is `IDisposable` and should be properly cleaned up at the end of the method.
```suggestion
            using var ms = new MemoryStream();
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,119 @@
...
+            }
+            ms.Position = 0;
+
+            DataLoader dataLoader = new DataLoader(ms);
+
+
+            //Act & Assert
+            Assert.IsNull(dataLoader.Load());
+
+            dataLoader.Dispose();
```
Keboo: This works but it would have been easier to just apply the `using` to the data loader.
```
using DataLoader dataLoader = new DataLoader(ms);
```

Mailbox.Tests/MailBoxTests.cs
```diff
@@ -0,0 +1,81 @@
...
+{
+    [TestClass()]
+    public class MailBoxTests
+    {
+        [TestMethod]
+        public void MailBoxConstrucotrTest_SizePremium()
+        {
+            //Arrange
+            Sizes size = Sizes.Premium;
+            ValueTuple<int, int> location = new ValueTuple<int, int>(15, 5);
```
Keboo: Tuple syntax can be simplified:
```suggestion
            (int, int) location = (15, 5);
```

Mailbox.Tests/MailBoxTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+            //Act
+            MailBox mailbox = new MailBox(size, location, person);
+
+            //Assert
+            Assert.AreEqual(mailbox.MailSize, Sizes.SmallPremium);
+        }
+
+        [TestMethod()]
+        public void ToStringTest_NotPremiumSize()
```
Keboo: As a general naming, you can omit the word "Test" from unit test method names.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,83 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        public Stream Source { get; }
```
Keboo: This could be made private.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,31 @@
...
+
+        public override string ToString()
+        {
+            string mailboxString = $"Mailbox Owner: {Owner.ToString()}, Location: x = {Location.X}, y = {Location.Y}";
+            if (MailSize == Sizes.Unspecfied)
+            {
+                return mailboxString;
+            }
 
+            return mailboxString + $", BoxSize: {MailSize}";
```

Mailbox/Program.cs
```diff
@@ -89,17 +88,49 @@ static void Main(string[] args)
...
+            }
+
+            string ownersList = "";
+
+            foreach (MailBox mailbox in mailboxes)
+            {
+                ownersList += mailbox.Owner.ToString() + "\n";
+            }
+
+            return ownersList;
```

Mailbox/DataLoader.cs
mmwoodfo: I am really going to have to start getting out of my java habits. This will hopefully be the last time you make that comment on my PR.

Mailbox.Tests/MailBoxTests.cs
mmwoodfo: So when I create the unit tests I just right click on my class and generate all the unit tests for each method (VS defaults to naming everything 'method'Test). I suppose I can specify that I don't want them to be named test when I generate them in the little options menu that pops up. Is it bad practice to have the word test as part of the method name? Or just a suggestion that I don't need it there?

PR #194, opened 2019-10-30
Assignment5

dnisyndrum:


Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,28 @@
...
+    {
+        public Size MailboxSize { get; set; }
+        public ValueTuple<int, int> Location { get; set; }
+        public Person Owner { get; set; }
+
+        public Mailbox(Size boxSize, ValueTuple<int, int> location, Person newOwner)
+        {
+            if(MailboxSize is 0 || location.Item1 == -1 || location.Item2 == -1)
+            {
+                throw new ArgumentNullException("Improper mailbox information.");
```
mmwoodfo: This might be better as an `ArgumentException`. Also, what is the purpose of checking for -1?

dnisyndrum: Thanks, I was not sure which of the two exception to go with. Looking at it, ArgumentException does seem more logical. 
I went with -1 partially as a placeholder for now, but I also had the idea that it might be set to -1 if a location had not been assigned. That'll likely be changed shortly.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,38 @@
...
+﻿using System;
+
+namespace Mailbox
 {
 
+    public class Mailbox
+    {
+        public Size _Size { get; set; }
+        public (int Row, int Column) _Location { get; set; }
+        public Person _Owner { get; set; }
```
RSXEnthusiast:
```suggestion
        public Person Owner { get; set; }
```

Believe these are just supposed to be regular camel case, I made the same mistake before Kevin set me straight.

As I understand it Properties should be normal camel case and fields should use an underscore. [Here's a link](https://stackoverflow.com/questions/295104/what-is-the-difference-between-a-field-and-a-property) to someone explaining the difference much better than I could.

Mailbox/Person.cs
```diff
@@ -1,4 +1,41 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        string _FirstName;
+        string _LastName;
```
RSXEnthusiast:
```suggestion
        private string _LastName;
```

Should these be private?

Mailbox.Tests/DataLoader.Tests.cs
```diff
@@ -0,0 +1,68 @@
...
+        }
+
+        [TestMethod]
+        public void DataLoader_Save_Success()
+        {
+            List<Mailbox> boxList = new List<Mailbox>();
+            MemoryStream source = new MemoryStream();
+            DataLoader loader = new DataLoader(source);
+
+            boxList.Add(new Mailbox(Size.Small, new ValueTuple<int, int>(1, 1), new Person("Some", "One")));
```
Keboo: Value tuple can be simplified.
```suggestion
            boxList.Add(new Mailbox(Size.Small, (1, 1), new Person("Some", "One")));
```

Mailbox.Tests/Mailbox.Tests.cs
```diff
@@ -0,0 +1,49 @@
...
+            Assert.AreEqual(mailbox.ToString(), "Owner: Dude, The, Location: 1, 1, Size: " + size.ToString());
+        }
+
+        [DataTestMethod]
+        [DataRow(Size.Unset)]
+        [DataRow(Size.Premium)]
+        public void Mailbox_Sizes_EmptyStringWithBadSizes(Size size)
+        {
+            Person theDude = new Person("The", "Dude");
+            ValueTuple<int, int> location = new ValueTuple<int, int>(1, 1);
```
Keboo:
```suggestion
            (int, int) location = (1, 1);
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,81 @@
...
 
-        public List<Mailbox> Load()
+        public List<Mailbox>? Load()
         {
+            List<Mailbox> mailboxes = new List<Mailbox>();
+            Source.Position = 0;
+
+            try
+            {
+                using (StreamReader reader = new StreamReader(Source))
```
Keboo: This will cause the stream to get closed.
```suggestion
                using (StreamReader reader = new StreamReader(Source, leaveOpen:true))
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,81 @@
...
+            GC.SuppressFinalize(this);
         }
+
+        protected virtual void Dispose (bool disposing)
+        {
+            if(!StreamDisposed)
+            {
+                Source.Dispose();
+            }
+            StreamDisposed = true;
```
Keboo: This works, but in general this boolean flag is designed to indicate if _this_ object has been disposed. Imagine if this object contained more than just a stream. Typically it is named something like `IsDisposed`.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,33 @@
...
+﻿using System;
+
+namespace Mailbox
 {
 
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public (int Row, int Column) Location { get; set; }
+        public Person Owner { get; set; }
```
Keboo: Properties can be make readonly.
```suggestion
        public Person Owner { get; }
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,33 @@
...
+        public override string ToString()
+        {
+            string output = $"Owner: {Owner.ToString()}, Location: {Location.Item1}, {Location.Item2}";
+
+            if(Size != Size.Unset && Size != Size.Premium)
+            {
+                output += $", Size: {Size.ToString()}";
+            }
+
+            return output;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+            {
+                throw new ArgumentNullException(nameof(LastName));
+            }
+            FirstName = fname;
+            LastName = lname;
+        }
+
+        public bool Equals([AllowNull] Person other)
+        {
+            if(ReferenceEquals(other, null))
```
Keboo: `Person` is a struct (value type) so it can't be null.

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+            {
+                return false;
+            }            
+            return FirstName == other.FirstName && LastName == other.LastName;
+        }
+
+        public override bool Equals(object obj)
+        {
+            return base.Equals(obj);
```
Keboo: This should be calling into `Equals(Person)`.

Mailbox/Program.cs
```diff
@@ -89,17 +89,50 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string allMailboxes = "";
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                allMailboxes += mailbox.ToString() + "\n";
+            }
+            return allMailboxes;
```
Keboo: The expectation on was that this method return a _distinct_ set of owner names. This return all of the mailboxes.

PR #195, opened 2019-10-31
Assignment 5

dougdoner:
@Keboo If this is complete enough for a code review you can use this pull request for tomorrow's class.

### DIFF
--- a/Mailbox/Mailbox.cs
+++ b/Mailbox/Mailbox.cs
@@ -1,4 +1,12 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public Person Owner { get; set; }
+        public (int X, int Y) Location { get; set; }
 
+
+
+    }
 }
--- a/Mailbox/Mailbox.cs
+++ b/Mailbox/Mailbox.cs
@@ -6,7 +6,7 @@
         public Person Owner { get; set; }
         public (int X, int Y) Location { get; set; }
 
-
+        public override string? ToString() => $"Owner: {Owner.ToString()} Location: {Location} Size: {(Size == Size.Default ? "" : Size.ToString())}";
 
     }
 }
--- a/Mailbox/Mailbox.cs
+++ b/Mailbox/Mailbox.cs
@@ -6,7 +6,29 @@
         public Person Owner { get; set; }
         public (int X, int Y) Location { get; set; }
 
-        public override string? ToString() => $"Owner: {Owner.ToString()} Location: {Location} Size: {(Size == Size.Default ? "" : Size.ToString())}";
+        public override string? ToString()
+        {
+            string size;
+            switch (Size)
+            {
+                case Size.Premium:
+                    size = "Premium";
+                    break;
+                case Size.SmallPremium:
+                    size = "Small Premium";
+                    break;
+                case Size.MediumPremium:
+                    size = "Medium Premium";
+                    break;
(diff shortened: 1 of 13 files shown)


Keboo:
We didn't get to this in class but because you had it ready in time I will still throw a review on it by this weekend.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,12 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public Person Owner { get; set; }
+        public (int X, int Y) Location { get; set; }
 
+        public override string? ToString() => $"Owner: {Owner.ToString()} Location: {Location} Size: {(Size == Size.Default ? "" : Size.ToString())}";
```
Keboo: You probably want to review the last bullet point for the Mailbox class. This handles the Default case for Size, but not the premium case.

Mailbox/Person.cs
```diff
@@ -1,4 +1,54 @@
...
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        public string? FirstName { get; set; }
+        public string? LastName { get; set; }
+
+        public bool Equals([AllowNull] Person other)
+        {
+            if (ReferenceEquals(other, null)) return false;
```
Keboo: Since Person is a stuct it is not possible for it to be null.

Mailbox/Person.cs
```diff
@@ -1,4 +1,54 @@
...
+            if (ReferenceEquals(this, other)) return true;
```
Keboo: This causes both `this` and `other` to get boxed. Effectively given them each their own instances so this condition will never be true.

Mailbox/Person.cs
```diff
@@ -1,4 +1,54 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return (FirstName, LastName).GetHashCode();
+        }
+
+        public static bool operator ==(Person left, Person right)
+        {
+            if (ReferenceEquals(left, right)) return true;
```
Keboo: Same as above, because `Person` is a struct and ReferenceEquals take in objects, this will box the parameters and not match.

Mailbox/Mailbox.cs
dougdoner: Thanks for the feedback. I think I've fixed these issues in 6dd221c .

Mailbox/Program.cs
```diff
@@ -89,17 +89,17 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            return "";
```
Keboo: Method to be implemented.

Mailbox/Program.cs
```diff
@@ -89,17 +89,17 @@ static void Main(string[] args)
...
         }
 
         public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y)
         {
-            
+            return "";
```
Keboo: Method to be implemented.

Mailbox/Program.cs
```diff
@@ -89,17 +89,17 @@ static void Main(string[] args)
...
         }
 
         public static Mailbox AddNewMailbox(Mailboxes mailboxes, string firstName, string lastName, Size size)
         {
-            
+            return new Mailbox();
```
Keboo: Method to be implemented.

PR #196, opened 2019-10-31
As5

ashermancinelli:
WIP

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -11,6 +11,47 @@ namespace Mailbox.Tests
     public class DataLoaderTests
     {
         [TestMethod]
+        public void DataLoader_LoadMailbox_Success()
+        {
+            string fileName = Path.GetRandomFileName();
+            Stream fileStream = File.Open(fileName, FileMode.Create);
+            List<Mailbox> mailboxes = new List<Mailbox>()
+            {
+                new Mailbox(Size.Small, (1, 1), new Person("Asher", "Mancinelli")),
+                new Mailbox(Size.Small, (2, 1), new Person("John", "Doe")),
+                new Mailbox(Size.Small, (1, 2), new Person("Jake", "FakeName")),
+            };
+
+            using (var sw = new StreamWriter(fileStream))
+            {
+                foreach (var box in mailboxes)
+                {
+                    sw.WriteLine(JsonConvert.SerializeObject(box));
+                }
+            }
+            
+            fileStream = File.Open(fileName, FileMode.Open);
+            var sut = new DataLoader(fileStream);
+
+            try
+            {
+                List<Mailbox>? loadedMailboxes = sut.Load();
+                var boxPairs = mailboxes.Zip(loadedMailboxes, (a, b) => (a, b));
+
+                Assert.IsNotNull(loadedMailboxes);
+                foreach (var (box, loadedBox) in boxPairs)
+                {
+                    Assert.AreEqual(box, loadedBox);
+                }
+            }
+            finally
+            {
+                File.Delete(fileName);
+                fileStream.Dispose();
+            }
+        }
+
+        [TestMethod]
         public void DataLoader_SaveMailbox_Success()
         {
             string fileName = Path.GetRandomFileName();
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -82,5 +82,15 @@ namespace Mailbox.Tests
                 fileStream.Dispose();
             }
         }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoader_LoadWithNullSource_ThrowsException()
+        {
+            #nullable disable
+            var sut = new DataLoader(null);
+            #nullable enable
+            sut.Load();
+        }
     }
 }
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,38 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        public void DataLoader_SaveMailbox_Success()
+        {
+            string fileName = Path.GetRandomFileName();
+            Stream fileStream = File.Open(fileName, FileMode.Create);
+            var mailboxes = new List<Mailbox>()
+            {
+                new Mailbox(Size.Small, (1, 1), new Person("Asher", "Mancinelli")),
+            }
+        
+            try
+            {
+                var sut = new DataLoader(fileStream);
+
+                sut.Save(mailboxes);
+                string[] lines = File.ReadLines(fileName).ToArray();
+                var testPairs = mailboxes.Zip(lines, (a, b) => (a, b));
+
+                foreach (var (box, line) in testPairs)
+                {
+                    Assert.AreEqual(JsonConvert.SerializeObject(box), line);
+                }
+            }
+            finally
+            {
+                File.Delete(fileName);
+                fileStream.Dispose();
+            }
+        }
+    }
+}
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,4 +1,8 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using System.IO;
+using System.Linq;
+using System;
 
 namespace Mailbox.Tests
 {
@@ -10,10 +14,12 @@ namespace Mailbox.Tests
         {
             string fileName = Path.GetRandomFileName();
             Stream fileStream = File.Open(fileName, FileMode.Create);
-            var mailboxes = new List<Mailbox>()
+            List<Mailbox> mailboxes = new List<Mailbox>()
             {
                 new Mailbox(Size.Small, (1, 1), new Person("Asher", "Mancinelli")),
-            }
+                new Mailbox(Size.Small, (2, 1), new Person("John", "Doe")),
+                new Mailbox(Size.Small, (1, 2), new Person("Jake", "FakeName")),
+            };
         
             try
             {
@@ -23,9 +29,9 @@ namespace Mailbox.Tests
                 string[] lines = File.ReadLines(fileName).ToArray();
                 var testPairs = mailboxes.Zip(lines, (a, b) => (a, b));
 
-                foreach (var (box, line) in testPairs)
+                foreach (var pair in testPairs)
                 {
-                    Assert.AreEqual(JsonConvert.SerializeObject(box), line);
+                    Assert.AreEqual(JsonConvert.SerializeObject(pair[0]), pair[1]);
                 }
             }
             finally
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,5 +1,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using Newtonsoft.Json;
+using System.Collections.Generic;
 using System.IO;
 using System.Linq;
 using System;
@@ -29,9 +30,9 @@ namespace Mailbox.Tests
                 string[] lines = File.ReadLines(fileName).ToArray();
                 var testPairs = mailboxes.Zip(lines, (a, b) => (a, b));
 
-                foreach (var pair in testPairs)
+                foreach (var (box, line) in testPairs)
                 {
-                    Assert.AreEqual(JsonConvert.SerializeObject(pair[0]), pair[1]);
+                    Assert.AreEqual(JsonConvert.SerializeObject(box), line);
                 }
             }
             finally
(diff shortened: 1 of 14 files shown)


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,94 @@
...
 
         public void Save(List<Mailbox> mailboxes)
         {
-            
+            using var sw = new StreamWriter(Source);
+
+            foreach (var mailbox in mailboxes)
+            {
+                sw.WriteLine(JsonConvert.SerializeObject(mailbox));
+            }
```
pnotequalnp: You can just serialize the `List<Mailbox>` directly rather than manually iterating over it. Same goes for deserializing it.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,76 @@
...
+     * The ToString method should include all property members.
+     * When including the Size in the ToString, it should output empty string when the Size 
+     * is the default (0) value, the name of the size with " Premium" it is a premium size,
+     * or simple the size name. For examples "Small" or "Medium Premium".
+     */
+
+    public class Mailbox
+    {
+        public Size MailboxSize;
+        public ValueTuple<int, int> Location;
```
pnotequalnp:
```suggestion
        public (int X, int Y) Location;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,76 @@
...
+        public Person Owner;
+
+        public Mailbox(Size size, ValueTuple<int, int> location, Person owner)
+        {
+            NullCheck("Mailbox Constructor", size);
+            NullCheck("Mailbox Constructor", location);
+            NullCheck("Mailbox Constructor", owner);
```
pnotequalnp: These are all value types and thus cannot be null here.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,76 @@
...
+            MailboxSize = size;
+            Location = location;
+            Owner = owner;
+        }
+
+        public override string ToString()
+        {
+            NullCheck("ToString", MailboxSize);
+            NullCheck("ToString", Location);
+            NullCheck("ToString", Owner);
```
pnotequalnp: Value types again, can't be null.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,76 @@
...
+
+            string s = $"Size: ";
+            s += MailboxSize switch
+            {
+                Size.Small => "",
+                Size.Medium => "Medium ",
+                Size.Large => "Large ",
+                Size.SmallPremium => "Premium ",
+                Size.MediumPremium => "Medium Premium ",
+                Size.LargePremium => "Large Premium ",
```
pnotequalnp: The default `.ToString()` on enums returns the name, perhaps that would be useful here.

Mailbox/Mailboxes.cs
```diff
@@ -59,5 +59,30 @@ public bool GetAdjacentPeople(int x, int y, out HashSet<Person> adjacentPeople)
...
         }
+
+        public Mailbox? GetMailbox(int x, int y)
+        {
+            foreach (var mailbox in this)
+            {
+                if (mailbox.Location == (x, y))
+                    return mailbox;
+            }
+            return null;
```
pnotequalnp: This is equivalent to `return this.FirstOrDefault(mailbox => mailbox.Location == (x, y));`

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
...
+     * Create a new Person struct
+     * A person has a first and last name
+     * Person should implement IEquatable<Person>
+     * Ensure appropriate equality operators are also written
+     * Implement a reasonable ToString method
+     */
+    public struct Person : IEquatable<Person>
+    {
+        public string? FirstName;
+        public string? LastName;
```
pnotequalnp: Why use public nullable fields?

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
...
+
+        public Person(string firstName, string lastName)
+        {
+            if (firstName is null)
+                throw new ArgumentNullException($"{nameof(firstName)} cannot be null.");
```
pnotequalnp: First argument is just the parameter name, second is message.
```suggestion
                throw new ArgumentNullException(nameof(firstName));
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
...
+                throw new ArgumentNullException($"{nameof(lastName)} cannot be null.");
+
+            FirstName = firstName;
+            LastName = lastName;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is Person)
+                return this.Equals((Person)obj);
```
pnotequalnp: Pattern matching makes this more concise.
`if (obj is Person p) return this.Equals(p);`

Mailbox/Program.cs
```diff
@@ -89,17 +98,61 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string displayStr = "";
+            foreach (Mailbox mailbox in mailboxes)
+            {
+                displayStr += mailbox.ToString() + Environment.NewLine;
```
pnotequalnp: This should be done with a `StringBuilder` to avoid excessive concatenation.

Assignment asks for a list of the _owners_, not the mailboxes.

Mailbox/Program.cs
```diff
@@ -89,17 +98,61 @@ static void Main(string[] args)
...
         }
 
         public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y)
         {
-            
+            Mailbox? box = mailboxes.GetMailbox(x, y);
+            if (box is null)
+                return "";
+            else
+                return box.ToString();
```
pnotequalnp: This is equivalent to `mailboxes.GetMailbox(x, y)?.ToString() ?? "";`

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,64 @@
...
+     * A Mailbox contains a Size, a value tuple (int X, int Y) for its Location, and a Person Owner
+     * The ToString method should include all property members.
+     * When including the Size in the ToString, it should output empty string when the Size 
+     * is the default (0) value, the name of the size with " Premium" it is a premium size,
+     * or simple the size name. For examples "Small" or "Medium Premium".
+     */
+
+    public class Mailbox
+    {
+        public Size MailboxSize;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,64 @@
...
+        public (int X, int Y) Location;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,64 @@
...
+        public Person Owner;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,64 @@
...
+
+        public override bool Equals(object obj)
+        {
+            if (obj is Mailbox) return this.Equals((Mailbox)obj);
+            return false;
+        }
+
+        public bool Equals(Mailbox other)
+        {
+            if (this.MailboxSize == other.MailboxSize &&
```
Keboo: `other` should be null checked before using it.

Mailbox/Mailbox.cs
Keboo: Fields should be private

Mailbox/Mailbox.cs
Keboo: Fields should be private

Mailbox/Mailbox.cs
Keboo: Fields should be private

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,94 @@
...
+     * clean up the Stream that was passed in.
+     *
+     * The DataLoader should handle the JsonReaderException. This exception will 
+     * be throw if the stream contents are not valid json. Such as "abc". 
+     * If this is thrown the method should gracefully return null.
+     *
+     */
+    public class DataLoader : IDisposable
     {
+        private Stream Source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,94 @@
...
+        
+        public void Dispose()
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,94 @@
...
-            
+            using var sw = new StreamWriter(Source);
+
+            foreach (var mailbox in mailboxes)
+            {
+                sw.WriteLine(JsonConvert.SerializeObject(mailbox));
+            }
         }
+
+        ~DataLoader() => Dispose();
```
Keboo: Modify 'DataLoader.Finalize' so that it calls Dispose(false) and then returns.

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
...
+    /*
+     * Create a new Person struct
+     * A person has a first and last name
+     * Person should implement IEquatable<Person>
+     * Ensure appropriate equality operators are also written
+     * Implement a reasonable ToString method
+     */
+    public struct Person : IEquatable<Person>
+    {
+        public string FirstName;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,50 @@
...
+        public string LastName;
```

Mailbox/Person.cs

Mailbox/Person.cs

Mailbox/Program.cs
```diff
@@ -5,9 +5,23 @@
...
+     * There is a method on Mailboxes that should be helpful here.
+     * It should not modify the passed in mailboxes.
+     */
     class Program
     {
-        private const int Width = 50;
+        private const int Width = 30;
         private const int Height = 10;
 
         static void Main(string[] args)
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,64 @@
...
+     * or simple the size name. For examples "Small" or "Medium Premium".
+     */
+
+    public class Mailbox
+    {
+        public Size MailboxSize;
+        public (int X, int Y) Location;
+        public Person Owner;
+
+        public Mailbox(Size size, ValueTuple<int, int> location, Person owner)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        public void DataLoader_LoadMailbox_Success()
+        {
+            string fileName = Path.GetRandomFileName();
+            Stream fileStream = File.Open(fileName, FileMode.Create);
```
Keboo: Stream implements IDisposable and should be cleaned up.
```suggestion
            using Stream fileStream = File.Open(fileName, FileMode.Create);
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+            
+            fileStream = File.Open(fileName, FileMode.Open);
+            var sut = new DataLoader(fileStream);
+
+            try
+            {
+                List<Mailbox>? loadedMailboxes = sut.Load();
+                var boxPairs = mailboxes.Zip(loadedMailboxes, (a, b) => (a, b));
+
+                Assert.IsNotNull(loadedMailboxes);
```
Keboo: If `loadedMailboxes` is null, the `.Zip` call above will throw an exception.

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+                foreach (var (box, loadedBox) in boxPairs)
+                {
+                    Assert.AreEqual(box, loadedBox);
+                }
+            }
+            finally
+            {
+                File.Delete(fileName);
```
Keboo: Using a `MemoryStream` will save you from needing to deal with files.

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoader_LoadWithNullSource_ThrowsException()
+        {
+            #nullable disable
+            var sut = new DataLoader(null);
+            #nullable enable
+            sut.Load();
```
Keboo: Since this is testing the constructor. You don't need the Load call.

Mailbox.Tests/MailboxesTests.cs
```diff
@@ -0,0 +1,32 @@
...
+            };
+
+            // Exception on bad values
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => new Mailboxes(boxes, 2, 0));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => new Mailboxes(boxes, 0, 2));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => new Mailboxes(boxes, -1, 3));
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => new Mailboxes(boxes, 3, -1));
+
+            // Exception on insufficient sized mailboxes
+            Assert.ThrowsException<ArgumentOutOfRangeException>(() => new Mailboxes(boxes, 2, 1));
```
Keboo: In general you want to separate out different conditions into different tests.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,94 @@
...
         {
+            if (source is null)
+                throw new ArgumentNullException(nameof(source));
+            Source = source;
         }
 
-        public List<Mailbox> Load()
+        public List<Mailbox>? Load()
         {
+            if (Source is null)
```
Keboo: Since you have null checked `Source` on its way in on the constructor you don't need to check it in here.

Mailbox/Program.cs
```diff
@@ -89,17 +98,55 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string displayStr = "";
+            foreach (Mailbox mailbox in mailboxes)
+            {
+                displayStr += mailbox.Owner.ToString() + Environment.NewLine;
+            }
+            return displayStr;
```

PR #197, opened 2019-10-31
Assignment5

twofingerrightclick:


Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,55 @@
...
+                Console.WriteLine(mailboxesStart[0]);
+                Console.WriteLine(mailboxesEnd[0]);
+
+                Assert.IsTrue(mailboxesEnd[0].Equals(mailboxesStart[0]));
+
+            }
+
+            finally
+            {
+                dl.Dispose();
```
Keboo: This is not necessary since you used the `using` for the DataLoader above.

Mailbox.Tests/MailboxesTests.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            new Mailbox(Sizes.LargePremium, (0, 0), new Person("Chris", "Martin")),
+            new Mailbox(Sizes.Small, (0, 1), new Person("Guy", "Berryman")),
+            new Mailbox(Sizes.Medium, (0, 2), new Person("Will", "Champion")),
+            new Mailbox(Sizes.Medium, (0, 3), new Person("Johnny", "Buckland"))
+        };
+
+        
+
+    }
```
Keboo: Would have been nice to see these tests written.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,93 @@
...
+            //need to read more on the following line
+            GC.SuppressFinalize(this);
         }
 
-        public void Save(List<Mailbox> mailboxes)
+        public void Dispose(bool dispose)
+        {
+            if (dispose)
+            {
+                Dispose();
```
Keboo: This is backwards, the `Dispose()` should be calling this method.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,93 @@
...
+        public void Dispose()
         {
+            if (!disposed)
+            {
+                StreamSource.Dispose();
+            }
+
+            disposed = true;
+            //need to read more on the following line
+            GC.SuppressFinalize(this);
```
Keboo: This should only ever be done inside of the finalizer, never inside of the Dispose method.

Mailbox/IConsole.cs
```diff
@@ -0,0 +1,39 @@
...
+
+        public void WriteLine(string output)
+        {
+            Console.WriteLine(output);
+        }
+    }
+
+    public class MockConsole : IConsole
+    {
+        public List<string> Input { get; set; } = new List<string>();
```
Keboo: Shouldn't this be called `Output` not `Input`?

Mailbox/IConsole.cs
```diff
@@ -0,0 +1,39 @@
...
+        public string ReadLine()
+        {
+            return Console.ReadLine();
```
Keboo: This is closed, but this is still going to the real Console. It would have been better to have something like a List<string> and on each call to `ReadLine` return the next item from the list.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,52 @@
...
+        public Mailbox(Sizes size, ValueTuple<int, int> location, Person Owner)
+        {
+            if (size.Equals(Sizes.Premium)) throw new ArgumentException(" \"Premium\" is not a size itself", (nameof(size)));
+
+            BoxSize = size;
+            Location = location;
+            this.Owner = Owner;
+        }
+
+        override
```
Keboo: Odd whitespace here.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,52 @@
...
+    public string ToString()
+        {
+            string mailboxString = $"Owner: {Owner.ToString()}, Location: x = {Location.X}, y = {Location.Y}";
```
Keboo: Typically inside of an interpolated string you do not need to explicitly call ToString(), it will convert to string for you.
```suggestion
            string mailboxString = $"Owner: {Owner}, Location: x = {Location.X}, y = {Location.Y}";
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,52 @@
...
+            if (BoxSize == Sizes.Unspecified)
+            {
+                return mailboxString;
+            }
+            return mailboxString + $", BoxSize: {BoxSize}";
```

Mailbox/Mailboxes.cs
```diff
@@ -59,5 +67,50 @@ public bool GetAdjacentPeople(int x, int y, out HashSet<Person> adjacentPeople)
 
             return isOccupied;
         }
+
+        public ValueTuple<int, int> GetOpenBox( Person person)
```
Keboo:
```suggestion
        public (int X, int Y) GetOpenBox(Person person)
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,56 @@
...
+            //object p2 = new Person()
+            //p1.Equals((Obj) p) versus p1.Equals(p2);
+        public override int GetHashCode()
+        {
+            return Name.GetHashCode();
+        }
 
+        public override bool Equals(object? obj)
+        {
+            return base.Equals(obj);
```
Keboo: Here you wanted to call your `Equals(Person)` overload above.

Mailbox/Program.cs
```diff
@@ -89,16 +104,65 @@ static void Main(string[] args)
...
+            if (mailboxes == null)
+            {
+                return "";
+            }
+            StringBuilder sb = new StringBuilder();
+            foreach (Mailbox mailbox in mailboxes) {
+                sb.Append(mailbox + Environment.NewLine);
+            }
+
+            return sb.ToString();
```
Keboo: he expectation on was that this method return a _distinct_ set of owner names. This just returns all mailbox names.

Mailbox/Program.cs
```diff
@@ -89,16 +104,65 @@ static void Main(string[] args)
...
+                                            Sizes size)
         {
+            if (mailboxes is null)
+            {
+                throw new ArgumentNullException(nameof(mailboxes));
+            }
+
+            if (string.IsNullOrEmpty(firstName))
+            {
+                throw new ArgumentException("message", nameof(firstName));
```
Keboo: A better exception message would have been good here.

Mailbox/Program.cs
```diff
@@ -89,16 +104,65 @@ static void Main(string[] args)
...
+            }
+            //validate text no nulls,.
+            Person person = new Person(firstName, lastName);
+
+            (int, int) newBoxCoordinates = mailboxes.GetOpenBox(person);
+
+            if (newBoxCoordinates == (-1, -1)) { 
+                // should do something to say that it can't add the mailbox.
+                Object o = new Object();
+                return (Mailbox)o;
```
Keboo: Not really sure what the intent was here. This cast will fail, perhaps simply returning `null`?

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,55 @@
...
+
+            }
+
+        }
+        
+    }
```
Keboo: Would have been nice to see more tests covering the other cases here.

PR #199, opened 2019-10-31
Assignment5_Joel

RSXEnthusiast:
I think I'm done. Please inform if you think otherwise.

### DIFF
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -1,20 +1,30 @@
-﻿using System.Collections.Generic;
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
 using System.IO;
 
-namespace Mailbox
-{
-    public class DataLoader
-    {
-        public DataLoader(Stream source)
-        {
+namespace Mailbox {
+    public class DataLoader {
+        private Stream _Stream;
+        public DataLoader(Stream source) {
+            _Stream = source ?? throw new ArgumentNullException("DataLoader Stream cannot be null");
         }
 
-        public List<Mailbox> Load()
-        {
+        public List<Mailbox> Load() {
+            List<Mailbox> result = new List<Mailbox>();
+            _Stream.Position = 0;
+            StreamReader sr = new StreamReader(_Stream);
+            while (!sr.EndOfStream) {
+                result.Add(JsonConvert.DeserializeObject<Mailbox>(sr.ReadLine()));
+            }
+            sr.Dispose();
+            return result;
         }
 
-        public void Save(List<Mailbox> mailboxes)
-        {
+        public void Save(List<Mailbox> mailboxes) {
+            _Stream.Position = 0;
+            StreamWriter sw = new StreamWriter(_Stream);
+            sw.WriteLine(JsonConvert.SerializeObject(mailboxes));
         }
     }
 }
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -3,22 +3,20 @@ using System;
 using System.IO;
 
 namespace Mailbox {
-    public class DataLoader {
-        private String _FileName;
-        public DataLoader(String fileName) {
-            _FileName = fileName;
+    public class DataLoader : IDisposable {
+        bool disposed = false;
+
+        private readonly Stream _Source;
+        public DataLoader(Stream source) {
+            _Source = source ?? throw new ArgumentNullException(nameof(source));
         }
 
         public Mailbox[,] Load() {
+            _Source.Position = 0;
             Mailbox[,] result;
-            if (!File.Exists(_FileName)) {
-                return null;
-            }
-            using (StreamReader sr = new StreamReader(_FileName)) {
+            using (StreamReader sr = new StreamReader(_Source, leaveOpen: true)) {
                 try {
-                    string temp = sr.ReadLine();
-                    Console.WriteLine(temp);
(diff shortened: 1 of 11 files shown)


Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
-﻿namespace Mailbox
-{
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
+namespace Mailbox {
+    public struct Person : IEquatable<Person> {
+        public string _FirstName { get; }
```
mmwoodfo:
```suggestion
        public string FirstName { get; }
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
...
+        public string _LastName { get; }
+
+        public Person(string _FirstName, string _LastName) {
+            this._FirstName = _FirstName;
```
mmwoodfo:
```suggestion
            FirstName = firstName;
```
Parameters should be named with camelCase

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,35 @@
...
 
-namespace Mailbox
-{
-    public class DataLoader
-    {
-        public DataLoader(Stream source)
-        {
+namespace Mailbox {
+    public class DataLoader {
+        private String _FileName;
```
mmwoodfo:
```suggestion
        private readonly Stream _Source;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
...
+            this._LastName = _LastName;
+        }
+
+        public bool Equals([AllowNull] Person other) {
+            return this.toString().Equals(other.toString());
+        }
+
+        public string toString() {
+            return $"{_FirstName} {" "} {_LastName}";
```
NikolasGraham: You don't need the extra brackets in the middle at all. What $" " does is it creates a string that allows variables to be dynamically added into the string.  ```return $"{_FirstName} {" "} {_LastName};"``` will actually return a string with 3 spaces between first and last name, but what I assume you're trying to do looks more like ```return $"{_FirstName} {_LastName};"```

RSXEnthusiast: Gotcha, I think I'd noticed this, but forgot to change it. Thanks for the explanation!

Mailbox/Person.cs
RSXEnthusiast: You're right. I don't remember what I was doing, but I swear I had a reason!

Thanks.

Mailbox/Person.cs
RSXEnthusiast: You're right, I'm finally getting the hang of naming conventions in c#.

Mailbox/DataLoader.cs
RSXEnthusiast: I had changed it to function differently, but i re-read the assignment instructions and realize that it does specify that we use a Stream. I'll have to fix that, thanks.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,50 @@
...
 
-namespace Mailbox
-{
-    public class DataLoader
-    {
-        public DataLoader(Stream source)
-        {
+namespace Mailbox {
+    public class DataLoader : IDisposable {
+        bool disposed = false;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
-﻿namespace Mailbox
-{
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
+namespace Mailbox {
+    public struct Person : IEquatable<Person> {
+        public string _FirstName;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
...
+        public string _LastName;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,22 @@
-﻿namespace Mailbox
-{
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
+namespace Mailbox {
+    public struct Person : IEquatable<Person> {
```
Keboo: Person should override Equals(object).

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,16 @@
-﻿namespace Mailbox
-{
+﻿namespace Mailbox {
+    public class Mailbox {
+        public Sizes Size { get; }
+        public (int, int) Location { get; }
+        public Person Owner { get; set; }
+        public Mailbox(Sizes Size, (int, int) Location, Person Owner) {
```
Keboo: Parameter names should be camelCase. This will prevent needing to use `this` below.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,16 @@
...
+            this.Size = Size;
+            this.Location = Location;
+            this.Owner = Owner;
+        }
 
+        public string toString() {
+            return $"{"Size: " + Size + " Location: (" + Location + ") Owner: " + Owner.toString()}";
```

Mailbox/Program.cs
```diff
@@ -72,34 +60,48 @@ static void Main(string[] args)
...
+            return result;
         }
 
-        public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y)
-        {
-            
+        public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y) {
+            try {
+                return mailboxes.MailboxesArray[x, y].toString();
+            } catch (NullReferenceException) {
```
Keboo: You never want to catch `NullReferenceException`. This exception type always means programmer error, so you should use one of the other C# operators to handle the null case rather than suppress it.

PR #200, opened 2019-11-01
Assignment 5

JerettLatimer:


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,73 @@
...
+            {
+                foreach (Mailbox mailbox in mailboxes)
+                {
+                    writer.WriteLine(JsonConvert.SerializeObject(mailbox));
+                }
+            }
         }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```

Mailbox/Program.cs
```diff
@@ -69,7 +69,9 @@ static void Main(string[] args)
                     case 4:
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,37 @@
...
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; }
+        public (int x, int y) Location { get; set; }
+        public Person Owner { get; set; }
+
+        public Mailbox(Size size, ValueTuple<int,int> location, Person owner)
```
```suggestion
        public Mailbox(Size size, (int,int) location, Person owner)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            list.Add(new Mailbox(Size.Large, (1, 6), new Person("first", "Last")));
+            list.Add(new Mailbox(Size.Large, (3, 1), new Person("first", "Last")));
+
+            // Act
+            sut.Save(list);
+            List<Mailbox> testList = sut.Load();
+
+            // Assert
+            Assert.AreEqual(testList.Count, list.Count);
+            sut.Dispose();
```
Keboo: You want to take advantage of the `using` for disposing objects. If the above code throws an exception (such as the case where an Assert fails) this `Dispose` will never get called.
```C#
using DataLoader sut = new DataLoader(source);
```

Mailbox.Tests/MailboxTests.cs
```diff
@@ -0,0 +1,44 @@
...
+        public void Mailbox_ToString_ReturnsCorrectlyFormattedString(Size size)
+        {
+            // Arrange
+            Person testPerson = new Person("First", "Last");
+            Mailbox sut = new Mailbox(size, (1, 1), testPerson);
+
+            // Act
+
+            // Assert
+            Assert.AreEqual(sut.ToString(), $"Size: {size} Location: (1, 1) Owner: First Last");
```
Keboo: This "works" but your two parameters are backwards. The first parameter is the "expected" value and the second is the "actual" value. With them backwards the error message that is displayed on failure will be wrong.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,73 @@
...
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private Stream Source { get; set; }
         public DataLoader(Stream source)
         {
+            Source = source;
```
Keboo: `source` should be null checked
```suggestion
            Source = source ?? throw new ArgumentNullException(nameof(source));
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,37 @@
...
+        public override string ToString()
+        {
+            String size;
+            if(Size == Size.Default)
+            {
+                size = "";
+            }
+            else if(Size == Size.Premium)
+            {
+                size = Size.ToString() + "Premium";
```
Keboo:
```suggestion
                size = Size.ToString() + " Premium";
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,66 @@
...
+namespace Mailbox
 {
+#pragma warning disable CA1067 // Override Object.Equals(object) when implementing IEquatable<T>
+    public struct Person : IEquatable<Person>
+#pragma warning restore CA1067 // Override Object.Equals(object) when implementing IEquatable<T>
+    {
+        private string? FirstName { get; set; }
+        private string? LastName { get; set; }
+
+        public Person(String? firstName, String? lastName)
```
Keboo:
```suggestion
        public Person(string? firstName, string? lastName)
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,66 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+#pragma warning disable CA1067 // Override Object.Equals(object) when implementing IEquatable<T>
+    public struct Person : IEquatable<Person>
+#pragma warning restore CA1067 // Override Object.Equals(object) when implementing IEquatable<T>
+    {
+        private string? FirstName { get; set; }
+        private string? LastName { get; set; }
```
Keboo: Given the null checks in the constructor these could be made readonly non-nullable properties.
```suggestion
        private string LastName { get; }
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,66 @@
...
+            }
+            else if(lastName is null)
+            {
+                throw new ArgumentNullException(nameof(lastName));
+            }
+            FirstName = firstName;
+            LastName = lastName;
+        }
+
+        public bool Equals(Person other)
```
Keboo: `other` should be null checked

Mailbox/Person.cs
```diff
@@ -1,4 +1,66 @@
...
+        public override string ToString()
+        {
+            return $"{FirstName} {LastName}";
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is null)
+                return false;
+            return Equals((Person)obj);
```
Keboo: This will cause an invalid cast exception if this is compared with something that is not a Person.
```suggestion
            return obj is Person person && Equals(person);
```

Mailbox/Program.cs
```diff
@@ -89,17 +91,41 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string owners = "";
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                owners += mailbox.Owner.ToString() + " ";
+            }
+            return owners;
```

PR #201, opened 2019-11-02
Nearly finished with tests

jacob-berger:
Saves/Loads, and Mailbox creation all work fine.

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,43 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Mailbox;
+using System;
+using System.Collections.Generic;
+using System.Text;
+using System.IO;
+
+namespace Mailbox.Tests
+{
+    [TestClass()]
+    public class DataLoaderTests
+    {
+        [TestMethod()]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoaderNullStreamThrowsException()
+        {
+            //Arrange
+
+            //Act
+            DataLoader dataLoader = new DataLoader(null);
+
+            //Assert
+        }
+
+        [TestMethod()]
+        public void SaveGoodData()
+        {
+            //Arrange
+            List<Mailbox> toSave = new List<Mailbox>();
+            List<Mailbox> saved = new List<Mailbox>();
+            MemoryStream source = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(source);
+
+            toSave.Add(new Mailbox(new Person("FirstName", "LastName"), Sizes.Small, (0, 0)));
+
+            //Act
+            dataLoader.Save(toSave);
+
+            //Assert
+
+        }
+    }
+}
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -23,21 +23,25 @@ namespace Mailbox.Tests
         }
 
         [TestMethod()]
-        public void SaveGoodData()
+        public void SaveAndLoadGoodData()
         {
             //Arrange
             List<Mailbox> toSave = new List<Mailbox>();
             List<Mailbox> saved = new List<Mailbox>();
             MemoryStream source = new MemoryStream();
             DataLoader dataLoader = new DataLoader(source);
+            List<Mailbox> loaded = new List<Mailbox>();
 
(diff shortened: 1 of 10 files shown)


TatyanaHubbard:
You are using an extra list in your DataLoader SaveandLoad test, but it's good!

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,83 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private Stream Source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,83 @@
...
+        private bool disposedValue = false;
```

Mailbox/Program.cs
```diff
@@ -43,9 +43,9 @@ static void Main(string[] args)
                         Console.WriteLine("Enter the last name");
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,47 @@
...
+            //Arrange
+
+            //Act
+            DataLoader dataLoader = new DataLoader(null);
+
+            //Assert
+        }
+
+        [TestMethod()]
+        public void SaveAndLoadGoodData()
```
Keboo: In general if you find yourself using word like "And" in a unit test name, it usually means these two cases should be split apart into their own unit tests.

Mailbox.Tests/MailboxTests.cs
```diff
@@ -0,0 +1,61 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass()]
+    public class MailboxTests
+    {
+        [TestMethod()]
+        public void CreateOwnerGoodValues()
```
Keboo: This appears to be a test on `Person` not on `Mailbox`

Mailbox.Tests/MailboxTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+            //Act
+            Person person = new Person(first, last);
+
+            //Assert
+            Assert.AreEqual(person.toString(), "Test Box");
+        }
+
+        [TestMethod]
+        public void CreateMailboxGoodValuesAllSizes()
```
Keboo: Since the `ToString` method is what is being tested here you want to make the unit test name reflect that.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,30 @@
...
+        public (int x, int y) Location { get; set; }
+
+        public Mailbox(Person owner, Sizes boxSize, (int x, int y) location)
+        {
+            Owner = owner;
+            BoxSize = boxSize;
+            Location = location;
+        }
+
+        public string toString()
```
Keboo: Though the implementation is fine, this is not quite right. There is a `ToString` method that exists on the base object type that you [should override](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/how-to-override-the-tostring-method) not create your own.

Mailbox/Person.cs
```diff
@@ -1,4 +1,61 @@
...
+        {
+            return Equals(a, b);
+        }
+
+        public static bool operator !=(Person a, Person b)
+        {
+            return !(a == b);
+        }
+
+        public string toString()
```
Keboo: This should be overridden from the base type.

Mailbox/Program.cs
```diff
@@ -88,18 +88,81 @@ static void Main(string[] args)
...
+                throw new ArgumentNullException(nameof(mailboxes));
+            }
+
+            string result = "";
+            foreach (Mailbox mailbox in mailboxes)
+            {
+                result += $"{mailbox.Owner.toString()}\n";
+            }
+
+            return result;
```

Mailbox/Program.cs
```diff
@@ -88,18 +88,81 @@ static void Main(string[] args)
...
+
+            Person owner = new Person(firstName, lastName);
+            for (int ix = 1; ix <= 10; ix++)
+            {
+                for (int iy = 1; iy < 30; iy++)
+                {
+                    if (!locationList.Contains((ix, iy)))
+                    {
+                        HashSet<Person> personHash = new HashSet<Person>();
+                        if (!mailboxes.GetAdjacentPeople(ix, iy, out personHash))
```
Keboo: The assignment called for asserting that a mailbox may not be placed adjacent to another mailbox that contains a person with the same name. I see that this calls `GetAdjacentPeople` but it does not actually do any checks with it.

PR #202, opened 2019-11-03
Assignment5

Cybotive:
JsonConvert doesn't preserve owner details, so mailboxes can't be loaded. Hence no unit tests on DataLoader.

### DIFF
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -11,6 +11,7 @@ namespace Mailbox
 
         public List<Mailbox> Load()
         {
+            return null;//TODO
         }
 
         public void Save(List<Mailbox> mailboxes)
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -17,12 +17,21 @@ namespace Mailbox
             this.source = source;
         }
 
-        public List<Mailbox> Load()
+        public List<Mailbox>? Load()
         {
             source.Position = 0;
             streamReader = new StreamReader(source);
+            List<Mailbox> mailboxList;
 
-            List<Mailbox> mailboxList = JsonConvert.DeserializeObject<List<Mailbox>>(streamReader.ReadToEnd());
+            try
+            {
+                mailboxList = JsonConvert.DeserializeObject<List<Mailbox>>(streamReader.ReadToEnd());
+            }
+            catch (JsonReaderException)
+            {
+                return null;
+            }
+            
 
             if(mailboxList == null)
             {
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -1,22 +1,82 @@
-﻿using System.Collections.Generic;
+﻿using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
 using System.IO;
+using System.Text;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream source;
+        private StreamWriter? streamWriter;
+        private StreamReader? streamReader;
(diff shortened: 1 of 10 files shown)


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+using System.Collections.Generic;
 using System.IO;
+using System.Text;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+        private StreamWriter? streamWriter;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+        private StreamReader? streamReader;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
-            
+            source.Position = 0;
+            streamWriter = new StreamWriter(source);
+
+            streamWriter.Write(JsonConvert.SerializeObject(mailboxes));
+            streamWriter.Flush();
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,56 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        private (string firstName, string lastName) name;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
         public DataLoader(Stream source)
         {
+            this.source = source;
         }
 
-        public List<Mailbox> Load()
+        public List<Mailbox>? Load()
         {
+            source.Position = 0;
+            streamReader = new StreamReader(source);
```
Keboo: Since this is only used here, it could easily be declared as a local variable. 
```suggestion
            using var streamReader = new StreamReader(source, leaveOpen:true);
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+
+            List<Mailbox> mailboxes = new Mailboxes(mailboxList, 30, 10);
+            return mailboxes;
         }
 
         public void Save(List<Mailbox> mailboxes)
         {
-            
+            source.Position = 0;
+            streamWriter = new StreamWriter(source);
```
Keboo: This could be declared as a local variable:
```suggestion
            using var streamWriter = new StreamWriter(source, leaveOpen:true);
```

Mailbox/Program.cs
```diff
@@ -87,19 +87,72 @@ static void Main(string[] args)
...
+            int width = mailboxes.Width;
+            int height = mailboxes.Height;
+
+            for (int i = 0; i < width; i++)
+            {
+                for (int a = 0; a < height; i++)
+                {
+                    if (!mailboxes.GetAdjacentPeople(i, a, out HashSet<Person> personSet))
+                    {
+                        Boolean valid = true;
```
Keboo: Use keyword over type name.
```suggestion
                        bool valid = true;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,56 @@
...
+
+            return Equals(other as Person?);
+        }
+
+        public override int GetHashCode()
+        {
+            return name.GetHashCode();
+        }
+
+        private (string first, string last) getName()
```
Keboo: Not sure what this method accomplishes. Since it is private, you can simply replace it with the code directly accessing the `name` field.

PR #204, opened 2019-11-03
Assignment 5

Fervidusletum:


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,64 @@
...
-            
-        }
+            Source.Position = 0;
+            using StreamWriter writer = new StreamWriter(Source, leaveOpen: true);
+
+            writer.Write(JsonConvert.SerializeObject(mailboxes));
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```

Mailbox/Program.cs
```diff
@@ -3,19 +3,19 @@
...
-namespace Mailbox
+namespace MailRoom
 {
-    class Program
+    public static class Program
     {
         private const int Width = 50;
         private const int Height = 10;
 
-        static void Main(string[] args)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,37 @@
...
+using System.Text;
+using System.IO;
+
+namespace MailRoom.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        public void SaveAndLoad_SaveMailboxesAndLoadMailboxes_ReturnsMailboxesMatchingOriginal()
```
Keboo: This works, but in general if you find yourself using using words like conjunctions in your unit test name, it probably means you should have a second unit test.

Mailbox.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,118 @@
...
+        [DataTestMethod]
+        public void GetMailboxDetails_GetsStringRepresentationOfMailbox_ReturnsConcatenatedMailbox()
+        {
+            Mailbox box = new Mailbox(new Person("F1", "L1"), (0, 0), Sizes.Small);
+            MailboxCollection mailboxes = new MailboxCollection(new List<Mailbox> { box }, 1, 1);
+            string expected = box.ToString();
+
+            string result = Program.GetMailboxDetails(mailboxes, 0, 0) ?? "";
+
+            Assert.AreEqual<string>(expected, result);
```
Keboo: In general you don't need to specify the generic parameter type unless the values passed in are of different types. The compiler can infer the correct generic type.
```suggestion
            Assert.AreEqual(expected, result);
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,37 @@
...
+        {
+            FirstName = firstName ?? throw new ArgumentNullException(nameof(firstName));
+            LastName = lastName ?? throw new ArgumentNullException(nameof(lastName));
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is null) return false;
+
+            return Equals((Person) obj);
```
Keboo: This will actually throw an exception if a different object instance that is not a `Person` gets passes in.
A better solution would be to use pattern matching.
```suggestion
            return obj is Person person && Equals(person);
```

PR #205, opened 2019-11-03
Assignment5 work branch

samrnixon:
Couldn't figure out how to properly test Load() and Save(), but that's about it.

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,80 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type.
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoaderStreamSourceIsNull_ThrowsException()
+        {
+            using var dataLoader = new DataLoader(null);
+        }
+
+/*        [TestMethod]
+        public void DataLoaderStreamSourceIsValid_LoadWorks()
+        {
+            var personA = new Person("Sam", "Nixon");
+
+            List<Mailbox>? mailboxes = new List<Mailbox>()
+            {
+                new Mailbox(personA, (0,0), Sizes.Large)
+            };
+
+            string jsonData = JsonConvert.SerializeObject(mailboxes);
+
+            using var ms = new MemoryStream();
+            using var writer = new StreamWriter(ms, leaveOpen: true);
+            writer.WriteLine(jsonData);
+            DataLoader dataLoader = new DataLoader(ms);
+
+            List<Mailbox>? testList = dataLoader.Load();
+
+            Assert.AreEqual(mailboxes.Count, testList.Count);
+        }
+*/
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoaderSaveNullPassed_ThrowsArgumentNullException()
+        {
+            using var ms = new MemoryStream();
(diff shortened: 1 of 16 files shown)


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,93 @@
...
+            foreach(Mailbox mb in mailboxes)
+            {
+                string data = JsonConvert.SerializeObject(mb);
+                writer.WriteLine(data);
+            }
+            writer.Dispose();
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false;
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,80 @@
...
+        public void DataLoaderSaveNullPassed_ThrowsArgumentNullException()
+        {
+            using var ms = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(ms);
+
+            dataLoader.Save(null);
+        }
+
+/*        [TestMethod]
+        public void DataLoaderStreamSourceIsValid_SaveWorks()
```
Keboo: Is there a reason these tests are not included?

Mailbox.Tests/MailboxTests.cs
```diff
@@ -0,0 +1,48 @@
...
+{
+    [TestClass]
+    public class MailboxTests
+    {
+        [TestMethod]
+        public void Mailbox_JustTestingToString()
+        {
+            Person person = new Person("Sam", "Nixon");
+            Mailbox mailbox = new Mailbox(person, (1,2), Sizes.None);
+            mailbox.ToString();
```
Keboo: This `.ToString()` can be removed as it doesn't do anything.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,93 @@
...
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    DisposableMember?.Dispose();
```
Keboo: This was just an example from lecture. The expectation was that you dispose the stream `Source`

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,93 @@
...
+                    }
+                    mailboxes.Add(JsonConvert.DeserializeObject<Mailbox>(jsonData));
+                }
+            }
+            catch(JsonReaderException)
+            {
+                return null;
+            }
+
+            reader.Dispose();
```
Keboo: This explicit `.Dispose` call is not needed since you properly put the `using` statement above.

Mailbox/Program.cs
```diff
@@ -89,17 +88,97 @@ static void Main(string[] args)
...
+            }
+
+            string owners = "";
+
+            foreach (Mailbox mb in mailboxes)
+            {
+                owners += (mb.ToString() + "\n");
+            }
+
+            return owners;
```
Keboo: The expectation on was that this method return a _distinct_ set of owner names. This just returns all of the mailbox strings.

Mailbox/Program.cs
```diff
@@ -89,17 +88,97 @@ static void Main(string[] args)
...
+            if (x < 0 || x > Width)
+            {
+                throw new ArgumentOutOfRangeException(nameof(x));
+            }
+            if(y < 0 || y > Height)
+            {
+                throw new ArgumentOutOfRangeException(nameof(y));
+            }
+
+            string details = "";
```
Keboo: This does not appear to add any value since you return as soon as you find the mailbox.

Mailbox/Program.cs
```diff
@@ -89,17 +88,97 @@ static void Main(string[] args)
...
+                    {
+                        if(!adjacentPeople.Contains(owner))
+                        {
+                            Mailbox mailbox = new Mailbox(owner, (x, y), size);
+                            return mailbox;
+                        }
+                    }
+                }
+            }
+            #pragma warning disable CS8603 // Possible null reference return.
```
Keboo: In this case rather than disabling the warning it would have been better to make the return type from this method nullable.

PR #206, opened 2019-11-03
Assignment5

skozubenko1:
All necessarry code is implemented. Just need to add tests


Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
...
+            return this._FirstName == other._FirstName && this._LastName == other._LastName;
+		}
+         
+        public override string ToString()
+        {
+            return $"{_FirstName}, {_LastName}";
+        }
+        
+    }
 }
```
GamerBah: Don't forget to write methods for the other equality operators `==` and `!=`.

Also missing a line feed (blank line) at the end, here.

PR #207, opened 2019-11-03
Assignment 5 - WIP

whereisdaniru:
Working on assignment 5

### DIFF
--- a/Mailbox/Person.cs
+++ b/Mailbox/Person.cs
@@ -14,16 +14,38 @@ namespace Mailbox
             LastName = lastName;
         }
 
-        public bool Equals(Person other)
+        public bool Equals([AllowNull]Person other)
         {
-            if(LastName == other.LastName)
+            if(LastName == other.LastName && FirstName == other.FirstName)
             {
-                if(FirstName == other.FirstName)
-                {
-                    return true;
-                }
+                return true;
             }
             return false;
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is null)
+            {
+                return false;
+            }
+            return Equals((Person)obj);
+        }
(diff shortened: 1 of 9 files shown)


Mailbox/Person.cs
```diff
@@ -1,4 +1,62 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person> 
+    {
+        private string FirstName { get; set; }
+        private string LastName { get; set; }
```
Keboo: Properties can be made readonly
```suggestion
        private string LastName { get; }
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,62 @@
...
+            return false;
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is null)
+            {
+                return false;
+            }
+            return Equals((Person)obj);
```
Keboo: This will throw a cast exception if a non-null object is passed into the equals method.
Such as:
```
Person p = new Person("Foo", "bar");
p.Equals(new object());
```

A better solution would be to use pattern matching.
```suggestion
            return obj is Person person && Equals(person);
```

Mailbox/Program.cs
```diff
@@ -89,17 +91,41 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string owners = "";
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                owners += mailbox.Owner.ToString() + ", ";
+            }
+            return owners;
```

PR #208, opened 2019-11-03
Assignment 5

AustinHall7798:


Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+                foreach (MailBox mailbox in testBoxes)
+                    sw.WriteLine(mailbox);
+            }
+            memory.Position = 0;
+
+            DataLoader dataLoader = new DataLoader(memory);
+
+            Assert.IsNull(dataLoader.Load());
+
+            dataLoader.Dispose();
```
Keboo: If the Assert method fails, you will never get to this dispose. It would be better to simply mark the `dataLoader` with a using.
```C#
using DataLoader dataLoader = new DataLoader(memory);
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,92 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Constructor_NullArgument_ThrowsException()
+        {
+            DataLoader dataLoader = new DataLoader(null);
+        }
+
+        [TestMethod]
+        public void Load_UnSerializedData_ReturnsNull()
+        {
+            MemoryStream memory = new MemoryStream();
```

Mailbox.Tests/PersonTests.cs
```diff
@@ -0,0 +1,54 @@
...
+    public class PersonTests
+    {
+        [TestMethod]
+        [DataRow("firstname", "lastname")]
+        [DataRow("first name", "Last-name")]
+        [DataRow("NotNull", "ALso not_null")]
+        public void Constructor_CorrectParameters_ReturnsPerson(string firstName, string lastName)
+        {
+            Person person = new Person(firstName, lastName);
+            Assert.IsNotNull(person);
```
Keboo: This Assert does not add anything since it effectively just tests that the C# constructor works.

Mailbox.Tests/SizesTest.cs
```diff
@@ -0,0 +1,29 @@
+﻿using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class SizesTest
+    {
+        [TestMethod]
+        public void IsValid_InvalidSizes_ReturnsFalse()
```
Keboo: This is fine but it would have been cleaner as a data driven test like what you did with the `PersonTests`

Mailbox/Person.cs
```diff
@@ -1,4 +1,47 @@
...
+        {
+            FirstName = firstName;
+            LastName = lastName;
+        }
+
+        public override int GetHashCode() => (FirstName, LastName).GetHashCode();
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is Person)
```
Keboo: A better approach would be pattern matching.

```suggestion
            if (obj is Person person ) return Equals(person);
```

Mailbox/Program.cs
```diff
@@ -89,17 +90,53 @@ static void Main(string[] args)
...
         {
-            
+            if (mailboxes is null) throw new ArgumentNullException(nameof(mailboxes));
+
+            string owners = "";
+            foreach (var item in mailboxes)
+            {
+                owners += item.Owner;
+            }
+            return owners;
```

PR #209, opened 2019-11-04
Assignment5

jake-mcg:
WIP

### DIFF
--- a/Mailbox.Tests/PersonTests.cs
+++ b/Mailbox.Tests/PersonTests.cs
@@ -1,9 +1,88 @@
 ﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Mailbox.Tests
 {
     [TestClass]
     public class PersonTests
     {
+        [TestMethod]
+        [DataRow(null, null)]
+        [DataRow("Test", null)]
+        [DataRow(null, "Name")]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void NullName(string firstName, string lastName)
+        {
+            Person person = new Person(firstName, lastName);
+        }
+
+
+        [TestMethod]
+        [DataRow("Test", "test")]
+        [DataRow("First", "Last")]
+        public void Equals_AreEqual(string firstName, string lastName)
+        {
+            Person first = new Person(firstName, lastName);
+            Person second = new Person(firstName, lastName);
+            bool success = first.Equals(second);
+            Assert.IsTrue(success);
+        }
+
+        [TestMethod]
+        [DataRow("Test", "test")]
+        [DataRow("First", "Last")]
+        public void Equals_AreNotEqual(string firstName, string lastName)
+        {
+            Person first = new Person(firstName, lastName);
+            Person second = new Person(lastName, firstName);
+
+            bool success = first.Equals(second);
+
+            Assert.IsFalse(success);
+        }
+
+        [TestMethod]
+        [DataRow("Test", "test")]
+        [DataRow("First", "Last")]
+        public void EqualsOperator_AreEqual(string firstName, string lastName)
+        {
+            Person first = new Person(firstName, lastName);
+            Person second = new Person(firstName, lastName);
+
+            Assert.IsTrue(first == second);
+        }
+
+        [TestMethod]
+        [DataRow("Test", "test")]
+        [DataRow("First", "Last")]
+        public void EqualsOperator_AreNotEqual(string firstName, string lastName)
+        {
+            Person first = new Person(firstName, lastName);
+            Person second = new Person(lastName, firstName);
+
+            Assert.IsFalse(first == second);
+        }
+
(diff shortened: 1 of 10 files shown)


Mailbox/Size.cs
```diff
@@ -1,4 +1,15 @@
...
+namespace Mailbox
 {
+    [Flags]
+    public enum Size 
+    {
+        Default,
+        Small,
+        Medium2,
+        Large,
+        Premium
```
CDobbins1: When assigning values to flag enums, you should use powers of 2 to ensure that all flag combinations are represented uniquely.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,53 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private Stream Source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,53 @@
...
+
         public DataLoader(Stream source)
         {
-        }
-
+            Source = source ?? throw new ArgumentNullException($"Stream source cannot be null: {nameof(source)}");
+        }
+
+        public void Dispose()
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,26 @@
-﻿namespace Mailbox
+﻿using System;
+
+namespace Mailbox
 {
+    public class Mailbox 
+    {
+        public ValueTuple<int, int> Location { get; }
```
```suggestion
        public (int X, int Y) Location { get; }
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,26 @@
...
+        public Person Owner { get; }
+        public Size Size { get; }
+
+
+
+        public Mailbox(ValueTuple<int,int> location, Person owner, Size size)
```
```suggestion
        public Mailbox((int, int) location, Person owner, Size size) 
```

Mailbox.Tests/PersonTests.cs
```diff
@@ -0,0 +1,88 @@
...
+
+        [TestMethod]
+        [DataRow("Test", "test")]
+        [DataRow("First", "Last")]
+        public void DoesNotEqualsOperator_AreEqual(string firstName, string lastName)
+        {
+            Person first = new Person(firstName, lastName);
+            Person second = new Person(firstName, lastName);
+
+            Assert.IsFalse(first != second);
```
Keboo: This does not match the name of the unit test

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,26 @@
...
+        public Mailbox(ValueTuple<int,int> location, Person owner, Size size) 
+        {
+            Location = location;
+            Owner = owner;
+            Size = size;
+        }
+
+        public override string ToString()
+        {
+            return Size == 0 ? string.Empty : $"{Owner.ToString()}'s {Size} box is located at row {Location.Item1} and column {Location.Item2}.";
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,43 @@
...
+
+        public Person(string firstName, string lastName) 
+        {
+            FirstName = firstName ?? throw new ArgumentNullException($"First name cannot be null: {nameof(firstName)}");
+            LastName = lastName ?? throw new ArgumentNullException($"Last name cannot be null: {nameof(lastName)}");
+        }
+        public bool Equals([AllowNull] Person other)
+        {
+            if (ReferenceEquals(other, null)) return false;
+            if (ReferenceEquals(this, other)) return true;
```
Keboo: This will never match. `ReferenceEquals` takes in `object` parameters so this will box the values giving each of them different references.

Mailbox/Person.cs
```diff
@@ -1,4 +1,43 @@
...
+
+            return FirstName == other.FirstName
+                && LastName == other.LastName;
+        }
+
+        public override bool Equals(Object obj) 
+        {
+            return obj is Person ? this.Equals((Person)obj) : false;
```
Keboo: A cleaner approach is to use pattern matching.
```suggestion
            return obj is Person person && this.Equals(person);
```

Mailbox/Program.cs
```diff
@@ -89,17 +89,43 @@ static void Main(string[] args)
 
         public static string GetOwnersDisplay(Mailboxes mailboxes)
         {
-            
+            string ownersDisplay = "";
+
+            foreach (Mailbox mailbox in mailboxes) 
+            {
+                ownersDisplay += $"{mailbox.Owner.ToString()}\n";
```
Keboo: Inside of an interpolated string, it will convert the parameters to strings for you.
```suggestion
                ownersDisplay += $"{mailbox.Owner}\n";
```

Mailbox/Program.cs
```diff
@@ -89,17 +89,43 @@ static void Main(string[] args)
...
+            }
+
+            return ownersDisplay;
```

PR #210, opened 2019-11-04
Assignment 5

CDobbins1:


Mailbox/Mailboxes.cs
```diff
@@ -1,6 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
-
+#pragma warning disable CA1710
```
Fervidusletum: Dont forget to add a comment justifying why you're disabling a warning.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,23 @@
...
+    public class Mailbox
+    {
+        public Sizes Size { get; }
+        public (int, int) Location { get; }
+        public Person Owner { get; }
 
+        public Mailbox(Sizes size, (int, int) location, Person owner)
+        {
+            Size = size;
+            Location = location;
```
Fervidusletum: Should have a check to make sure the location is valid for `Mailboxes` class.  Would be a good idea to verify size also.  For example, you dont want a `Sizes.Large|Sizes.Medium`.  I could see reasons for allowing owner to be null.

Mailbox/Program.cs
```diff
@@ -89,17 +90,57 @@ static void Main(string[] args)
...
+            var sb = new StringBuilder();
+            foreach (Mailbox mb in mailboxes)
+            {
+                if (!hs.Contains(mb.Owner))
+                {
+                    sb.AppendLine(mb.Owner.ToString());
+                    hs.Add(mb.Owner);
+                }
+            }
+            return sb.ToString();
```
Fervidusletum:
```c#
            var hs = new HashSet<Person>();
            foreach (Mailbox mb in mailboxes)
            {
                 hs.Add(mb.Owner);
            }
            return string.Join(Environment.NewLine, hs);
```
Bit of a cleaner option for concatenating and returning all unique owners.

Mailbox.Tests/PersonTests.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class PersonTests
```
Fervidusletum: Couldn't hurt to do a few more edge case checks, like `person.Equals(null)`, and maybe a simple check just to ensure the operators `==` and `!=` are working as expected.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,23 @@
...
+            Owner = owner;
+        }
+
+        public override string ToString()
+        {
+            string outputSize = Size != 0 ? Size.ToString() : "";
```

PR #212, opened 2019-11-04
Assignment5

dbosco051394:
~~WIP. Will update this comment when it's complete.~~

Edit: This assignment is complete.
Shortcomings:
- When adding a new mailbox through program.cs, it always defaults to 'Small' even when no input occurred for that item rather than 'Default.'

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -2,6 +2,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace Mailbox.Tests
@@ -17,13 +18,14 @@ namespace Mailbox.Tests
             List<Mailbox> mailboxes = new List<Mailbox>()
             {
                 new Mailbox(Size.Small, 0, 0, new Person("Drew", "Bosco")),
-                new Mailbox(Size.Premium | Size.Small, 0, 0, new Person("Audrey", "Halfhill"))
+                new Mailbox(Size.Premium | Size.Small, 0, 1, new Person("Audrey", "Halfhill"))
             };
 
             dl.Save(mailboxes);
             List<Mailbox> retrieveBoxes = dl.Load();
 
             Assert.IsNotNull(retrieveBoxes);
+            Assert.IsTrue(mailboxes.Count == retrieveBoxes.Count);
         }
     }
 }
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,29 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        public void SaveToStream_LoadFromStream_CompareLists()
+        {
+            DataLoader dl = new DataLoader(File.Open("Mailboxes.json", FileMode.OpenOrCreate, FileAccess.ReadWrite));
+
+            List<Mailbox> mailboxes = new List<Mailbox>()
+            {
+                new Mailbox(Size.Small, 0, 0, new Person("Drew", "Bosco")),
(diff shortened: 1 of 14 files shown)


kenny2892:
You need to check for null parameters in the constructors of Person and DataLoader.

Mailbox/Person.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
+{
+    public struct Person : IEquatable<Person>
+    {
+        public string fname;
```
Fields should be private

Mailbox/Person.cs
```diff
@@ -0,0 +1,31 @@
...
+        public string lname;
```
Fields should be private

Mailbox/Person.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
+{
+    public struct Person : IEquatable<Person>
```

Mailbox/DataLoader.cs
```diff
@@ -0,0 +1,58 @@
+﻿using System.Collections.Generic;
+using System.IO;
+using System;
+using Newtonsoft.Json;
+
+namespace Mailbox
+{
+    public class DataLoader : IDisposable
+    {
+        Stream stream;
```
Keboo: Fields should be specified in _PascalCase, or switched to use an auto property. Members should specify access modifiers.

Mailbox/DataLoader.cs
```diff
@@ -0,0 +1,58 @@
...
+        public DataLoader(System.IO.Stream source)
+        {
+            if (source is null) throw new ArgumentNullException();
+            stream = source;
+        }
+
+        public void Dispose()
```

Mailbox/DataLoader.cs
```diff
@@ -0,0 +1,58 @@
...
+using Newtonsoft.Json;
+
+namespace Mailbox
+{
+    public class DataLoader : IDisposable
+    {
+        Stream stream;
+        public DataLoader(System.IO.Stream source)
+        {
+            if (source is null) throw new ArgumentNullException();
```
Keboo: The parameter passed to this exception's constructor should be the parameter name
```suggestion
            if (source is null) throw new ArgumentNullException(nameof(source));
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        public void SaveToStream_LoadFromStream_CompareLists()
+        {
+            DataLoader dl = new DataLoader(File.Open("Mailboxes.json", FileMode.OpenOrCreate, FileAccess.ReadWrite));
```
Keboo: DataLoader is disposable and should be cleaned up.
```suggestion
            using DataLoader dl = new DataLoader(File.Open("Mailboxes.json", FileMode.OpenOrCreate, FileAccess.ReadWrite));
```

Mailbox/Mailbox.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace Mailbox
+{
+    public class Mailbox
+    {
+        public Size MailboxSize { get; set; }
+        public (int x, int y) Location { get; set; }
+        public Person Owner { get; set; }
```
Keboo: Properties can be made readonly.

Mailbox/Mailbox.cs
```diff
@@ -0,0 +1,31 @@
...
+
+        public Mailbox(Size size, int locX, int locY, Person owner)
+        {
+            MailboxSize = size; Location = (locX, locY); Owner = owner;
+        }
+
+        public override string ToString()
+        {
+            string s = Enum.GetName(typeof(Size), MailboxSize);
```
Keboo: The value being set here is then properly overridden in the next line.

Mailbox/Program.cs
```diff
@@ -0,0 +1,171 @@
...
+            if(owners.Count > 0)
+            {
+                foreach(string name in owners)
+                {
+                    s += (name + "\n");
+                }
+            }
+
+            return s;
+        }
```
Keboo: This works but there are simpler options. Take a look at the Assignment5Solution branch for an example.

PR #213, opened 2019-11-04
Assignment5

skozubenko1:


Mailbox/Mailbox.csproj
```diff
@@ -1,8 +1,9 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
```
mmwoodfo:
```suggestion
    <TargetFramework>netcoreapp3.0</TargetFramework>
     <Nullable>enable</Nullable>
```

Mailbox/Program.cs
```diff
@@ -89,17 +90,98 @@ static void Main(string[] args)
...
+            foreach(var own in owners)
+            {
+                sb.AppendLine($"{own._LastName}, {own._FirstName}");
+            }
+
+            return sb.ToString();
         }
 
-        public static string GetMailboxDetails(Mailboxes mailboxes, int x, int y)
+        public static string? GetMailboxDetails(Mailboxes mailboxes, int x, int y)
```
mmwoodfo: Enable nullability to use this syntax.

Mailbox.Tests/Tests.cs
```diff
@@ -0,0 +1,92 @@
...
+        }
+
+        [TestMethod]
+        public Mailboxes CreateTestMailboxes()
+        {
+            Mailboxes mailboxes = new Mailboxes(new List<Mailbox>(), 30, 10);
+
+            for (int i = 0; i < 30; i++)
+            {
+                if (Mailboxes.AddNewMailbox(mailboxes, "12345" + i, "54321" + i, Size.Small) is Mailbox mailbox)
```
mmwoodfo: Mailboxes doesn't contain this method, which is probably why your test cases are erroring out. This occurs in a few other tests in this file, but VS will point those out to you.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,71 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream source;
```
mmwoodfo:
```suggestion
        private readonly Stream _Source;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,71 @@
...
+        #region IDisposable Support
+        private bool disposedValue = false;
         public DataLoader(Stream source)
         {
+            if(source is null)
+            {
+                throw new ArgumentNullException(nameof(source));
+            }
+
+            this.source = source;
```
mmwoodfo:
```suggestion
            _Source = source;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,43 @@
...
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; } //default set to 0 in the enum cant be null
+
+        public (int, int) Location { get; set; } //cant be null because value
+
+        //and a Person Owner
+        public Person Owner; //cant be null because struct
```
mmwoodfo: Did you mean to make this a property with a getter/setter?

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,43 @@
-﻿namespace Mailbox
+﻿using System;
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size Size { get; set; } //default set to 0 in the enum cant be null
```
mmwoodfo:
```suggestion
        public Size Size { get; } //default set to 0 in the enum cant be null
```
Theses could be readonly

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,43 @@
...
+
+        public (int, int) Location { get; set; } //cant be null because value
+
+        //and a Person Owner
+        public Person Owner; //cant be null because struct
+
+        public Mailbox(Size Size, (int, int) Location, Person Owner)
```
mmwoodfo:
```suggestion
        public Mailbox(Size size, (int, int) location, Person owner)
```
Parameters should be camel case

Mailbox/Program.cs
```diff
@@ -89,17 +90,98 @@ static void Main(string[] args)
...
+                }
+            }
+
+            return null;
+
         }
 
         public static Mailbox AddNewMailbox(Mailboxes mailboxes, string firstName, string lastName, Size size)
         {
-
```
mmwoodfo: This method should use the `mailboxes.GetAdjacentPeople` method to check if space is already occupied AND to check if people with the same name are in an adjacent space. This method is overwriting existing mailboxes & adding similarly named people next to each other.

skozubenko1: Thank you for all the corrections. I will try to get to them as best as I can! :)

markdietzler:
You are getting a argumentnullexception on your Dataloader.cs, line 22:

returnMe = JsonConvert.DeserializeObject<Mailbox[,]>(streamReader.ReadLine());

Try refactoring your catch block to a general exception and adding an if statement for both the Json error and the argument null exception.

} catch (Exception exceptionCaught) {
    if(exceptionCaught is JsonReaderException || exceptionCaught is ArgumentNullException)
        {
            return null;
        }
        throw;
    }
}

Mailbox/DataLoader.cs
skozubenko1: it does not like making it _Source

Mailbox/Mailbox.cs
skozubenko1: I think All should be able to have a get and a set in case the order changes. But I did make your suggested correction in my Assignment_5 branch

Mailbox/Mailbox.cs
skozubenko1: I have two different branches going on for this assignment. I made your corrections in one but not the other. waiting to merge them correctly. Thank you for your input!

Mailbox/Mailbox.cs
Fields should be private

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,43 @@
...
+                case Size.Small:
+                case Size.Medium:
+                case Size.Large:
+                    size = this.Size.ToString();
+                    break;
+            }
+
+            if(this.Size.IsPremium())
+            {
+                if(size != "")
```
Keboo: Test for empty strings using 'string.Length' property or 'string.IsNullOrEmpty' method instead of an Equality check.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,69 @@
...
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream source;
+        #region IDisposable Support
+        private bool disposedValue = false;
```

Mailbox/Mailboxes.cs
```diff
@@ -59,5 +59,80 @@ public bool GetAdjacentPeople(int x, int y, out HashSet<Person> adjacentPeople)
...
+                }
+            }
+            return false;
+        }
+        public static Mailbox AddNewMailbox( Mailboxes mailboxes, string firstName, string lastName, Size size)
+        {
+            int x = -1, y = -1;
+            bool found = false;
+
+            if (mailboxes.Count < 1)
```
Keboo: Arguments should be null checked

Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        public string _FirstName; //{ get; set; }
+        public string _LastName; //{ get; set; }
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        public string _FirstName; //{ get; set; }
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
...
+        public string _LastName; //{ get; set; }
+
+        public Person(string _FirstName, string _LastName)
```
Keboo: In member Mailbox.Person.Person(string, string), remove the underscores from parameter name _FirstName.
In member Mailbox.Person.Person(string, string), remove the underscores from parameter name _LastName.

```suggestion
        public Person(string firstName, string lastName)
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
```

Mailbox.Tests/Tests.cs
```diff
@@ -0,0 +1,91 @@
...
+
+            Assert.AreEqual(p1.Equals(p2_same), true);
+            Assert.AreEqual(p1.Equals(p3_difFname), false);
+            Assert.AreEqual(p1.Equals(p4_diflname), false);
+        }
+
+        [TestMethod]
+        public void TestSaveAndLoad()
+        {
+            var FileName = "test.json";
```
Keboo: Local variables should be camelCase.
```suggestion
            var fileName = "test.json";
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,33 @@
...
+
+        public Person(string _FirstName, string _LastName)
+        {
+            this._FirstName = _FirstName ?? throw new ArgumentNullException(nameof(_FirstName));
+            this._LastName = _LastName ?? throw new ArgumentNullException(nameof(_LastName));
+        }
+
+        public bool Equals([AllowNull] Person other)
+        {
+            if (ReferenceEquals(this, other))
```
Keboo: Because Person is a struct and ReferenceEquals parameters are `object` this will box the value. This will give both `this` and `other` different reference. So this check will never be true.

Mailbox/Program.cs
```diff
@@ -89,17 +90,60 @@ static void Main(string[] args)
...
+            }
+
+            StringBuilder sb = new StringBuilder();
+
+            foreach(var own in owners)
+            {
+                sb.AppendLine($"{own._LastName}, {own._FirstName}");
+            }
+
+            return sb.ToString();
```

Mailbox/Size.cs
```diff
@@ -1,4 +1,22 @@
-﻿namespace Mailbox
-{
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
 
+namespace Mailbox
+{
+    public static class SizeExt
```
Keboo: In general you want to avoid abbreviations for identifier names.
```suggestion
    public static class SizeExtensions
```

Mailbox.Tests/Tests.cs
```diff
@@ -0,0 +1,91 @@
...
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+     public class Tests
```
Keboo: For unit testing you want to separate out the tests into separate classes based upon what is being tested. `PersonTests`, `MailboxTests`, etc.

Mailbox.Tests/Tests.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            string lname = "Kozubenko";
+            Person p1 = new Person("Alla", lname);
+            Person p2_same = new Person("Alla", lname);
+            Person p3_difFname = new Person("Sveta", lname);
+            Person p4_diflname = new Person("Alla", "Jacobs");
+
+            Assert.AreEqual(p1.Equals(p2_same), true);
+            Assert.AreEqual(p1.Equals(p3_difFname), false);
+            Assert.AreEqual(p1.Equals(p4_diflname), false);
```
Keboo: Typically you would separate each of these tests into separate unit tests.

PR #215, opened 2019-11-05
Assignment5

NikolasGraham:
Assignment 5 finished - no extra credit.


Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void DataLoader_Load_StreamIsNull_ThrowNullExeption()
```
AaronSauther: No need to have "DataLoader" in the name `DataLoader_Load_StreamIsNull_ThrowNullExeption()` since it is in the `DataLoaderTests` class. Here are [Best Practices](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices) for writing unit tests if your curious, really helped me out earlier.
```suggestion
        public void Load_StreamIsNull_ThrowNullExeption()
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,103 @@
...
+            DataLoader dL = new DataLoader(stream);
+            List<Mailbox> worked = dL.Load();
+
+            stream.Close();
+            File.Delete(filePath);
+            Assert.IsNull(worked);
+        }
+
+        [TestMethod]
+        public void DataLoader_Load()
```
AaronSauther: You might want the name of your test(s) to consist of three parts:
- The name of the method being tested.
- The scenario under which it's being tested.
- The expected behavior when the scenario is invoked.
```suggestion
        public void Load_GeneratedFile_ReturnsCorrectStringRepresentation()
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,103 @@
...
+        {
+            string filePath = Path.GetRandomFileName();
+            Stream stream = GenerateJsonFile(filePath);
```
AaronSauther: I believe we are supposed to use a [MemoryStream](https://docs.microsoft.com/en-us/dotnet/api/system.io.memorystream?view=netframework-4.8) to unit test the IO portion of our load and save methods
```suggestion
            MemoryStream memoryStream = new MemoryStream();
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,103 @@
...
+                new Mailbox(Sizes.Large, (25, 7), new Person("firstName4", "lastName4")),
+                new Mailbox(Sizes.LargePremium, (30, 10), new Person("firstName5", "lastName5"))
+            };
+        }
+
+        private Stream GenerateJsonFile(string filePath)
+        {
+            Stream source = File.Open(filePath, FileMode.OpenOrCreate);
+
+            using(StreamWriter writer = new StreamWriter(source, leaveOpen: true))
```
AaronSauther: Nicely done on the StreamWriter constructor!

Mailbox.Tests/PersonTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            bool worked = person1.Equals(person2);
+
+            Assert.IsTrue(worked);
+        }
+
+        [TestMethod]
+        [DataRow("firstName", "lastName")]
+        [DataRow("firstName2", "lastName2")]
+        [DataRow("firstName3", "lastName3")]
+        public void Person_EqualsOperator_PeopleAreNotEqual(string firstName, string lastName)
```
AaronSauther: Great tests and code coverage!

Mailbox.Tests/DataLoaderTests.cs
NikolasGraham: Thank you! Totally forgot when I was writing my tests.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+                foreach (Mailbox mB in mailboxes)
+                {
+                    string json = JsonConvert.SerializeObject(mB);
+                    writer.WriteLine(json);
+                }
+            }
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false; // To detect redundant calls
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,41 @@
...
+        public override bool Equals(object obj)
+        {
+            if (obj is Mailbox) return this.Equals((Mailbox)obj);
+            return false;
+        }
 
+        public bool Equals(Mailbox other)
+        {
+            return MailSize == MailSize &&
+                Location == other.Location &&
```
Keboo: Arguments should be null checked

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Load_StreamIsNull_ThrowNullException()
+        {
+            DataLoader dL = new DataLoader(null);
+        }
+
+        [TestMethod]
+        public void Load_NotValidJson_ReturnsNull()
+        {
+            MemoryStream stream = GenerateTextFile();
```
```suggestion
            using MemoryStream stream = GenerateTextFile();
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            DataLoader dL = new DataLoader(stream);
```
Keboo: DataLoader is disposable and should be cleaned up.
```suggestion
            using DataLoader dL = new DataLoader(stream);
```

Mailbox.Tests/PersonTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            bool worked = person1 == person2;
+
+            Assert.IsFalse(worked);
+        }
+
+        [TestMethod]
+        public void Person_EqualsOperator_NullCheck()
+        {
+            Person person1 = new Person("firstName", "lastName");
+            Object obj = null;
```
Keboo: Use keyword instead of type name.
```suggestion
            object obj = null;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,91 @@
...
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    // TODO: dispose managed state (managed objects).
+                    Source?.Dispose();
```
Keboo: Since you null checked `Source` on the way in and it is a readonly on property. You don't need the null check here.
```suggestion
                    Source.Dispose();
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,41 @@
-﻿namespace Mailbox
+﻿using System;
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Sizes MailSize { get; set; }
+        public (int x, int y) Location { get; set; }
+        public Person Owner { get; set; }
```
Keboo: Properties can be readonly

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,41 @@
...
+                throw new ArgumentException("MailSize must have a premium size, not premium!");
+            }
+            MailSize = mailSize;
+            Location = location;
+            Owner = owner;
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (obj is Mailbox) return this.Equals((Mailbox)obj);
```
Keboo: Pattern matching would be a little cleaner here.
```suggestion
            return obj is Mailbox mailbox and Equals(mailbox);
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,41 @@
...
+        {
+            return MailSize == MailSize &&
+                Location == other.Location &&
+                Owner == Owner;
+        }
+        public override string ToString()
+        {
+            string mailSize = $"{MailSize}";
+            if(MailSize == Sizes.Default) { mailSize = ""; }
+            return $"{Owner.ToString()} owns mailbox {Location.x},{Location.y} with size '{mailSize}'";
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,46 @@
...
+        }
+
+        public bool Equals([AllowNull] Person other)
+        {
+            return FirstName == other.FirstName && LastName == other.LastName;
+        }
+
+        public static bool operator ==(Person a, Person b)
+        {
+            if (ReferenceEquals(a, b)) return true;
```
Keboo: Because `Person` is a struct and ReferenceEquals parameters are of type `object` it will box both `a` and `b` giving each of them their own reference. So this will never be true.

PR #216, opened 2019-11-05
Assignment5

pnotequalnp:
WIP


pnotequalnp:
Some commits are out of order because of system clock troubles...

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
+            if (mailboxes is null) throw new ArgumentNullException(nameof(mailboxes));
+
+            Source.Position = 0;
+            using var writer = new StreamWriter(Source, leaveOpen: true);
+            writer.WriteLine(JsonConvert.SerializeObject(mailboxes));
         }
 
-        public void Save(List<Mailbox> mailboxes)
+        #region IDisposable Support
+        private bool disposedValue = false;
```

Mailbox/Program.cs
```diff
@@ -43,9 +44,9 @@ static void Main(string[] args)
                         Console.WriteLine("Enter the last name");
                         string lastName = Console.ReadLine();
                         Console.WriteLine("What size?");
-                        if (!Enum.TryParse(Console.ReadLine(), out Size size))
+                        if (!Enum.TryParse(Console.ReadLine(), out Sizes size))
                         {
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,32 @@
...
+        {
+            FirstName = firstName ?? throw new ArgumentNullException(nameof(firstName));
+            LastName = lastName ?? throw new ArgumentNullException(nameof(lastName)); ;
+        }
+
+        public override bool Equals(object? obj) =>
+            obj is Person other && Equals(other);
+
+        public bool Equals(Person other) =>
+            (FirstName, LastName) == (other.FirstName, other.LastName);
```
Keboo: Nice use of tuples for the comparison

Mailbox/Sizes.cs
```diff
@@ -0,0 +1,27 @@
...
+        Medium    = 0b0010,
+        Large     = 0b0100,
+        Premium   = 0b1000,
+    }
+
+    public static class SizesExtensions
+    {
+        public static Sizes Size(this Sizes sizes) => sizes & ~Sizes.Premium;
+
+        public static bool Premium(this Sizes sizes) => sizes.HasFlag(Sizes.Premium);
```
Keboo: Similar to the coding standard for properties, you typically would do the same for methods.
`DO favor prefixing Boolean properties with “Is,” “Can,” or “Has,” when that practice adds value.`

```suggestion
        public static bool IsPremium(this Sizes sizes) => sizes.HasFlag(Sizes.Premium);
```

PR #217, opened 2019-11-05
Assignment 5

kenny2892:


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,63 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream Stream;
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,63 @@
...
+            if (mailboxes is null || mailboxes.Count == 0)
+                return;
+
+            Stream.Position = 0;
+            using StreamWriter sw = new StreamWriter(Stream, leaveOpen: true);
+            foreach (Mailbox mail in mailboxes)
+                sw.WriteLine(JsonConvert.SerializeObject(mail));
+        }
+
+        public void Dispose()
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,22 @@
...
+        public Sizes Size { get; set; }
+        public (int X, int Y) Location;
+        public Person Owner { get; set; }
 
+        public override string ToString()
+        {
+            string sizeStr = Size.ToString();
+
+            if (sizeStr.Contains("Premium"))
+                sizeStr.Replace("Premium", " Premium");
```
Keboo: The Replace method returns a new string, it does not mutate the `sizeStr`

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,22 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Sizes Size { get; set; }
+        public (int X, int Y) Location;
```
Fields should be private

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Constructor_PassNull_ThrowException()
+        {
+            DataLoader dl = new DataLoader(null);
+            dl.Dispose();
```
Keboo: Because the constructor throws, this Dispose call will never be called.

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+        public void Save_PassNull()
+        {
+            MemoryStream stream = new MemoryStream();
+            DataLoader dl = new DataLoader(stream);
+
+            dl.Save(null);
+
+            Assert.AreEqual(0, dl.Load().Count);
+
+            stream.Close();
```
Keboo: If the assert method throws and exception (which is what happens when it doesn't pass). It is better to mark the disposable items with a `using` statement.
```C#
            using MemoryStream stream = new MemoryStream();
            using DataLoader dl = new DataLoader(stream);
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,95 @@
...
+                Assert.AreEqual(mailboxes[i].Owner, loaded[i].Owner);
+                Assert.AreEqual(mailboxes[i].Size, loaded[i].Size);
+                Assert.AreEqual(mailboxes[i].Location, loaded[i].Location);
+            }
+
+            stream.Close();
+            dl.Dispose();
+        }
+
+        private List<Mailbox> getTestMail()
```
Keboo: Method names should be PascalCase
```suggestion
        private List<Mailbox> GetTestMail()
```

Mailbox.Tests/MailboxTests.cs
```diff
@@ -0,0 +1,40 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Collections.Generic;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class MailboxTests
+    {
+        [TestMethod]
+        public void toString_Success()
```
Keboo:
```suggestion
        public void ToString_Success()
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,63 @@
...
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        private readonly Stream Stream;
+
         public DataLoader(Stream source)
         {
+            if (source is null)
+                throw new ArgumentNullException($"{nameof(source)} was found to be null.");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

Mailbox/Person.cs
```diff
@@ -1,4 +1,49 @@
...
+        }
+
+        public override int GetHashCode()
+        {
+            return (FirstName, LastName).GetHashCode();
+        }
+
+        public static bool operator ==(Person left, Person right)
+        {
+            if (ReferenceEquals(left, right)) return true;
```
Keboo: This will never be true. The `ReferenceEquals` method takes in `object` for its parameter types. This will box `left` and `right` into their own references.

Mailbox/Program.cs
```diff
@@ -89,17 +89,73 @@ static void Main(string[] args)
...
+
+            return result;
         }
 
-        public static Mailbox AddNewMailbox(Mailboxes mailboxes, string firstName, string lastName, Size size)
+        public static Mailbox AddNewMailbox(Mailboxes mailboxes, string firstName, string lastName, Sizes size)
         {
-            
+            if (mailboxes is null)
+                throw new ArgumentNullException($"{nameof(mailboxes)} was found to be null.");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message
```suggestion
                throw new ArgumentNullException(nameof(mailboxes));
```

PR #219, opened 2019-11-05
Assignment 5

crackerJackTech:
In the DataLoaderTests file, I could not figure out how to properly test the Load() and Save() methods. I commented-out my attempt to try and create a test using a memory stream.


Mailbox/Program.cs
```diff
@@ -89,17 +88,91 @@ static void Main(string[] args)
...
+            {
+                return "There are no mailbox owners";
+            }
+
+            string ownersOfMailboxes = "";
+            foreach (Mailbox mailbox in mailboxes)
+            {
+                ownersOfMailboxes += mailbox.ToString() + "\n";
+            }
+            return ownersOfMailboxes;
```

PR #220, opened 2019-11-05
Completed Assignment 5

matHopg:
Got everything completed. Feedback for dataloader tests would be appreciated as I feel I could get more code coverage. Thanks

### DIFF
--- a/Mailbox.Tests/DataLoaderTests.cs
+++ b/Mailbox.Tests/DataLoaderTests.cs
@@ -1,0 +1,56 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Mailbox.Tests
+{
+    [TestClass]
+    public class DataLoaderTests
+    {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Load_StreamIsNull_ThrowException()
+        {
+            DataLoader dataLoader = new DataLoader(null);
+        }
+        [TestMethod]
+        public void DataLoader_Save_Success()
+        {
+            List<Mailbox> boxList = new List<Mailbox>();
+            MemoryStream memoryStream = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            boxList.Add(new Mailbox(Sizes.Small, new ValueTuple<int, int>(1, 1), new Person("New", "Person")));
+            dataLoader.Save(boxList);
+            dataLoader.Dispose();
+        }
+
+        [TestMethod]
+        public void Load_ReturnMailboxes()
+        {
+            List<Mailbox> boxList = new List<Mailbox>();
+            MemoryStream memoryStream = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            boxList.Add(new Mailbox(Sizes.Small, new ValueTuple<int, int>(1, 1), new Person("New", "Person")));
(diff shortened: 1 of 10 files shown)


Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,80 @@
...
+                foreach(Mailbox mailbox in mailboxes)
+                {
+                    string json = JsonConvert.SerializeObject(mailbox);
+                    streamWriter.WriteLine(json);
+                }
+            }
+        }
+
+        #region IDisposable Support
+        private bool disposedValue = false;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,48 @@
...
+            if (obj is Mailbox)
+            {
+                return this.Equals((Mailbox)obj);
+            }
+            return false;
+        }
+
+        public bool Equals(Mailbox box)
+        {
+            return MailSize == MailSize && Location == box.Location && Owner == Owner;
```
Keboo: Arguments should be null checked.
```suggestion
            return MailSize == box?.MailSize && Location == box.Location && Owner == box.Owner;
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,56 @@
...
+        {
+            DataLoader dataLoader = new DataLoader(null);
+        }
+        [TestMethod]
+        public void DataLoader_Save_Success()
+        {
+            List<Mailbox> boxList = new List<Mailbox>();
+            MemoryStream memoryStream = new MemoryStream();
+            DataLoader dataLoader = new DataLoader(memoryStream);
+            boxList.Add(new Mailbox(Sizes.Small, new ValueTuple<int, int>(1, 1), new Person("New", "Person")));
```
Keboo: ValueTuple syntax can be simplified
```suggestion
            boxList.Add(new Mailbox(Sizes.Small, (1, 1), new Person("New", "Person")));
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,56 @@
...
+            boxList.Add(new Mailbox(Sizes.Small, new ValueTuple<int, int>(1, 5), new Person("New", "Person3")));
+            dataLoader.Save(boxList);
+            boxList = dataLoader.Load();
+            Assert.AreEqual(boxList.Count, 3);
+            Assert.IsTrue(boxList.Contains(new Mailbox(Sizes.Small, new ValueTuple<int, int>(1, 3), new Person("New", "Person2"))));
+        }
+        [TestMethod]
+        public void DataLoader_Load_JsonReaderException()
+        {
+            MemoryStream memoryStream = new MemoryStream();
```
```suggestion
            using MemoryStream memoryStream = new MemoryStream();
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,80 @@
...
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    Source?.Dispose();
```
Keboo: Since you checked `Source` for `null` in the constructor and it is readonly the null check is not needed here.
```suggestion
                    Source.Dispose();
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,48 @@
...
+            Owner = owner;
+        }
+        public override string ToString()
+        {
+            string mailSize = $"{MailSize}";
+            if (MailSize == Sizes.Default)
+            {
+                mailSize = "";
+            }
+            return $"{Owner.ToString()} owns mailbox {Location.one}, {Location.two} with size '{mailSize}'";
```

PR #221, opened 2019-11-05
Assignment 5 completed

Tpirtle:


Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        public string FirstName { get; set; }
+        public string LastName { get; set; }
+
+        public Person(string _FirstName, string _LastName)
```
crackerJackTech: Incoming parameters should be Camal case

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
+                {
+                    Mailbox mailbox = JsonConvert.DeserializeObject<Mailbox>(line);
+                    mailboxes.Add(mailbox);
+                }
+                return mailboxes;
+            }
+            catch (JsonReaderException)
+            {
+                return null;
+            }
```
crackerJackTech: The StreamReader needs to be Disposed at the end of the Load method and the StreamWriter should also be Disposed at the end of the Save method

Mailbox/Program.cs
```diff
@@ -89,17 +89,61 @@ static void Main(string[] args)
...
-            
+            if (mailboxes == null)
+                throw new ArgumentNullException("Mailboxes is null");
+
+            string list = "";
+            foreach(Mailbox mailbox in mailboxes)
+            {
+                list += mailbox.Owner.ToString()+"\n";
+            }
+            return list;
```
NikolasGraham: From my understanding he was asking for 'distinct' owners. Meaning if an owner owned 2 mailboxes, to only return their name once, not twice. You're just going through every mailbox and adding their name regardless of who they are, even if they are already on the list.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,27 @@
-﻿namespace Mailbox
+﻿using System;
+
+namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Size _Size { get; set; }
```
Keboo: Properties should be PascalCase

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
+using System;
+using System.Collections.Generic;
 using System.IO;
 
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        public Stream stream { get; set; }
```
Keboo: Properties should be PascalCase
```suggestion
        public Stream Stream { get; }
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
+            stream.Position = 0;
+
+            StreamWriter writer = new StreamWriter(stream);
+            foreach (Mailbox mailbox in mailboxes)
+            {
+                writer.Write(JsonConvert.SerializeObject(mailbox));
+            }
+        }
+
+        public void Dispose()
```

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
 namespace Mailbox
 {
-    public class DataLoader
+    public class DataLoader : IDisposable
     {
+        public Stream stream { get; set; }
         public DataLoader(Stream source)
         {
+            if (source == null)
+                throw new ArgumentNullException("The source is null");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,58 @@
...
+            {
+                return null;
+            }
         }
 
         public void Save(List<Mailbox> mailboxes)
         {
-            
+            if (mailboxes == null)
+                throw new ArgumentNullException("mailboxes are null");
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+        public bool Equals([AllowNull] Person other)
+        {
+            return FirstName == other.FirstName && LastName == other.LastName;
+        }
+
+        public override bool Equals(object? obj)
+        {
+            if (obj is null)
+            {
+                throw new ArgumentNullException(nameof(obj));
```
Keboo: Equals creates an exception of type ArgumentNullException. Exceptions should not be raised in this type of method. If this exception instance might be raised, change this method's logic so it no longer raises an exception.

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,27 @@
...
+        public (int x, int y) Location { get; set; }
+        public Person Owner { get; set; }
+
+        public Mailbox(Size size, ValueTuple<int,int> location, Person owner)
```
```suggestion
        public Mailbox(Size size, (int,int) location, Person owner)
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            }
+            dataLoader.Dispose();
+        }
+
+        public List<Mailbox> getTestData()
+        {
+            List<Mailbox> mailboxes = new List<Mailbox>();
+
+            //Locations for the mailboxes
+            ValueTuple<int, int> loc = (0, 0);
```
Keboo: ValueTuple syntax can be simplified
```suggestion
           (int, int) loc = (0, 0);
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+    [TestClass]
+    public class DataLoaderTests
+    {
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ConstructorPassedNull()
+        {
+            DataLoader dataLoader = new DataLoader(null);
+            dataLoader.Dispose();
```
Keboo: If the constructor throws, you will never get to this Dispose

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void SavePassedNull()
+        {
+            MemoryStream stream = new MemoryStream();
```
```suggestion
            using MemoryStream stream = new MemoryStream();
```

Mailbox.Tests/DataLoaderTests.cs
```diff
@@ -0,0 +1,69 @@
...
+            DataLoader dataLoader = new DataLoader(stream);
```
Keboo: Rather than trying to dispose this manually, it would be better to dispose it with a using statement.
```suggestion
            using DataLoader dataLoader = new DataLoader(stream);
```

Mailbox.Tests/ProgramTests.cs
```diff
@@ -0,0 +1,130 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void AddMailbox_nullLastName()
+        {
+            Mailboxes mailboxes = new Mailboxes(getTestData(), 50, 10);
+            Program.AddNewMailbox(null, "Jim", null, Size.Small);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void getMailBoxDetails_NullMailboxes()
```
Keboo: Method names should be CamelCase
```suggestion
        public void GetMailBoxDetails_NullMailboxes()
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,27 @@
...
+            Owner = owner;
+        }
 
+        public override string ToString()
+        {
+            if (_Size == Size.Unspecified)
+            {
+                return "";
+            }
+            return $"Owner: {Owner.LastName} {Owner.FirstName}, Location: {Location.x},{Location.y}, Size: {_Size}";
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+            }
+
+            return Equals(obj as Person?);
```
Keboo: This method is recursive because `Person?` will call back into itself.

Mailbox/Person.cs
```diff
@@ -1,4 +1,52 @@
...
+        }
+
+        public override string ToString()
+        {
+            return $"{FirstName} {LastName}";
+        }
+
+        public static bool operator ==(Person a, Person b)
+        {
+            if (ReferenceEquals(a, b)) return true;
```
Keboo: This will actually never match. The parameters of `ReferenceEquals` are `object`. So it boxes the values into their own references.

PR #222, opened 2019-11-06
Submit Assignment 5

markdietzler:
Write branch this time.

### DIFF
--- a/Mailbox/DataLoader.cs
+++ b/Mailbox/DataLoader.cs
@@ -1,21 +1,61 @@
-﻿using System.Collections.Generic;
+﻿using Newtonsoft.Json;
+using System;
 using System.IO;
 
-namespace Mailbox
-{
-    public class DataLoader
-    {
-        public DataLoader(Stream source)
-        {
+namespace Mailbox {
+    public class DataLoader : IDisposable {
+        bool disposedFlag = false;
+
+        private readonly Stream _Source;
+        public DataLoader(Stream newStreamSource) {
+            _Source = newStreamSource ?? throw new ArgumentNullException($"{nameof(newStreamSource)} was null.");
         }
 
-        public List<Mailbox> Load()
-        {
+        public Mailbox[,] Load() {
+
+            _Source.Position = 0;
+            Mailbox[,] returnMe;
+
+
+            using (StreamReader streamReader = new StreamReader(_Source, leaveOpen: true)) {
+                try {
+                    returnMe = JsonConvert.DeserializeObject<Mailbox[,]>(streamReader.ReadLine());
+                } catch (Exception exceptionCaught) {
+                    if(exceptionCaught is JsonReaderException || exceptionCaught is ArgumentNullException)
+                    {
+                        return null;
+                    }
+                    throw;
+                }
+            }
+            return returnMe;
         }
 
-        public void Save(List<Mailbox> mailboxes)
+        public void Save(Mailbox[,] mailboxes)
         {
-            
+            _Source.Position = 0;
+            using (StreamWriter streamWriter = new StreamWriter(_Source, leaveOpen: true))
+            {
+                streamWriter.WriteLine(JsonConvert.SerializeObject(mailboxes));
+            }
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposedFlag)
+                return;
+            if(disposing)
+            {
+                _Source.Dispose();
+            }
+
+            disposedFlag = true;
         }
     }
 }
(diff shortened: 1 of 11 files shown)


Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,20 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Sizes mSize { get; }
+        public (int, int) mLocation { get; }
+        public Person mOwner { get; set; }
```
samrnixon: How come mOwner is also "settable"? I think it can be read only.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,61 @@
...
 using System.IO;
 
-namespace Mailbox
-{
-    public class DataLoader
-    {
-        public DataLoader(Stream source)
+namespace Mailbox {
+    public class DataLoader : IDisposable {
+        bool disposedFlag = false;
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,20 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Sizes mSize { get; }
```
Keboo: Properties should be PascalCase

Mailbox/Mailbox.cs
Keboo: Properties should be PascalCase

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,20 @@
...
+        public (int, int) mLocation { get; }
```
Keboo: Properties should be PascalCase

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,20 @@
...
+        public Person mOwner { get; set; }
+        public Mailbox(Sizes newSize, (int, int) newLocation, Person newOwner)
+        {
+            this.mSize = newSize;
+            this.mLocation = mLocation;
```
Keboo: The property mLocation should not be assigned to itself.

Mailbox/Mailboxes.cs
```diff
@@ -1,63 +1,57 @@
-﻿using System;
-using System.Collections.Generic;
-
-namespace Mailbox
+﻿namespace Mailbox
 {
-    public class Mailboxes : List<Mailbox>
+    public class Mailboxes
     {
-        public Mailboxes(IEnumerable<Mailbox> collection, int width, int height)
```
Keboo: Properties should be PascalCase

Mailbox/Person.cs
```diff
@@ -1,4 +1,27 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
+    {
+        public string _FirstName;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,27 @@
...
+        public string _LastName;
```

Mailbox/Person.cs
```diff
@@ -1,4 +1,27 @@
-﻿namespace Mailbox
+﻿using System;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Mailbox
 {
+    public struct Person : IEquatable<Person>
```

Mailbox/Mailbox.cs
```diff
@@ -1,4 +1,20 @@
 ﻿namespace Mailbox
 {
+    public class Mailbox
+    {
+        public Sizes mSize { get; }
+        public (int, int) mLocation { get; }
+        public Person mOwner { get; set; }
+        public Mailbox(Sizes newSize, (int, int) newLocation, Person newOwner)
```
Keboo: Parameter newLocation of method .ctor is never used. Remove the parameter or use it in the method body.

Mailbox/DataLoader.cs
```diff
@@ -1,21 +1,61 @@
...
+
+            _Source.Position = 0;
+            Mailbox[,] returnMe;
+
+
+            using (StreamReader streamReader = new StreamReader(_Source, leaveOpen: true)) {
+                try {
+                    returnMe = JsonConvert.DeserializeObject<Mailbox[,]>(streamReader.ReadLine());
+                } catch (Exception exceptionCaught) {
+                    if(exceptionCaught is JsonReaderException || exceptionCaught is ArgumentNullException)
```
Keboo: Rather than this `if` it would be better to have multiple catch blocks for each of these exception types.

Mailbox/Program.cs
```diff
@@ -1,105 +1,107 @@
...
+            string result = "";
+            foreach (Mailbox mailbox in mailboxes.mMailBoxesArray) {
+                if(mailbox is null) {
+                    continue;
+                }
+                if(!result.Contains(mailbox.mOwner.toString())) {
+                    result += mailbox.mOwner.toString() + "\n";
+                }
+            }
+            return result;
```

PR #223, opened 2019-11-06
Assignment6

pnotequalnp:
WIP

### DIFF
--- a/Assignment6.Tests/Array.Tests.cs
+++ b/Assignment6.Tests/Array.Tests.cs
@@ -1,5 +1,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
 using System;
+using System.Collections.Generic;
 using System.Linq;
 
 namespace Assignment6.Tests
@@ -58,6 +59,13 @@ namespace Assignment6.Tests
         }
 
         [TestMethod]
+        [ExpectedException(typeof(InvalidOperationException))]
+        public void Add_NoRoom_ThrowsException()
+        {
+            new Array<int>(0).Add(5);
+        }
+
+        [TestMethod]
         public void Clear_ValidArray_ClearsValues()
         {
             var array = new Array<int>(100) { 1, 2, 3, 4, 5, 6, 7, 8, 9 };
@@ -66,17 +74,7 @@ namespace Assignment6.Tests
 
             Assert.AreEqual(0, array.Count);
             foreach (int i in Enumerable.Range(0, array.Capacity))
-            {
-                try
-                {
-                    _ = array[i];
-                    Assert.Fail();
-                }
-                catch (InvalidOperationException)
-                {
-                    // Exception indicates successful Clear()
-                }
-            }
+                Assert.ThrowsException<InvalidOperationException>(() => array[i]);
         }
 
         [DataTestMethod]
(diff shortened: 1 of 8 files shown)


pnotequalnp:
Finished, I think.

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
-        
+        public int Capacity => Values.Length;
+
+        public int Count => Values.Count(element => !(element is null));
+
+        public bool IsReadOnly => false;
+
+        private ValueTuple<T>?[] Values { get; }
+
+        private IEnumerable<T> Elements =>
```
twofingerrightclick: I am learning a lot from this syntax. Your underlying data structure is a valueTuple array. My first thought was is this only intended to be used to store structs? And if so there is no where to ensure that Items stored in the array a can be compared as structs don't do interfaces, so I am trying to figure out what the contains method would do with a struct that didn't override equals. Or is this not pertinent to the use of the Array<T> ?

pnotequalnp: You can store any type in it (including nullable types). If you wanted to enforce a constraint on the type you'd just do that at the class level generic `T`. But you don't need to do that for contains since `.Equals` is one of the methods on `object` and is guaranteed to exist on every type. For structs that do not implement a custom `.Equals` (which shouldn't happen), it will end up boxing them and never returning true, but there is nothing this class could do to prevent that.

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+    public class Array<T> : ICollection<T>, IReadOnlyArray<T>
     {
-        
+        public int Capacity => Values.Length;
+
+        public int Count => Values.Count(element => !(element is null));
+
+        public bool IsReadOnly => false;
+
+        private ValueTuple<T>?[] Values { get; }
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+
+        private IEnumerable<T> Elements =>
+            Values.Where(element => !(element is null))
+                  .Cast<ValueTuple<T>>()
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+                  .Select(element => element.Item1);
+
+        public Array(int capacity) =>
+            Values = capacity >= 0 ? new ValueTuple<T>?[capacity] : throw new ArgumentException("Capacity cannot be negative.", nameof(capacity));
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+
+        public void Add(T item)
+        {
+            for (int i = 0; i < Values.Length; i++)
+            {
+                if (Values[i] is null)
+                {
+                    Values[i] = new ValueTuple<T>(item);
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+            if (array.Length - arrayIndex < Count) throw new ArgumentException("Not enough space in array for collection.");
+
+            foreach (var element in Elements) array[arrayIndex++] = element;
+        }
+
+        public bool Remove(T item)
+        {
+            for (int i = 0; i < Values.Length; i++)
+            {
+                if (Values[i] is ValueTuple<T> element &&
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Design", "CA1065:Do not raise exceptions in unexpected locations", Justification = "Index-related exceptions make sense in an indexing operation.")]
+        public T this[int index]
+        {
+            get => index switch
+            {
+                int ix when ix < 0 || ix >= Capacity => throw new IndexOutOfRangeException(),
+                int ix => Values[ix] switch
+                {
+                    null => throw new InvalidOperationException("Element does not exist."),
+                    ValueTuple<T> element => element.Item1
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+                }
+            };
+
+            set => Values[index] = index switch
+            {
+                int ix when ix < 0 || ix >= Capacity => throw new IndexOutOfRangeException(),
+                _ => new ValueTuple<T>(value)
```
MarkMichaelis: ValueTuple syntax can be simplified

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+
+        private IEnumerable<T> Elements =>
+            Values.Where(element => !(element is null))
+                  .Cast<ValueTuple<T>>()
+                  .Select(element => element.Item1);
+
+        public Array(int capacity) =>
+            Values = capacity >= 0 ? new ValueTuple<T>?[capacity] : throw new ArgumentException("Capacity cannot be negative.", nameof(capacity));
+
+        public void Add(T item)
```
MarkMichaelis: This functionality is not typical to an array.  Consider implementing ICollection<T>'s Add/Remove methods explicitly instead.

Assignment6/Array.cs
```diff
@@ -1,10 +1,109 @@
...
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            if (array is null) throw new ArgumentNullException(nameof(array));
+            if (arrayIndex < 0) throw new IndexOutOfRangeException();
+            if (array.Length - arrayIndex < Count) throw new ArgumentException("Not enough space in array for collection.");
+
+            foreach (var element in Elements) array[arrayIndex++] = element;
+        }
+
+        public bool Remove(T item)
```
MarkMichaelis: This functionality is not typical to an array.  Consider implementing ICollection<T>'s Add/Remove methods explicitly instead.

Assignment6/Array.cs
MarkMichaelis: Why both the Values and the Elements collections?  Surely this can be implemented with just one?

MarkMichaelis: Why a `ValueTuple<T>` rather than no just and `T[]`?  I can't think of any realworld scenario you would use `ValueTuple<T>` (just one item in the tuple), unless trying to program generically across tuples of different sizes perhaps.

Assignment6/Array.cs
MarkMichaelis: Why the call to `Cast<ValueTuple<T>>()`? Isn't `Values` storing `ValueTuple<T>` so there is no reason to cast?

Assignment6.Tests/Array.Tests.cs
```diff
@@ -0,0 +1,285 @@
...
+        [DataRow(1234)]
+        [DataRow(987654)]
+        [DataRow(314159)]
+        public void Remove_DoesNotContainItem_DoesntModifyArray(int value)
+        {
+            var range = Enumerable.Range(0, 100);
+            var array = new Array<int>(100);
+            foreach (int i in range) array.Add(i);
+
+            Assert.IsFalse(array.Remove(value));
```
MarkMichaelis: Consider using `CollectionAssert` for comparing collections.

Assignment6.Tests/Array.Tests.cs
```diff
@@ -0,0 +1,285 @@
...
+                .Concat(Enumerable.Range(40, 30))
+                .Concat(Enumerable.Range(80, 20));
+            var array = new Array<int>(100);
+            foreach (int i in ranges) array[i] = i;
+            var list = new List<int>();
+
+            foreach (object? i in (System.Collections.IEnumerable)array)
+                list.Add((int)i!);
+
+            var zipped = ranges.Zip(list);
```
MarkMichaelis: Nice use of `Zip()`.

Assignment6/IReadOnlyArray.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace Assignment6
+{
+    interface IReadOnlyArray<out T> : IEnumerable<T>
```
MarkMichaelis: Yes.... out is all you need. :)

Assignment6/Array.cs
MarkMichaelis: ValueTuple<T> adds significant complication without adding any value.  You could use List<T> or T[] and achieve the same purpose IMO.

MarkMichaelis:
##  (pnotequalnp)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), **7 Warning(s)**)| **10** |


| Assignment  | **15*** |
*<sub><sup>Why so many warnings lefts?  Storing `ValueTuple<T>` makes little sense except to over complicate your implementation significantly.</sub></sup>

| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **✔** |


| Testing | **20** |
| Test Run Succeeded: Total tests=76, Passed=76, Failed=0, Skipped=0 | **98%** |


| Extra Credit | **5** |

PR #226, opened 2019-11-07
Assignment6 : mmwoodfo

mmwoodfo:


Assignment6/ArrayCollection.cs
```diff
@@ -0,0 +1,89 @@
...
+        public int Capacity { get; }
+
+        public ArrayCollection(int capacity)
+        {
+            if (capacity <= 0)
+            {
+                throw new ArgumentOutOfRangeException(nameof(capacity));
+            }
+            Capacity = capacity;
+            _Items = new List<TCollection>(Capacity);
```
CDobbins1: A List with the capacity property set doesn't prevent items from being added once the capacity is reached. It specifies the number of elements that the List can contain before resizing is required. I suggest adding a test method to check if you can add items beyond the specified size.

mmwoodfo: Thank you for pointing that out. I think I'll take care of that in the add method then.

MarkMichaelis:
Already graded extra credit so closing this pull request.

PR #230, opened 2019-11-08
Assignment 6

JerettLatimer:
Completed Assignment

### DIFF
--- a/Assignment6.Tests/ArrayTests.cs
+++ b/Assignment6.Tests/ArrayTests.cs
@@ -1,0 +1,13 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTests
+    {
+        [TestMethod]
+        public void TestMethod1()
+        {
+        }
+    }
+}
--- a/Assignment6.Tests/ArrayTests.cs
+++ b/Assignment6.Tests/ArrayTests.cs
@@ -5,9 +5,27 @@ namespace Assignment6.Tests
     [TestClass]
     public class ArrayTests
     {
-        [TestMethod]
-        public void TestMethod1()
+        [DataTestMethod]
+        [DataRow(1)]
+        public void Array_Add_AddsSuccessfully(int capacity)
         {
+            // Arrange
+            Array<int> sut = new Array<int>(capacity);
+
+            // Act
+            int i = 0;
+            while(i < capacity)
+            {
+                sut.Add(i);
+                i++;
+            }
+
+            // Assert
+            i = 0;
+            foreach(int value in sut)
+            {
+                Assert.AreEqual(i, value);
+            }
         }
     }
 }
--- a/Assignment6.Tests/ArrayTests.cs
+++ b/Assignment6.Tests/ArrayTests.cs
@@ -1,12 +1,40 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Assignment6.Tests
 {
     [TestClass]
     public class ArrayTests
     {
+        [TestMethod]
+        public void Array_Constructor_CreatesSuccessfully()
+        {
+            // Arrange
+            Array<string> sut = new Array<string>(12);
+
+            // Act
+
+
+            // Assert
+            Assert.IsNotNull(sut);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentException))]
+        public void Array_Constructor_ThrowsException()
+        {
+            // Arrange
+            Array<string> sut = new Array<string>(-12);
+
+            // Act
+
+            // Assert
+        }
+
         [DataTestMethod]
         [DataRow(1)]
+        [DataRow(100)]
+        [DataRow(110000)]
         public void Array_Add_AddsSuccessfully(int capacity)
         {
             // Arrange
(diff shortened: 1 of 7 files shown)


Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+            int capacity = 12;
+            Array<Object> sut = new Array<Object>(capacity);
+            int i = 0;
+            while (i < capacity)
+            {
+                sut.Add(i);
+                i++;
+            }
+
+            Object toRemove = null;
```
MarkMichaelis: Use keyword instead of type name

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
     {
-        
+        private int Capacity { get; set; }
+        private List<T>? Items { get; }
+        public int Count => Items is null ? 0 : Items.Count;
+        public bool IsReadOnly => false;
+
+        public Array(int capacity)
+        {
+            if(capacity < 1)
```
MarkMichaelis: Why is 0 not a valid value?

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+            }
+        }
+
+        public void Add(T item)
+        {
+            if(Capacity == Count)
+            {
+                throw new InvalidOperationException("Array is a at max capacity.");
+            }
+            if(Items is null)
```
MarkMichaelis: If you set Items in your constructor as you do, and the property is read-only as it is, how could it ever be null?

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+            {
+                throw new NullReferenceException("Items is null");
+            }
+            Items.Add(item);
+        }
+
+        public void Clear()
+        {
+            if (Items is null)
```
MarkMichaelis: Again, a seemingly unnecessary check.

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+            Items.Clear();
+        }
+
+        public bool Contains(T item)
+        {
+            if (Items is null)
+            {
+                throw new NullReferenceException("Items is null");
+            }
+            if (Items.Contains(item))
```
MarkMichaelis: Why not just return `Items.Contains(item)` using an expression bodied member in fact:
```
bool Contains(T item) => Items.Contains(item);
```
If you did have to check for null in items, you could use:
```
bool Contains(T item) => Items?.Contains(item)??throw new InvalidOperationException(...);`
```

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+                throw new NullReferenceException("Items is null");
+            }
+            Items.Clear();
+        }
+
+        public bool Contains(T item)
+        {
+            if (Items is null)
+            {
+                throw new NullReferenceException("Items is null");
```
MarkMichaelis: Only the framework should ever throw a NullReferenceException.  If this were a valid check, the correct exception would be InvalidOperationExxception, indicating there was something wrong in the programming logic.

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+                throw new ArgumentException("Index must be postive", nameof(arrayIndex));
+            }
+            if(Count > array.Length - arrayIndex + 1)
+            {
+                throw new ArgumentException("Not enough space in array", nameof(array));
+            }
+            Items.CopyTo(array, arrayIndex);
+        }
+
+        public bool Remove(T item)
```
MarkMichaelis: This is an unusual method to have on an array.

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+                throw new ArgumentNullException(nameof(item), "Item can't be null");
+            }
+            if (Items.Remove(item))
+            {
+                return true;
+            }
+            return false;
+        }
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Again, consider using an expression bodied member here.

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
+            {
+                if(Items is null)
+                {
+                    throw new NullReferenceException("Items is null");
+                }
+                Items[index] = value;
+            }
+        }
+
+        public T[] ToArray()
```
MarkMichaelis: FYI: Not sure you intended it to be, but this is not a cast operator.

MarkMichaelis: FYI: Not sure you intended it to be, but this is not a cast operator.  Regardless, the method is automatically implemented when you use LINQ (which we hadn't covered yet. :)

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>
```
MarkMichaelis: To have to check `Items` in virtually every method indicates you a code smell that you should look into.  Just having `Items` as a property that includes validation (or is read-only like you did) should be sufficient to avoid all the checks for null that you have.

Assignment6/Array.cs
```diff
@@ -1,10 +1,143 @@
...
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>
     {
-        
+        private int Capacity { get; set; }
+        private List<T>? Items { get; }
```
MarkMichaelis: Why is this nullable?  I think it caused all sorts of problems for you that you didn't actually have if the nullability was declared correctly.

Assignment6.Tests/TestSuppressions.cs
```diff
@@ -0,0 +1,6 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "Test class naming conventions allow underscores")]
```
MarkMichaelis: Glad to see a test specific suppression file.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+        public void Array_Constructor_CreatesSuccessfully()
+        {
+            // Arrange
+            Array<string> sut = new Array<string>(12);
+
+            // Act
+
+
+            // Assert
+            Assert.IsNotNull(sut);
```
MarkMichaelis: A constructor either throws an exception or returns an instance.  It never returns null.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+            sut.Add(42);
+
+            // Assert
+
+        }
+
+        [DataTestMethod]
+        [DataRow(-1)]
+        [DataRow(-100)]
+        [DataRow(0)]
```
MarkMichaelis: Why is null not allowed?

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+
+        [TestMethod]
+        [ExpectedException(typeof(NullReferenceException))]
+        public void Array_Add_ItemsIsNull()
+        {
+            // Arrange
+            Array<int> sut = null;
+
+            // Act
+            sut.Add(1);
```
MarkMichaelis: What are you testing here?  Add is never called because `sut` is null.  None of our code is invoked in this test.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+
+        [TestMethod]
+        [ExpectedException(typeof(NullReferenceException))]
+        public void Array_Clear_ItemsIsNull()
+        {
+            // Arrange
+            Array<int> sut = null;
+
+            // Act
+            sut.Clear();
```
MarkMichaelis: Again, none of your code is invoked in this test.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,384 @@
...
+            // Act
+
+            // Assert
+        }
+
+        [DataTestMethod]
+        [DataRow(1)]
+        [DataRow(100)]
+        [DataRow(110000)]
+        public void Array_Add_AddsSuccessfully(int capacity)
```
MarkMichaelis: I would like to have seen a "foreach" test that was more explicit than just buried in the `Add` test.

MarkMichaelis:
##  (JerettLatimer)

## Assignment6

| Assignment  | **10** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |


| Testing | **20** |
| Test Run Succeeded: Total tests=28, Passed=28, Failed=0, Skipped=0 | **80%** |


| Extra Credit | **0** |

PR #233, opened 2019-11-10
Assignment 6

Fervidusletum:
Finished, has casting to array for extra credit.


Assignment6/ArrayCollection.cs
```diff
@@ -0,0 +1,230 @@
...
+    /// A generically typed, fixed length array.
+    /// </summary>
+    /// <remarks>
+    /// Elements are coalesced toward the beginning of the array. Only existing elements can
+    /// be accessed, and any new elements are added at the first available empty index.
+    /// </remarks>
+    /// <typeparam name="T"></typeparam>
+    public class ArrayCollection<T> : ICollection<T>
+    {
+        private T[] Data { get; }
```
jacob-berger: Would be much easier to use a List<T> instead of an array. Many of the functions are already implemented for the List<T> class.

Assignment6/ArrayCollection.cs
```diff
@@ -0,0 +1,230 @@
...
+        private int _Count;
+        public int Count { get => _Count; }
+        public int Capacity { get; }
+        public bool IsReadOnly { get => false; } // readonly not implemented in this class (yet?)
```
jacob-berger: I don't think this was ever really specified in class, so I did the same...?

Assignment6/ArrayCollection.cs
```diff
@@ -0,0 +1,230 @@
...
+        /// </summary>
+        /// <remarks>
+        /// Only allows indexing operations on existing elements, and assignment on first empty index.
+        /// Will throw exceptions for invalid operations.
+        /// </remarks>
+        /// <param name="index">
+        /// Must be nonzero positive int less or equal to Count
+        /// </param>
+        /// <returns></returns>
+#pragma warning disable CA1065 // It's expected that an indexing operation will throw on invalid index
```
jacob-berger: Is it necessary to disable this warning? Genuinely not sure because of your implementation.

Fervidusletum: I figured it was better to disable and justify rather than ignore it, or not throw and rely on array itself to throw.

Assignment6/ArrayCollection.cs
Fervidusletum: Originally I was going to use a List. However, I felt it reduced the assignment to the point it's just letting List handle all the generics, and the only real implementation of the assignment is a fixed-capacity wrapper of List<T>.

MarkMichaelis:
## <Full Name Missing> (Fervidusletum)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), 4 Warning(s))| **10** |


| Assignment  | **20** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **✔** |


| Testing | **20** |
| Test Run Succeeded: Total tests=31, Passed=31, Failed=0, Skipped=0 | **91%** |
| Most assigned work is tested or contain some errors. | **5* |


| Extra Credit | **5** |

PR #235, opened 2019-11-10
Assignment6

Cybotive:


Assignment6/Array.cs
```diff
@@ -1,10 +1,197 @@
...
+        public IEnumerator<T> GetEnumerator()
+        {
+            return new ArrayEnumerator<T>(this);
+        }
+
+        public bool Remove(T item)
+        {
+            if (item is null) { throw new ArgumentNullException(nameof(item)); }
+
+            for (int i = 0; i < Count; i++)
```
GamerBah: Shouldn't `for (int i = 0; i < Count; i++)` be `for (int i = 0; i < Capacity; i++)`?

Cybotive: Yes. I was envisioning all the elements would be together with no null entries separating them, but that's not always the case.

Assignment6/Array.cs
```diff
@@ -1,10 +1,197 @@
...
+
+            return false;
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        internal T GetValue(int index)
```
MarkMichaelis: Why internal rather than private?

Assignment6/Array.cs
```diff
@@ -1,10 +1,197 @@
...
+
+        public T this[int i]
+        {
+            get => GetValue(i);
+            set => SetValue(i, value);
+        }
+    }
+
+
+    public class ArrayEnumerator<T> : IEnumerator<T>
```
MarkMichaelis: It is rarely necessary to implement an enumerator unless you use the yield return statement.

Assignment6/Array.cs
```diff
@@ -1,10 +1,197 @@
...
+            if(capacity < 1) { throw new ArgumentOutOfRangeException(nameof(capacity)); }
+
+            Capacity = capacity;
+            Count = 0;
+            IsReadOnly = false;
+
+            _InternalArray = new T[Capacity];
+        }
+
+        public void Add(T item)
```
MarkMichaelis: Since arrays don't usually have Add and Remove functionality, I would likely have implemented this explicitly using ICollection.Add().

Assignment6Tests/ArrayTests.cs
```diff
@@ -0,0 +1,315 @@
...
+            string concat = "";
+
+            //Act
+            foreach (string str in array)
+            {
+                concat += str;
+            }
+
+            //Assert
+            Assert.AreEqual("ABCDE", concat);
```
MarkMichaelis: Consider using CollectionAssert to compare collections instead of ToString()

Assignment6Tests/ArrayTests.cs
```diff
@@ -0,0 +1,315 @@
...
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void ArrayAddThrowsExceptionWhenNull()
+        {
+            //Arrange
+            Array<string> array = new Array<string>(5);
+
+            //Act
+#pragma warning disable CS8625 // Cannot convert null literal to non-nullable reference type. Needed to test what happens when null
```
MarkMichaelis: I would use the null forgiveness operator (null!) here in your test rather than disable the warning.

MarkMichaelis:
##  (Cybotive)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), 3 Warning(s))| **10** |


| Assignment  | **15** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **✔** |


| Assignment6 | Enable | Assignment6.sln | IntelliTectAnalyzer.dll, Microsoft.CodeAnalysis.VersionCheckAnalyzer.dll, Humanizer.dll, Microsoft.CodeQuality.Analyzers.dll, Microsoft.CodeQuality.CSharp.Analyzers.dll, Microsoft.NetCore.Analyzers.dll, Microsoft.NetCore.CSharp.Analyzers.dll, Microsoft.NetFramework.Analyzers.dll, Microsoft.NetFramework.CSharp.Analyzers.dll | CS1701, CS1702, CS2008 | 4 |
| Assignment6Tests | Enable | Assignment6.sln | IntelliTectAnalyzer.dll, Microsoft.CodeAnalysis.VersionCheckAnalyzer.dll | CS1701, CS1702 | 4 |


| Testing | **20** |
| Test Run Succeeded: Total tests=23, Passed=23, Failed=0, Skipped=0 | **91%** |


| Extra Credit | **0** |

PR #239, opened 2019-11-10
Assignment 6

AaronSauther:


Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+
+    public class Array<T> : ICollection<T>
+#pragma warning restore CA1710
     {
-        
+        private int Capacity { get; }
+
+        public int Count { get; private set; }
+
+        private List<T> Collection { get; } //I used a list under the hood, so I do not have to worry about null f or missing items
```
MarkMichaelis: Consider a name like `InternalCollection`

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+
+        public bool IsReadOnly { get; private set; }
```
MarkMichaelis: You only set this in the constructor, and, therefore, you should remove the set entirely.

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+
+        public Array(int capacity)
+        {
+            Count = 0;
+            Capacity = capacity;
+            Collection = new List<T>(capacity);
+            IsReadOnly = false;
+        }
+
+        public void Add(T item)
```
MarkMichaelis: I would implement this explicitly as part of `ICollection<T>` since it isn't generally possible with an array.

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+        {
+            Collection.Clear();
+            Count = 0;
+        }
+
+        public bool Contains(T item)
+        {
+            if (item == null)
+                throw new ArgumentNullException(nameof(item));
+            if (Collection.Contains(item))
```
MarkMichaelis: Why not just `return (Collection.Contains(item)`

MarkMichaelis: Why not just `return Collection.Contains(item)`

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+                return true;
+            return false;
+        }
+
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            Collection.CopyTo(array, arrayIndex);
+        }
+
+        public bool Remove(T item)
```
MarkMichaelis: It isn't bad, but this isn't generally array type behavior.

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
-    public class Array<T>
+#pragma warning disable CA1710 // This was the name given in the assignment
+
+    public class Array<T> : ICollection<T>
+#pragma warning restore CA1710
     {
-        
+        private int Capacity { get; }
+
+        public int Count { get; private set; }
```
MarkMichaelis: Can't `Count` just return `Collection.Count`, saving you having to maintain it?

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+            if (item is null)
+                throw new ArgumentNullException(nameof(item));
+            if (!Collection.Contains(item))
+                throw new ArgumentException($"{nameof(item)} does not exist");
+            bool result = Collection.Remove(item);
+            Count--;
+            return result;
+        }
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: I recommend using expression bodied members here:
```
public IEnumerator<T> GetEnumerator() => Collection.GetEnumerator();
```

Assignment6/Array.cs
```diff
@@ -1,10 +1,99 @@
...
+        {
+            get
+            {
+                ValidateIndex(i);
+                return Collection[i];
+            }
+            set
+            {
+                ValidateIndex(i);
+                Collection[i] = value;
```
MarkMichaelis: Couldn't you just rely on `List<T>`'s validation and not even check?

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,470 @@
...
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTests
+    {
+        [TestMethod]
+        public void Constructor_FixedWidth_NoErrors()
+        {
+            _ = new Array<int>(5);
```
MarkMichaelis: No exceptions is a valid test but consider testing the capacity or something here.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,470 @@
...
+            int count = array.Count;
+            //Assert
+            Assert.AreEqual(0, count);
+        }
+
+        [TestMethod]
+        public void Add_OneItem_ReturnsCorrectCount()
+        {
+            //Arrange
+#pragma warning disable IDE0028 // Testing add method, dont want to simplify in this case
```
MarkMichaelis: Good call!

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,470 @@
...
+            List<int> recievedData = new List<int>();
+            //Act
+            foreach (int i in array)
+            {
+                recievedData.Add(i);
+            }
+            //Assert
+            Assert.IsTrue(recievedData.Contains(1));
+            Assert.IsTrue(recievedData.Contains(2));
+            Assert.IsTrue(recievedData.Count == 2);
```
MarkMichaelis: Use Assert.AreEqual() here instead - that way if it fails you know the values.

MarkMichaelis:
##  (AaronSauther)

## Assignment6

| Assignment  | **20** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **✔** |


| Testing | **20** |
| Test Run Succeeded: Total tests=38, Passed=38, Failed=0, Skipped=0 | **95%** |


| Extra Credit | **0** |

PR #240, opened 2019-11-11
Assignment6

NikolasGraham:
No extra credit attempted.


Assignment6/Array.cs
```diff
@@ -1,10 +1,116 @@
...
+            {
+                throw new ArgumentNullException(nameof(array), "array null!");
+            }
+
+            if((arrayIndex + Count) > array.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(array), "Not enough space!");
+            }
+
+            foreach(T item in _Items)
```
JerettLatimer: I just used the CopyTo method provided by the List class. Is there a reason you used a foreach loop and index operator?

NikolasGraham: I don't have a good reason for not using what's provided by List, so I modified my code accordingly and it functions the same as before.

Assignment6/Array.cs
```diff
@@ -1,10 +1,117 @@
...
+        }
+
+        public bool Contains(T item)
+        {
+            if (_Items.Contains(item))
+            {
+                return true;
+            }
+            
+            throw new ArgumentException(nameof(item), "Item not in list!");
```
MarkMichaelis: Why throw an exception rather than returning false?  There is no point in a return if the only valid return is true. 
Returnining `_Items.Contains(item)` seems like all you need.

Assignment6/Array.cs
```diff
@@ -1,10 +1,117 @@
...
+
+            if((arrayIndex + Count) > array.Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(array), "Not enough space!");
+            }
+
+            _Items.CopyTo(array, arrayIndex);
+        }
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Consider implementing as an expression bodied member:
```
        public IEnumerator<T> GetEnumerator() => _Items.GetEnumerator();
```

Assignment6/Array.cs
```diff
@@ -1,10 +1,117 @@
...
+        }
+
+        public bool Remove(T item)
+        {
+            if (_Items.Contains(item))
+            {
+                return _Items.Remove(item);
+            }
+
+            throw new ArgumentException(nameof(item), "Item not in list!");
```
MarkMichaelis: Since there is a return of bool, it again it doesn't seem right to throw an exception.  Simply returning `_Items.Remove(item);` seems sufficient.

Assignment6/Array.cs
```diff
@@ -1,10 +1,117 @@
...
+
+        public T this[int index]
+        {
+            get 
+            {
+                T[] tempAra = _Items.ToArray();
+                return tempAra[index]; 
+            }
+            set 
+            {
```
MarkMichaelis: Wait what?  Why instantiate an entirely new List<T> rather than just setting the specific item at the specified index location - and throwing an exception if over capacity?  This implementation doesn't seem to make sense.

Assignment6/Array.cs
```diff
@@ -1,10 +1,117 @@
...
+                _Items.CopyTo(tempAra, 0);
+                tempAra[index] = value;
+                _Items = new List<T>(tempAra);
+            }
+        }
+
+        public override string ToString()
+        {
+            string returnString = "";
+            for(int i=0; i<Count; i++)
```
MarkMichaelis: Use something like `string.Join(" ", _Items)` instead.

Tests/ArrayTests.cs
```diff
@@ -0,0 +1,275 @@
...
+        public void AddCount_AddItemsAndCountThem(int size, int expected)
+        {
+            ArrayCollection<int> collection = new ArrayCollection<int>(size);
+
+            for (int i = 0; i < expected; i++)
+            {
+                collection.Add(i);
+            }
+
+            Assert.AreEqual(expected, collection.Count);
```
MarkMichaelis: Why not assert the value was added effectively, not just that the count increased?

Tests/ArrayTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            ArrayCollection<int> collection = new ArrayCollection<int>(1);
+
+            for (int i = 0; i < 2; i++)
+            {
+                collection.Add(i);
+            }
+        }
+
+        [TestMethod]
+        [DataRow(4, 2)]
```
MarkMichaelis: Nice job with the DataRow.

Tests/ArrayTests.cs
```diff
@@ -0,0 +1,275 @@
...
+            {
+                collection[i] = i;
+            }
+
+            for (int i = 0; i < 5; i++)
+            {
+                collectionExpected.Add(i);
+            }
+
+            Assert.AreEqual(collectionExpected.ToString(), collection.ToString());
```
MarkMichaelis: An interesting way to test equality.  Consider using the `CollectionAssert` class.

Tests/ArrayTests.cs
```diff
@@ -0,0 +1,275 @@
...
+        }
+
+        [TestMethod]
+        public void Collection_ForEach()
```
MarkMichaelis: Glad to see you are testing enumerator with foreach.

MarkMichaelis:
##  (NikolasGraham)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), 20 Warning(s))| **10** |


| Assignment  | **15** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |


| Assignment6 | Enable | Assignment6.sln | IntelliTectAnalyzer.dll, Microsoft.CodeAnalysis.VersionCheckAnalyzer.dll, Humanizer.dll, Microsoft.CodeQuality.Analyzers.dll, Microsoft.CodeQuality.CSharp.Analyzers.dll, Microsoft.NetCore.Analyzers.dll, Microsoft.NetCore.CSharp.Analyzers.dll, Microsoft.NetFramework.Analyzers.dll, Microsoft.NetFramework.CSharp.Analyzers.dll | CS1701, CS1702, CS2008 | 4 |
| Tests | Disable | Assignment6.sln |  | CS1701, CS1702 | 4 |


| Testing | **20** |
| Test Run Succeeded: Total tests=34, Passed=34, Failed=0, Skipped=0 | **90%** |


| Extra Credit | **0** |

PR #244, opened 2019-11-12
Assignment6

twofingerrightclick:


Assignment6/Array.cs
```diff
@@ -1,10 +1,210 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>, IEnumerable<T> where T : IEquatable<T>
```
AustinHall7798: Did you include an index operator?

twofingerrightclick: nope! merci beaucoup!

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>, IEnumerable<T> where T : IEquatable<T>
     {
+        public int Capacity { get; private set; }
```
MarkMichaelis: You don't and shouldn't change this value.  Therefore, remove the private setter.

MarkMichaelis: You don't and shouldn't change this value once set by the constructor.  Therefore, remove the private setter.

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+        {
+            if (capacity <= 0)
+            {
+                throw new IndexOutOfRangeException($"{nameof(capacity)} must be greater than 0");
+            }
+            Capacity = capacity;
+            _Array = new List<T>();
+        }
+
+        public Array()
```
MarkMichaelis: It would seem that `Capacity` is never set with the default constructor, so the size is never fixed?

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>, IEnumerable<T> where T : IEquatable<T>
     {
+        public int Capacity { get; private set; }
+
+        private List<T> _Array;
```
MarkMichaelis: Glad to see this is private.

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+            }
+        }
+
+        public T this[int index]
+        {
+
+
+            get
+            {
+                if (index > Count - 1)
```
MarkMichaelis: When, if ever, is `Capacity` used?

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+        public void Clear()
+        {
+            _Array.Clear();
+            Count = 0;
+
+        }
+
+        public bool Contains(T item)
+        {
+            if (Count > Capacity)
```
MarkMichaelis: Why is this checked here?  This method doesn't change the state of the Array, so why check count vs capacity?

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+                throw new InvalidOperationException($"{nameof(item)} was not found");
+
+            }
+
+            if (_Array.Contains(item))
+            {
+                return true;
+            }
+
+            throw new InvalidOperationException($"{nameof(item)} was not found");
```
MarkMichaelis: Why throw an exception rather than return false?  What is the point of the return if you always return true or throw an exception?

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+
+        public bool Contains(T item)
+        {
+            if (Count > Capacity)
+            {
+                throw new InvalidOperationException($"{nameof(item)} was not found");
+
+            }
+
+            if (_Array.Contains(item))
```
MarkMichaelis: Just return `_Array.Contains(item)`.

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+                }
+                _Array[index] = value;
+                Count++;
+            }
+        }
+
+
+
+
+        public bool Remove(T item)
```
MarkMichaelis: An unusual method for an array.

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+            }
+
+            return success;
+
+        }
+
+     
+
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Consider using an expression bodied member:
```
public IEnumerator<T> GetEnumerator() => new ArrayEnumerator(_Array);
```

MarkMichaelis: Consider using an expression bodied member:
```
public IEnumerator<T> GetEnumerator() => new ArrayEnumerator(_Array);
```
Better yet, just return `_Array.GetEnumerator()`.

Assignment6/Array.cs
```diff
@@ -1,10 +1,220 @@
...
+        {
+            return new ArrayEnumerator(_Array);
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        private class ArrayEnumerator : IEnumerator<T>
```
MarkMichaelis: You shouldn't need this since you can just return the enumerator from `_Array`.  

And, if you do, use the yield return statement, reducing this down to a handful of lines.

Array.Tests/UnitTest1.cs
```diff
@@ -0,0 +1,13 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
MarkMichaelis: If you aren't using it, delete the file.

Array.Tests/GlobalSuppressions.cs
```diff
@@ -0,0 +1,11 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>", Scope = "member", Target = "~M:Array.Tests.ArrayTests.Exception_When_Item_Isnt_Present_Array_Not_Empty")]
```
MarkMichaelis: If you remove the target, you only need one of these suppression rules for the project.

Array.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,316 @@
...
+
+            foreach (string s in testArray)
+            {
+                testArrayGeneric.Add(s);
+            }
+
+            int i = 0;
+            foreach ( string itemFromGenericArray in testArrayGeneric ) {
+
+                Assert.IsTrue(itemFromGenericArray == testArray[i]);
```
MarkMichaelis: Allways use `Assert.AreEqual` to check for a particular value because it shows the two values if the comparison fails.

Array.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,316 @@
...
+        {
+            int genericSize = 2;
+            
+            Array<string> testArray = new Array<string>(genericSize);
+
+            string testString = "montoya";
+
+            testArray.Add(testString);
+            Assert.IsTrue(testArray.Remove(testString));
+            Assert.IsTrue(testArray.Count == 0);
```
MarkMichaelis: Use Asssert.AreEqual instead.

Array.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,316 @@
...
+        [TestMethod]
+        [ExpectedException(typeof(IndexOutOfRangeException))]
+        public void Index_Operator_Validates_Index_Exception_When_Item_At_Index_Has_Not_Been_Initialized()
+        {
+            string[] basicArray = new string[] { "one", "two", "three" };
+
+            Array<string> testArray = new Array<string>(3);
+
+
+            for (int i = 0; i < basicArray.Length; i++)
```
MarkMichaelis: Surely you don't need the for loop for this to throw an exception?

Array.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,316 @@
...
+
+        [TestMethod]
+
+        public void Index_Operator_Sets_Correctly()
+        {
+            string[] basicArray = new string[] { "one", "two", "three" };
+
+            Array<string> testArray = new Array<string>() { "1", "2", "3" };
+    
+
```
MarkMichaelis: Couldn't you add more items to the array here - such that it isn't actually a fixed size?

MarkMichaelis:
##  (twofingerrightclick)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), 14 Warning(s))| **10** |


| Assignment  | **10** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **✔** |


| Array.Tests | Enable | Assignment6.sln |  | CS1701, CS1702 | 4 |
| Assignment6| Enable | Assignment6.sln |  | CS1701, CS1702 | 4 |

| Testing | **20** |
| Test Run Succeeded: Total tests=18, Passed=18, Failed=0, Skipped=0 | **87%** |


| Extra Credit | **0** |

PR #245, opened 2019-11-12
Assignment6

skozubenko1:


Assignment6/Array.cs
```diff
@@ -1,10 +1,166 @@
...
+            if(Count < _Array.Length)
+            {
+                _Array[Count] = item;
+                Count++;
+            }
+        }
+
+        public void Clear()
+        {
+            Count = 0;
```
CDobbins1: Clear should remove the elements from the array as well as reset the count, to ensure an old value cannot be retrieved. Storing the length parameter in the constructor to a Capacity property should allow you to reinitialize the internal array using the proper length.

Assignment6/Array.cs
```diff
@@ -1,10 +1,166 @@
...
+        }
+
+        private T GetValue(int index)
+        {
+            return _Array[index];
+        }
         
+        private void SetValue(int index, T value)
+        {
+            _Array[index] = value;
```
CDobbins1: If adding a value to the array using the index operator, count will not be incremented, resulting in an inaccurate count of elements in the array. Having this method increment Count would solve this, but would create another issue with Count if overwriting an existing value rather than adding a new value.

Assignment6.Tests/TestArray.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            array.Remove(test);
+
+            Assert.IsFalse(array.Contains(1));
+
+            array.Add(test);
+            array.Add(2);
+            array.Clear();
+
+            Assert.AreEqual(0, array.Count);
```
CDobbins1: I recommend splitting this test into multiple tests.

skozubenko1: Thank you for all the corrections. I will do my best to get tot the corrections before noon!

Assignment6/Array.cs
skozubenko1: Yeah, needs to be a judgement call on when it is appropriate to use it with the way I implemented the methods.  That's why I just use the Add method to add elements to my array in my test cases. And would only use set value to change an existing value  with an index I know exists.

Assignment6/Array.cs
skozubenko1: Count is filled array length. array.length is with the null as well. But okay, I will make a loop through count and set each index to null.

Assignment6.Tests/TestArray.cs
skozubenko1: I did not want to have a lot of duplicate code so I refactored it into one test.

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+
+                Count--;
+            }
+
+            return found;
+        }
+
+        class MyIterable : IEnumerator<T>
+        {
+            Array<T> array;
```
MarkMichaelis: Fields should be specified in _PascalCase, or switched to use an auto property

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+            int index = -1;
```
MarkMichaelis: Fields should be specified in _PascalCase, or switched to use an auto property

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T> 
     {
+        private T[] _Array { get; }
```
MarkMichaelis: Properties should be PascalCase

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+        {
+            if (length < 0)
+                throw new IndexOutOfRangeException();
+
+            _Array = new T[length];           
+        }
+        public void Add(T item)
+        {
+            if (item == null)
+                throw new ArgumentNullException();
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+        }
+        public bool IsReadOnly => false;
+
+        public T this[int key]
+        {
+            get => GetValue(key);
+            set => SetValue(key, value);
+        }
+
+        private T GetValue(int index)
```
MarkMichaelis: Why a separate method - if this is one line, why not just put it into the index operator implementation?

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+            _Array[index] = value;
+        }
+        public Array(int length)
+        {
+            if (length < 0)
+                throw new IndexOutOfRangeException();
+
+            _Array = new T[length];           
+        }
+        public void Add(T item)
```
MarkMichaelis: This method is unusual on an array.  Consider implementing it explicitly with `ICollection.Add`

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+            if(Count < _Array.Length)
+            {
+                _Array[Count] = item;
+                Count++;
+            }
+        }
+
+        public void Clear()
+        {
+            for(int i = 0; i < Count; i++)
```
MarkMichaelis: Why not just recreate the array (and add a private setter to the _Array property). :)

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+            {
+                _Array[i] = default;
+            }
+            Count = 0;
+        }
+
+        public bool Contains(T item)
```
MarkMichaelis: Well done on this implementation.

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+
+        public void CopyTo(T[] array, int arrayIndex)
+        {
+            if (arrayIndex < 0 || arrayIndex > Count - 1)
+                throw new IndexOutOfRangeException(nameof(Count));
+
+            Array.Copy(_Array, 0, array, arrayIndex, Count);
+        }
+      
+        public bool Remove(T item)
```
MarkMichaelis: It is unusual to support this on an array type.

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+
+                return false;
+            }
+
+            public void Reset()
+            {
+                index = -1;
+            }
+        }
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Consider implementing as an expression bodied member:
```
public IEnumerator<T> GetEnumerator() => _Array.GetEnumerator();
```

Assignment6/Array.cs
```diff
@@ -1,10 +1,170 @@
...
+                    _Array[i - 1] = _Array[i];                    
+                }
+
+                Count--;
+            }
+
+            return found;
+        }
+
+        class MyIterable : IEnumerator<T>
```
MarkMichaelis: You very (if ever?) need to implement an enumerator.  Instead use the yield return operator.

Assignment6.Tests/TestArray.cs
```diff
@@ -0,0 +1,70 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class TestArray
+    {
+        [TestMethod]
+        public void TestIterable()
```
MarkMichaelis: Just test with foreach - after all, that is what the task in the assignment called for. :)

Assignment6.Tests/TestArray.cs
```diff
@@ -0,0 +1,70 @@
...
+
+            while(iterator.MoveNext())
+            {
+                Console.WriteLine(iterator.Current);
+                str2 += iterator.Current.ToString();
+            }
+
+            foreach(var o in array)
+            {
+                Console.WriteLine(o);
```
MarkMichaelis: What is the point of a WriteLine() call (except perhaps debugging).  It doesn't test anything, however, because it is never checked.

Assignment6.Tests/TestArray.cs
```diff
@@ -0,0 +1,70 @@
...
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void TestAddNull()
+        {
+            Array<Object> array = new Array<Object>(500);
+
+            array.Add(null);
+        }
+
+        [TestMethod]
+        public void TestRemoveContainsClear()
```
MarkMichaelis: This is testing two different things.  I suggest splitting into two different tests.

MarkMichaelis:
##  (skozubenko1)

## Assignment6

| Assignment  | **12** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |


| Assignment6 | Disable | Assignment6.sln |  | CS1701, CS1702, CS2008 | 4 |


| Testing | **13** |
| Test Run Succeeded: Total tests=3, Passed=3, Failed=0, Skipped=0 | **75%** |


| Extra Credit | **0** |

PR #246, opened 2019-11-12
Complete bas tasts for assignment 6

markdietzler:
Interim pull request, still working on a few things...

### DIFF
--- a/Assignment6.Tests/ArrayTesting.cs
+++ b/Assignment6.Tests/ArrayTesting.cs
@@ -1,0 +1,120 @@
+using System;
+using System.Linq;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTesting
+    {
+        [TestMethod]        
+        [ExpectedException(typeof(IndexOutOfRangeException))]
+        public void Array_Add_At_OutOfBounds_Index()
+        {
+            var _TestArray1 = new Array<int>();
+            _TestArray1.Add(1);
+
+            _TestArray1[4] = 4;
+        }        
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentOutOfRangeException))]
+        public void Array_Of_Negative_Size_Throws_Exception()
+        {
+            var _TestArray3 = new Array<int>(-1);
+        }
+
+        [DataTestMethod]
+        [DataRow(1)]
+        [DataRow(2)]
+        [DataRow(3)]
+        [ExpectedException(typeof(ArgumentOutOfRangeException))]
+        public void Array_Add_To_Full_Array_Throws_Exception(int addMe)
+        {
+            var _TestArray = new Array<int>(3)
+            {
+                1,2,3
+            };
+
+            _TestArray.Add(4);            
+        }
+
+        [DataTestMethod]
+        [DataRow(1)]
+        [DataRow(2)]
+        [DataRow(3)]
+        public void Array_Add_Success(int addMe)
+        {
+            var _TestArray = new Array<int>
+            {
+                1,2,3
+            };
+
+            _TestArray.Add(4);
+
+            Assert.AreEqual(_TestArray.Count, 4);
+            Assert.AreEqual(_TestArray[3], 4);
+        }
+
+        [TestMethod]
+        public void Array_Clear_Array_Success()
+        {
+            var _TestArray = new Array<int>
+            {
+                1,2,3
+            };
+
+            _TestArray.Clear();
+
+            Assert.AreEqual(0, _TestArray.Count);
+        }
+
+        [DataTestMethod]
+        [DataRow(1)]
+        [DataRow(2)]
+        [DataRow(3)]
+        public void Array_Contains_Success(int item)
+        {
+            var _TestArray = new Array<int>
+            {
+                1,2,3
+            };
+
+            bool doesArrayContinThese = _TestArray.Contains(item);
+
+            Assert.IsTrue(doesArrayContinThese);
+        }
+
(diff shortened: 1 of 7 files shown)


Assignment6/Array.cs
```diff
@@ -1,10 +1,71 @@
...
+            _Collection.Clear();            
+        }
+
+        public bool Contains(T checkMe) => _Collection.Contains(checkMe);
+
+        public void CopyTo(T[] array, int arrayIndex) => _Collection.CopyTo(array, arrayIndex);
+
+        public IEnumerator<T> GetEnumerator() => _Collection.GetEnumerator();
+
+        public bool Remove(T item) => _Collection.Remove(item);
```
AaronSauther: I could be mistaken, but I believe Remove should throw an exception if an item does not exist

Assignment6/Array.cs
```diff
@@ -1,10 +1,71 @@
...
+        {
+            //          if int is < 0           false, do this                                   true, do this
+            _Collection = newArraySize < 0 ? new List<T>(newArraySize) : throw new ArgumentOutOfRangeException($"{nameof(newArraySize)} cannot be less than zero.");
+        }
+
+        public T this[int givenIndex]
+        {
+            get
+            {
+                ValidateIndexIsWithinArrayRange(givenIndex); //throws out of bounds exception
```
AaronSauther: Like the use of a helper method here!

GenericArrayTests/GenericArrayTests.cs
```diff
@@ -0,0 +1,100 @@
...
+    class GenericArrayTests
+    {
+        [TestClass]
+        public class ArrayTests
+        {
+            [DataTestMethod]
+            [DataRow(1)]
+            [DataRow(2)]
+            [DataRow(3)]
+            public void Add_Success(int item)
```
AaronSauther: According to [Microsoft's unit testing best practices](https://docs.microsoft.com/en-us/dotnet/core/testing/unit-testing-best-practices), you might want the name of your test(s) to consist of three parts:

The name of the method being tested.
The scenario under which it's being tested.
The expected behavior when the scenario is invoked.
```suggestion
            public void Add_OneItem_Success(int item)
```

GenericArrayTests/GenericArrayTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                1,
+                2,
+                3
+            };
+
+                sut.Clear();
+
+                Assert.AreEqual(0, sut.Count);
+
```
AaronSauther: You may want to consider adding the arrange, act, and assert comments for readability

GenericArrayTests/GenericArrayTests.cs
```diff
@@ -0,0 +1,100 @@
...
+                sut.Add(item);
+
+                Assert.AreEqual(sut.Count, 4);
+                Assert.AreEqual(sut[3], item);
+            }
+
+            [TestMethod]
+            public void Clear_Success()
+            {
+                var sut = new Array<int>
```
AaronSauther:
```suggestion
                //Arrange
                var sut = new Array<int>
```

GenericArrayTests/GenericArrayTests.cs
```diff
@@ -0,0 +1,100 @@
...
+            {
+                1,
+                2,
+                3
+            };
+
+                sut.Clear();
```
AaronSauther:
```suggestion
                //Act
                sut.Clear();
```

GenericArrayTests/GenericArrayTests.cs
```diff
@@ -0,0 +1,100 @@
...
+
+                Assert.AreEqual(0, sut.Count);
```
AaronSauther:
```suggestion
                //Assert
                Assert.AreEqual(0, sut.Count);
```

Assignment6.Tests/ArrayTesting.cs
```diff
@@ -0,0 +1,120 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTesting
+    {
+        [TestMethod]        
+        [ExpectedException(typeof(IndexOutOfRangeException))]
+        public void Array_Add_At_OutOfBounds_Index()
```
AaronSauther:
```suggestion
        public void Add_OutOfBounds_ThrowsException()
```

Assignment6.Tests/ArrayTesting.cs
```diff
@@ -0,0 +1,120 @@
...
+        public void Array_Of_Negative_Size_Throws_Exception()
+        {
+            var _TestArray3 = new Array<int>(-1);
+        }
+
+        [DataTestMethod]
+        [DataRow(1)]
+        [DataRow(2)]
+        [DataRow(3)]
+        [ExpectedException(typeof(ArgumentOutOfRangeException))]
```
AaronSauther: Like your use of data test methods, clean code!

Assignment6/Array.cs
markdietzler: LinkedList Remove() in Java returns a boolean, and returns false if it can't find the item to remove. I don't know how Icollection does it, I haven't seen the source, but judging by the method signature Icollecion _should_ throw a false as well, which gets passed up the line. Since I'm using the array class basically as a wrapper for ICollection, I don't think altering the not found behavior for Remove is warranted.

GenericArrayTests/GenericArrayTests.cs
markdietzler: Best practices also say to use one assert per test. I will split them up next time.

More typing, though.... :D

PR #249, opened 2019-11-12
Assignment6

dbosco051394:
Finished Assignment 6. Did not attempt extra credit.


kenny2892:
Your Add method doesn't check for null being passed or if the Collection is at Capacity.

Assignment6/Array.cs
```diff
@@ -0,0 +1,164 @@
...
+            }
+        }
+
+        public bool IsReadOnly { get;}
+
+        public T this[int i]
+        {
+            get => Collection[i];
+            set => Collection.Add(value);
+        }
```
NikolasGraham: Your index operator doesn't look like it functions the same as for an array. What doesn't make sense to me is say you have a collection and you want to 'change' collection[2] of a collection of size 5 that is full. Well to do that for an int collection you use collection[2] = 5. Well in your case collection[5] (which doesn't exist), is going to get that value, and collection[2] will still be as it was before.

Assignment6/Array.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public void Clear()
+        {
+            Collection.Clear();
+        }
+
+        public bool Contains(T item)
+        {
+            return Collection.Contains(item);
+        }
```
NikolasGraham: Doesn't throw an exception if something doesn't exist.

Assignment6/Array.cs
```diff
@@ -0,0 +1,164 @@
...
+
+        public bool Remove(T item)
+        {
+            return Collection.Remove(item);
+        }
```
NikolasGraham: Doesn't throw an exception if something doesn't exist (I believe he wanted this on remove also?).

Assignment6/Array.cs
```diff
@@ -0,0 +1,164 @@
...
+
+            public void Reset()
+            {
+                _Position = -1;
+            }
+
+            public void Dispose()
+            {
+
+            }
```
NikolasGraham: What is dispose going to be used for in the context of a collection?

Assignment6/Array.cs
```diff
@@ -0,0 +1,164 @@
...
+        {
+            if (array is null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+            foreach (T obj in Collection)
+            {
+                array.SetValue(obj, arrayIndex);
+                arrayIndex++;
+            }
```
NikolasGraham: Instead of this foreach, would it not be easier to use the CopyTo provided by List itself? Also what happens in the case that your collection contains more elements than the array has space from the provided index? It doesn't appear that you handle this scenario, so correct me if I'm wrong.

MarkMichaelis:
##  (dbosco051394)

## Assignment6

### It was frustrating you didn't fork the `Assignment6` branch, forcing me to have special handling for your PR.  Please always fork the assignment branch.

| Build/Compile (Build Succeeded: 0 Error(s), 8 Warning(s))| **10** |


| Assignment  | **10** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |


| ArrayTests | Disable | Assignment6.sln |  | CS1701, CS1702 | 4 |
| Assignment6| Enable | Assignment6.sln |  | CS1701, CS1702 | 4 |

| Testing | **20** |
| Test Run Succeeded: Total tests=13, Passed=13, Failed=0, Skipped=0 | **90%** |


| Extra Credit | **0** |

PR #251, opened 2019-11-12
Assignment 6

whereisdaniru:


Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>
     {
-
```
MarkMichaelis: Change 'Items' to be read-only by removing the property setter.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                Items.Clear();
+            }
+        }
+
+        public bool Contains(T item)
+        {
+            if (Items is null)
+            {
+                throw new ArgumentNullException(nameof(Items));
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                Items.Add(item);
+            }
+        }
+
+        public void Clear()
+        {
+            if (Items is null)
+            {
+                throw new ArgumentNullException(nameof(Items));
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+
+        public bool Contains(T item)
+        {
+            if (Items is null)
+            {
+                throw new ArgumentNullException(nameof(Items));
+            }
+            if (item is null)
+            {
+                throw new ArgumentNullException(nameof(Items));
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                Items.CopyTo(array, arrayIndex);
+            }
+        }
+
+        public bool Remove(T item)
+        {
+            if (Items is null)
+            {
+                throw new ArgumentNullException(nameof(Items));
```
MarkMichaelis: The parameter passed to this exception's constructor should be the parameter name; not a message

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T> : ICollection<T>
     {
-        
+        public List<T> Items { get; set; }
```
MarkMichaelis: Yikes... public with a setter?  This seems problematic.  I would think this should be private with no setter.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+        public int Capacity { get; set; }
```
MarkMichaelis: Again... a public setter?  This seems to break encapsulation pretty significantly.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(items));
+            }
+
+            Items = new List<T>(items);
+            Capacity = Items.Count;
+            
+        }
+
+        public void Add(T item)
```
MarkMichaelis: This is an unusual method to have on an array.  Consider implementing explicitly instead.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            }
+            else
+            {
+                Items.Add(item);
+            }
+        }
+
+        public void Clear()
+        {
+            if (Items is null)
```
MarkMichaelis: The fact that items is public forces you to make this check but it shouldn't be necessary.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                throw new IndexOutOfRangeException(nameof(arrayIndex));
+            }
+            else
+            {
+                Items.CopyTo(array, arrayIndex);
+            }
+        }
+
+        public bool Remove(T item)
```
MarkMichaelis: It is unusual to have this method on an array.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+                return Items.GetEnumerator();
+            }
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        public int IndexOf(T item)
```
MarkMichaelis: This is not an index operator - although the functionality is the same.

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+            {
+                throw new ArgumentNullException(nameof(item));
+            }
+            else
+            {
+                return Items.Remove(item);
+            }
+        }
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Consider implementing with an expression bodied member.
```
public IEnumerator<T> GetEnumerator() => Items.GetEnumerator();
```
Or, if you have to check for null (which you shouldn't):
```
public IEnumerator<T> GetEnumerator() => Items?.GetEnumerator()??throw new InvalidOperationException(...);
```

Assignment6/Array.cs
```diff
@@ -1,10 +1,155 @@
...
+        {
+            if (Items is null)
+            {
+                throw new NullReferenceException(nameof(Items));
```
MarkMichaelis: You should never throw a `NullReferenceException`.  In this case you would have to use an `InvalidOperationException`.

Assignment6/Array.cs
MarkMichaelis: This does not suffice as an index operator (although the functionality is similar) - it isn't an operator implementation.

MarkMichaelis: I'm not sure if you intended it to me, but this does not suffice as an index operator (although the functionality is similar) - it isn't an operator implementation.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,170 @@
+using System;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTests
+    {
```
MarkMichaelis: You don't test the foreach task that was part of the assignment.

MarkMichaelis:
##  (whereisdaniru)

## Assignment6

| Assignment  | **10** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |


| Testing | **10** |
| Test Run Succeeded: Total tests=11, Passed=11, Failed=0, Skipped=0 | **63%** |
| Most assigned work is tested or contain some errors. | **0** |


| Extra Credit | **0** |

PR #252, opened 2019-11-13
Done

matHopg:


Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Cannot_Add_Null_Type()
+        {
+            Array<string> testArray = new Array<string>(1);
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
+            string? t = null;
+#pragma warning restore CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```
crackerJackTech: Pragma warning can be removed if you use the null forgiveness operator.
string t = null!

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
     {
+        public int Size { get; private set; }
+        private List<T> _Collection;
+        public int Count { get; private set; } = 0;
+        public bool IsReadOnly => throw new NotImplementedException();
+        public Array(int size)
+        {
+            if(size <= 0)
+            {
+                throw new IndexOutOfRangeException($"{nameof(size)} cannot be zero or less");
```
crackerJackTech: Great idea for using string interpolation in the message argument. I have never thought of doing that!

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            if (success)
+            {
+                Count--;
+            }
+            return success;
+        }
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            return new ArrayEnumerator(_Collection);
```
crackerJackTech: For the return value, I think you can use 'return _Collection.GetEnumerator' instead of using the ArrayEnumerator class. I may be incorrect.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+        }
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return GetEnumerator();
+        }
+
+        private class ArrayEnumerator : IEnumerator<T>
```
crackerJackTech: This ArrayEnumerator class can be relocated to its own file without having to be contained in the same class as the Array<T> class.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T>: ICollection<T>, IEnumerable<T> where T: IEquatable<T>
     {
+        public int Size { get; private set; }
+        private List<T> _Collection;
+        public int Count { get; private set; } = 0;
+        public bool IsReadOnly => throw new NotImplementedException();
```
MarkMichaelis: get_IsReadOnly creates an exception of type NotImplementedException, an exception type that should not be raised in a property. If this exception instance might be raised, use a different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            }
+        }
+
+        public T this[int index]
+        {
+            get
+            {
+                if(index> Count - 1)
+                {
+                    throw new IndexOutOfRangeException($"Value at {nameof(index)} is not initialized");
```
MarkMichaelis: get_Item creates an exception of type IndexOutOfRangeException, an exception type that should not be raised in a property. If this exception instance might be raised, use a different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+                }
+                if (index < 0)
+                {
+                    throw new IndexOutOfRangeException($"Index cannot be zero or lower");
```
MarkMichaelis: get_Item creates an exception of type IndexOutOfRangeException, an exception type that should not be raised in a property. If this exception instance might be raised, use a different exception type, convert this property into a method, or change this property's logic so that it no longer raises an exception.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+                    {
+                        return _Collection[_Index];
+                    }
+                    catch(IndexOutOfRangeException)
+                    {
+                        throw new InvalidOperationException();
+                    }
+                }
+            }
+            Object IEnumerator.Current
```
MarkMichaelis: Use keyword instead of type name

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
 ﻿using System;
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T>: ICollection<T>, IEnumerable<T> where T: IEquatable<T>
```
MarkMichaelis: Wow... nice job with specifying the constraint on `T`.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+using System.Collections;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Assignment6
 {
-    public class Array<T>
+    public class Array<T>: ICollection<T>, IEnumerable<T> where T: IEquatable<T>
     {
+        public int Size { get; private set; }
```
MarkMichaelis: Why do you have a setter on size?  It shouldn't change after the constructor I would think (and you never assign it outside of the constructor it would seem).

Assignment6/Array.cs
MarkMichaelis: It would seem that `false` would be a better return (see https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icollection-1.isreadonly?view=netcore-3.0)

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+        {
+            if(size <= 0)
+            {
+                throw new IndexOutOfRangeException($"{nameof(size)} cannot be zero or less");
+            }
+            Size = size;
+            _Collection = new List<T>();
+        }
+
+        public Array()
```
MarkMichaelis: Won't this result in a size of zero, which never changes?  It seems like no default constructor should be allowed.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+        {
+            _Collection = new List<T>();
+        }
         
+        public void Add(T item)
```
MarkMichaelis: Since Add is not generally an array type behavior, preferably implement this explicitly.

MarkMichaelis: This isn't common array behavior so I would implement this member of `ICollection<T>` explicitly.

Assignment6/Array.cs
MarkMichaelis: Why a private setter?  Shouldn't this be read-only?

Also, I would use the name `Capacity`, which is consistent with the assignment and with other collections like `List<T>`.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+
+        public void Clear()
+        {
+            _Collection.Clear();
+            Count = 0;
+        }
+
+        public bool Contains(T item)
+        {
+            if(Count > Size)
```
MarkMichaelis: How could this scenario occur and why are you checking for it in Contains which is a read-only (doesn't change the state of the class) method?

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            {
+                throw new InvalidOperationException($"{nameof(item)} not found");
+
+            }
+            if (_Collection.Contains(item))
```
MarkMichaelis: Why not just return `_Collection.Contains(item)`?

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            {
+                return true;
+            }
+            throw new InvalidOperationException($"{nameof(item)} not found");
```
MarkMichaelis: What is the point of a return if you always return true or throw an exception?

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            {
+                array[i] = _Collection[i];
+            }
+        }
+
+        public T this[int index]
+        {
+            get
+            {
+                if(index> Count - 1)
```
MarkMichaelis: How are your checks different from those provided by `_Collection[index]`?

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+                if(index < 0)
+                {
+                    throw new IndexOutOfRangeException($"Index cannot be zero or lower");
+                }
+                _Collection[index] = value;
+                Count++;
+            }
+        }
+
+        public bool Remove(T item)
```
MarkMichaelis: This is unusual behavior for an array.

Assignment6/Array.cs
```diff
@@ -1,10 +1,185 @@
...
+            bool success = false;
+            success = _Collection.Remove(item);
+            if (success)
+            {
+                Count--;
+            }
+            return success;
+        }
+
+        public IEnumerator<T> GetEnumerator()
```
MarkMichaelis: Consider using an expression bodied member:
```
        public IEnumerator<T> GetEnumerator() => new ArrayEnumerator(_Collection);
```
Better yet, why not just return `_Collection.GetEnumerator()`?

Assignment6/Array.cs
MarkMichaelis: It is very rare to ever have to do this.  Instead use the yield return operator to implement IEnumerable.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+            Array<int> testArray = new Array<int>(0);
+            testArray.Add(2);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Cannot_Add_Null_Type()
+        {
+            Array<string> testArray = new Array<string>(1);
+#pragma warning disable CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
```
MarkMichaelis: Don't disable the warning, rather enable nullable at the project level.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+#pragma warning restore CS8632 // The annotation for nullable reference types should only be used in code within a '#nullable' annotations context.
+            testArray.Add(t);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(InvalidOperationException))]
+        public void Exception_When_Item_Doesnt_Exist_Array_Empty()
+        {
+            Array<string> testArray = new Array<string>(1);
+            testArray.Contains("Matthew");
```
MarkMichaelis: No, Contains should return false when the item doesn't exist.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+
+        [TestMethod]
+        public void Index_Operator_Retrieves_Succesfully()
+        {
+            string[] basicArray = new string[] { "1", "2", "3" };
+            Array<string> testArray = new Array<string>() { "1", "2", "3" };
+            for(int i =0; i<basicArray.Length; i++)
+            {
+                Assert.IsTrue(testArray[i] == basicArray[i]);
```
MarkMichaelis: Use Assert.AreEqual() instead... that way you can see what the two values are when they are not equal.

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+        }
+        [TestMethod]
+        public void Remove_Returns_True_Decrements_Count()
+        {
+            int size = 2;
+            Array<string> testArray = new Array<string>(size);
+            string testString = "test";
+            testArray.Add(testString);
+            Assert.IsTrue(testArray.Remove(testString));
+            Assert.IsTrue(testArray.Count == 0);
```
MarkMichaelis: I would also test to see if the item is removed (using a non-exception throwing Contains() invocation. :)

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        [TestMethod]
+        public void Clear_Sets_Count_to_Zero()
+        {
+            int csize = 1;
+            Array<string> testArray = new Array<string>(csize);
+            string name = "test";
+            testArray.Add(name);
+            testArray.Clear();
+            Assert.IsTrue(testArray.Size == csize);
```
MarkMichaelis: Use Assert.AreEqual to check for specific values.

Assignment6/Assignment6.csproj
```diff
@@ -1,7 +1,15 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
-  </PropertyGroup>
-
-</Project>
+    <nullable>enable</nullable>
```
MarkMichaelis: Why do you have this twice?

MarkMichaelis: Why do you have this twice?

And, isn't the element `Nullable`?

Assignment6.Tests/ArrayTests.cs
```diff
@@ -0,0 +1,121 @@
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+
+namespace Assignment6.Tests
+{
+    [TestClass]
+    public class ArrayTests
+    {
```
MarkMichaelis: You don't appear to test iterating with foreach? (or even GetEnumerator() explicitly?)

MarkMichaelis:
##  (matHopg)

## Assignment6

| Build/Compile (Build Succeeded: 0 Error(s), 24 Warning(s))| **10** |


| Assignment  | **7** |


| Project Analysis |   |
| -----------------| - |
| Nullable Enabled: | **❌** |
| CodeReview Performed | **-5** |


| Testing | **15** |
| Test Run Succeeded: Total tests=10, Passed=10, Failed=0, Skipped=0 | **72%** |


| Extra Credit | **0** |

PR #254, opened 2019-11-14
Assignment7

twofingerrightclick:
I think I did the rebase right!

### DIFF
--- a/DelegateFun/Sorter.Tests/SortUtilityTests.cs
+++ b/DelegateFun/Sorter.Tests/SortUtilityTests.cs
@@ -1,4 +1,6 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Sorter;
+using System;
 
 namespace Sorter.Tests
 {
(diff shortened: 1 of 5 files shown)


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -21,7 +21,7 @@ public bool IsGreater(int first, int second)
 
         public bool IsLess(int first, int second)
         {
-            return first > second;
+            return first < second;
         }
```
ashermancinelli: Is this not implicit in the `<` operator?

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -14,41 +13,82 @@ public class SortUtilityTests
...
+        }
+
+
+        [TestMethod]
+        public void SortUtility_ShouldSortDescending_UsingLambdaStatement()
+        {
+            //setup
+            int[] array = arrayAscending;
+
+            Func< int, int, bool> sortByDescending = (x, y) => { return x > y; };
```
ashermancinelli: Just pointing out that the readme says not to use `Func`

twofingerrightclick: Thanks! I realize now I can implicitly cast to the defined delegate.

skozubenko1:
I like how you break everything down into sections. I found it helpful in understanding the assignment when I was working on it. Thank you!

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -1,10 +1,81 @@
...
+            SortBy = sortBy;
+        }
         // Sort method should be implemented here
         // It should accept an int[] and a delegate you define that performs the actual comparison
+
+        /* low  --> Starting index,  high  --> Ending index */
+
+        public void Sort(int[] arr)
+        {
+            if (arr is null)
```
breaman: Where is the unit test around this check?

PR #255, opened 2019-11-14
Assignment7

dnisyndrum:


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,57 @@
...
+            ints = SortUtility.Sort(ints, sorter);
+
+            Assert.IsTrue(ints.SequenceEqual(intsOrdered));
+        }
+
+        [TestMethod]
+        public void SortUtility_SortAlphabetically_UsingALambdaExpresion()
+        {
+            int[] ints = { 0, 2, 1, 5, 19, 34, 5, 6 };
+            int[] intsOrdered = { 0, 1, 2, 5, 5, 6, 19, 34 };
```
breaman: This result does not match what your method says it should do. This sort looks to be the same as the sort in the test below it. Also, a lambda expression should look like (first, second) => first > second;

PR #256, opened 2019-11-14
Assignment7

pnotequalnp:
WIP

### DIFF
--- a/DelegateFun/Sorter.Tests/SortUtilityTests.cs
+++ b/DelegateFun/Sorter.Tests/SortUtilityTests.cs
@@ -8,7 +8,7 @@ namespace Sorter.Tests
     public class SortUtilityTests
     {
         [TestMethod]
-        public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
+        public void Sort_ShouldSortAscending_UsingAnAnonymousMethod()
         {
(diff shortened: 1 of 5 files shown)


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -8,6 +10,40 @@ public class SortUtilityTests
         [TestMethod]
         public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
         {
+            var list = new[] { 5, 2, 7, 8, 3, 6, 1, 9, 0, 4 };
+            var correct = new[] { 0, 1, 2, 3, 4, 5, 6, 7, 8, 9 };
+
+            int[] sorted = SortUtility.Sort(list, delegate(int i, int j)
```
twofingerrightclick: So an lambda expression is the same in this case as an anonymous method?

pnotequalnp: No the anonymous method syntax uses the delegate keyword and has a multiline body with braces, the lambda expression syntax does not use the delegate keyword and has a single line body with no braces.

twofingerrightclick: Thanks for clarifying for me!

PR #258, opened 2019-11-15
Assignment7_Joel

RSXEnthusiast:
Completed

### DIFF
--- a/DelegateFun/Sorter.Tests/SortUtilityTests.cs
+++ b/DelegateFun/Sorter.Tests/SortUtilityTests.cs
@@ -1,13 +1,62 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using static Sorter.SortUtility;
 
-namespace Sorter.Tests
-{
+namespace Sorter.Tests {
     [TestClass]
-    public class SortUtilityTests
-    {
+    public class SortUtilityTests {
         [TestMethod]
-        public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
-        {
+        public void SelectionSort_ShouldSortAscending_UsingAnAnonymousMethod() {
+            //Arrange
+            int[] testArray = CreateRandomArray(200);
+
+            //Act
+            SelectionSort(testArray, delegate (int first, int second) { return first < second; });
+
+            //Assert
+            Assert.IsTrue(IsSorted(testArray));
+        }
+
+        [TestMethod]
+        public void SelectionSort_ShouldSortAscending_UsingLambdaExpression() {
+            //Arrange
+            int[] testArray = CreateRandomArray(200);
+
+            //Act
+            SelectionSort(testArray, (int first, int second) => first < second);
+
+            //Assert
+            Assert.IsTrue(IsSorted(testArray));
+        }
+
+        [TestMethod]
+        public void SelectionSort_ShouldSortAscending_UsingLambdaStatement() {
+            //Arrange
+            int[] testArray = CreateRandomArray(200);
+
+            //Act
+            SelectionSort(testArray, (int first, int second) => { return first < second; });
+
+            //Assert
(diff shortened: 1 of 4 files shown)


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,62 @@
...
+        public int[] CreateRandomArray(int length) {
+            Random rand = new Random();
+            int[] randomArray = new int[length];
+            for (int i = 0; i < length; i++) {
+                randomArray[i] = rand.Next();
+            }
+            return randomArray;
+        }
+
+        public bool IsSorted(int[] a) {
```
AustinHall7798: Stokes said in class that we should sort the array in a different way for each test

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -1,10 +1,20 @@
...
-namespace Sorter
-{
-    public class SortUtility
-    {
-        // Sort method should be implemented here
-        // It should accept an int[] and a delegate you define that performs the actual comparison
+namespace Sorter {
+    public class SortUtility {
+        public delegate bool Comparer(int first, int second);
+        public static void SelectionSort(int[] array, Comparer comparer) {
```
AustinHall7798: You might want to check and unit test for array being null

DelegateFun/Sorter/Sorter.csproj
```diff
@@ -1,7 +1,10 @@
-<Project Sdk="Microsoft.NET.Sdk">
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
AustinHall7798: Nullability should be enabled

DelegateFun/Sorter.Tests/SortUtilityTests.cs
RSXEnthusiast: Yea, just realized and fixed that. Thanks!

DelegateFun/Sorter/Sorter.csproj
RSXEnthusiast: It was enabled, but I was getting a weird error. Now I'm not. Whatever, it's working now. Thanks!

DelegateFun/Sorter/SortUtility.cs
RSXEnthusiast: Added a test for null reference.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,94 @@
...
+
+            //Act
+            SelectionSort(testArray, delegate (int first, int second) { return first < second; });
+
+            //Assert
+            Assert.IsTrue(IsSortedNumberValueInOrder(testArray));
+        }
+
+        [TestMethod]
+        public void SelectionSort_ShouldSortAscending_UsingLambdaExpression() {
```
breaman: These test method names don't match what is actually being tested.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,94 @@
...
+            Console.WriteLine(string.Join(",", testArray));
+            //Assert
+            Assert.IsTrue(IsSortedNumberLengthInOrderReverseThenNumberValueInOrder(testArray));
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(NullReferenceException))]
+        public void SelectionSort_ThrowsArgumentNullExceptionForNullArray() {
+            //Act
+            SelectionSort(null,(int first,int second) => (first>second));
```
breaman: Instead of testing for a null reference exception, your code should actually have checked for a null and then done an ArgumentNullException so that things would be more clear to the user since many things can throw a null ref exception.

PR #260, opened 2019-11-15
Assignment 7

AaronSauther:
I have addressed all warnings and messages except for a strange "AD0001" warning that the IntelliTect analyzer is throwing. It claims that the source of the error is coming form a "CSC" file under Sorter.tests. When I click on in the warning nothing happens/it does not bring me to any code that is responsible for generating this error. It does not give any kind of good explanation for the error as well so I have no idea what the analyzer is on about. Since this seems to be a bug with the analyzer I have tried to suppress it but have been unable to. Any suggestions here would be appreciated.


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,92 @@
...
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Sort_NullArray_ThrowsException()
+        {
+            //Arrange
+            int[]? nullArray = null;
+            //Act
+#pragma warning disable CS8604 // Possible null reference argument.
+            SortUtility.Sort(nullArray, delegate (int first, int second)
+#pragma warning restore CS8604 // Passing null for testing purposes
```
Fervidusletum:
```
SortUtility.Sort(null!, delegate (int first, int second)
```
you can use the null-forgiveness operator here to tell the compiler not to worry about a null. Lets you remove the `#pragma` and the `int?`.

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,31 @@
...
+                throw new ArgumentNullException(nameof(unsortedArray));
+            if (sort is null)
+                throw new ArgumentNullException(nameof(sort));
+            //implement selection sort
+            for (int firstIndex = 0; firstIndex < unsortedArray.Length - 1; firstIndex++)
+            {
+                int targetIndex = firstIndex;
+                for (int secondIndex = firstIndex + 1; secondIndex < unsortedArray.Length; secondIndex++)
+                {
+                    if (sort(unsortedArray[secondIndex], unsortedArray[targetIndex]))
```
Fervidusletum: `sort` might be a bit misleading name for your delegate, I think something along the lines of `compare` would indicate its function more clearly.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,92 @@
...
         {
+            //Arrange
+            int[] unsortedArray = { 5, 2, 4, 1, 3 };
+            int[] sortedArray = { 1, 2, 3, 4, 5 };
+            //Act
+            SortUtility.Sort(unsortedArray, delegate (int first, int second)
+                {
+                    return first < second;
+                });
+            bool arrayIsSortedAscending = Enumerable.SequenceEqual(unsortedArray, sortedArray);
```
Fervidusletum: Good idea! I'm going to use this in my own tests.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
AaronSauther: Good point, always for get about that operator. Thanks!

DelegateFun/Sorter/SortUtility.cs
AaronSauther: I do suppose that narrows it down a little bit and is more intent revealing. Thank for the tip!

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,90 @@
...
+using System.Linq;
 
 namespace Sorter.Tests
 {
     [TestClass]
     public class SortUtilityTests
     {
         [TestMethod]
-        public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
+        public void Sort_UsingAnAnonymousMethod_ShouldSortAscending()
```
breaman: Where are the tests that use a lambda expression and lambda statement?

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,90 @@
...
+                return false;
+            });
+            bool arrayIsSortedLexicographicly = Enumerable.SequenceEqual(unsortedArray, sortedArray);
+            //Assert
+            Assert.IsTrue(arrayIsSortedLexicographicly);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void Sort_NullArray_ThrowsException()
```
breaman: nice null check tests

PR #262, opened 2019-11-16
Assignment 7

Fervidusletum:
Finished

### DIFF
--- a/DelegateFun/Sorter.Tests/SortUtilityTests.cs
+++ b/DelegateFun/Sorter.Tests/SortUtilityTests.cs
@@ -1,13 +1,74 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Sorter.Tests
 {
     [TestClass]
     public class SortUtilityTests
     {
+        private static readonly int[] _UnsortedList = new int[] { 4, 1, 10, 5, 8, 9, 3, 7, 2, 6 };
+
+        private static readonly int[] _SortedList = new int[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 10 };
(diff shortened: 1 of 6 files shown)


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -36,10 +42,7 @@ public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
                 return i > j;
             });
 
-            for (int i = 0; i < items.Length; i++)
-            {
-                Assert.AreEqual<int>(_SortedList[i], items[i]);
-            }
+            Assert.IsTrue(Enumerable.SequenceEqual<int>(items, _SortedList));
```
crackerJackTech: I was also iterating through each list and comparing the elements until I found "CollectionAssert.AreEqual" and I wonder if it would work here instead of using "Assert.IsTrue(Enumerable.SequenceEqual<int>(items, _SortedList));" you could change it to "CollectionAssert.AreEqual(_SortedList, items)"

Fervidusletum: Thanks for the tip! Makes the assert a bit cleaner.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,72 @@
...
+using System;
+using System.Linq;
 
 namespace Sorter.Tests
 {
     [TestClass]
     public class SortUtilityTests
     {
+        // contains a variety of values for testing
+        private static int[] UnsortedList { get; } = new int[] { 4, 37, 1, 10, 5, 8, 24, 9, 3, 7, -6, 2, 11, 6 };
```
breaman: I think if you would have left this as not being static, then it would have created a new instance for you for each test run, dropping the need for GetUnsortedList()

PR #266, opened 2019-11-17
Assignment7

NikolasGraham:
Finished


DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,32 @@
...
         // Sort method should be implemented here
         // It should accept an int[] and a delegate you define that performs the actual comparison
+        public void selectionSort(int[] ara, intComparer compare)
+        {
+            if(ara == null)
+            {
+                throw new ArgumentNullException(nameof(ara));
+            }
+
+            for(int i=0; i < ara.Length-1; i++)
```
GamerBah: What happens if someone passes a null intComparer?

PR #267, opened 2019-11-17
Assignment7

Cybotive:
~Why can the test class not see the SortUtility class?~

### DIFF
--- a/DelegateFun/Sorter.Tests/GlobalSuppressions.cs
+++ b/DelegateFun/Sorter.Tests/GlobalSuppressions.cs
@@ -1,0 +1,6 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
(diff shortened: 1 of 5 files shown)


markdietzler:
`if (items is null) { return; }`

Don't let Stu see that! ; also, shouldn't this be throwing a ArgumentNullException, or did you decide to not do that?

Cybotive:
> 
> 
> `if (items is null) { return; }`
> 
> Don't let Stu see that! ; also, shouldn't this be throwing a ArgumentNullException, or did you decide to not do that?

This is correct for my implementation. A null array and an empty array are inherently sorted.

markdietzler:
> 
> 
> > `if (items is null) { return; }`
> > Don't let Stu see that! ; also, shouldn't this be throwing a ArgumentNullException, or did you decide to not do that?
> 
> This is correct for my implementation. A null array and an empty array are inherently sorted.
> 
> Never actually met the guy.

Null array is indeed a sorted one. Whether a null array is allowed or not is a software design issue, I suppose. I chose to throw an exception for it.

PR #269, opened 2019-11-17
Assignment 7

CDobbins1:


DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,27 @@
...
+                int j = i - 1;
+
+                while (j >= 0 && compare(key, array[j]))
+                {
+                    array[j + 1] = array[j];
+                    j = j - 1;
+                }
+                array[j + 1] = key;
+            }
+        }
```
pnotequalnp: This fails if passed `null` for the `compare` parameter.

CDobbins1: Thanks for pointing this out, I should have this fixed in the next commit.

PR #270, opened 2019-11-18
Assignment 7

JerettLatimer:


dbosco051394:
I just looked through your tests and source code. Looks great! Good naming conventions and testing seems to cover everything.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -8,6 +10,81 @@ public class SortUtilityTests
...
+            CollectionAssert.AreEqual(test, expected);
+
+        }
+
+        [TestMethod]
+        public void SortUtility_SortDescendingAlphabetical_UsingLambdaStatement()
+        {
+            // Arrange
+            int[] test = { 1, 4, 2, 3 };
+            int[] expected = { 4, 3, 2, 1 };
```
breaman: This test may have been better if you put an 11 or something like that in there since this would technically still pass even if you were only doing a numeric descending sort. There is no data here that proves it is really doing an alphabetical sort.

PR #271, opened 2019-11-18
Finished code. Ready for review.

dbosco051394:
Finished writing the SortUtilityTests and SortUtility. If I made any mistakes or misunderstood a concept, please let me know. Always looking to improve.

### DIFF
--- a/DelegateFun/Sorter.Tests/SortUtilityTests.cs
+++ b/DelegateFun/Sorter.Tests/SortUtilityTests.cs
@@ -1,4 +1,5 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
 
 namespace Sorter.Tests
 {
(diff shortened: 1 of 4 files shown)


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -8,6 +9,62 @@ public class SortUtilityTests
         [TestMethod]
         public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
         {
+            int[] arr = { 3, 4, 1, 8, 10, 2, 100, 33 };
+            int[] expectedResults = { 1, 2, 3, 4, 8, 10, 33, 100 };
+            SortUtility.ComparisonDel del = delegate(int num1, int num2)
```
Cybotive: You could get rid of the assignment to del and throw everything after the equals, in place of "del" on line 18. I believe "int" before num1 and num2 is also unnecessary.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -8,6 +9,62 @@ public class SortUtilityTests
...
+            }
+        }
+
+        [TestMethod]
+        public void SortUtility_ShouldSortDescending_UsingLambdaExpression()
+        {
+            int[] arr = { 3, 4, 1, 8, 10, 2, 100, 33 };
+            int[] expectedResults = {100, 33, 10, 8, 4, 3, 2, 1};
+            SortUtility.InsertionSort(arr, (num1, num2) =>
+            {
```
Cybotive: This is a lambda statement, not expression, solely due to using curly brackets

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -6,5 +6,25 @@ public class SortUtility
...
+        public delegate bool ComparisonDel(int num1, int num2);
+        public static void InsertionSort(int[] arr, ComparisonDel comparison)
+        {
+            int n = arr.Length;
+            for (int i = 1; i < n; i++)
+            {
+                int num2 = arr[i];
+                int j = i - 1;
+
+                while(j>=0 && comparison(arr[j], num2))
```
Cybotive: Looks good

DelegateFun/Sorter.Tests/SortUtilityTests.cs
dbosco051394: Thanks. I got mixed up on the terminology between statement and expression. I'll update my method names to reflect this.

breaman: Looks like this didn't get updated because this is still using a statement but the test says expression. You at least did get an expression in a different test so won't mark you down, but be aware of naming methods to make things clear.

PR #272, opened 2019-11-18
Assignment 7

ashermancinelli:


mmwoodfo:
You're pushing to master. I always do that when I make PR's on GitHub instead of GitKraken. :)

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -0,0 +1,119 @@
...
+        public void SortUtility_ShouldSortDescending_UsingAnAnonymousMethod(int[] arr)
+        {
+            var sut = new SortUtility();
+
+            Key key = delegate (int a, int b)
+            {
+                return a < b;
+            };
+
+            int[] expected = arr;
```
pnotequalnp: These methods aren't testing the sort at all. This line is making a separate reference to the same array, so when you manipulate `expected` below, it is also doing it to `arr` (since they point to the same object), and you are then calling QuickSort on an already sorted array.

ashermancinelli:
@mmwoodfo thanks, I branched from Assignment7 and used the github cli and assumed it would figure it out :/

ashermancinelli:
@mmwoodfo Any idea why this is still failing? This passes all tests on my machine, I'm not sure what's going on here.

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,50 @@
 
 namespace Sorter
 {
+    public delegate bool Key(int first, int second);
+
     public class SortUtility
     {
-        // Sort method should be implemented here
-        // It should accept an int[] and a delegate you define that performs the actual comparison
+        private static void QuickSort(int[] arr, int l, int r, Key key)
```
samrnixon: Very interesting way of doing the sort method, Nice!

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,50 @@
...
+        {
+            if (key is null)
+                throw new ArgumentNullException(nameof(key));
+            QuickSort(arr, 0, arr.Length-1, key);
+        }
+
+        private static int Partition(int[] arr, int l, int r, Key key)
+        {
+            int piv = arr[r],
+                i = l - 1;
```
Fervidusletum: This appears to be where you're running into trouble with `IndexOutOfRangeException`.  Initial call to quicksort sets `L` to 0, so `I` gets set to -1, causing the exception when you attempt to index at `I` at L#39

ashermancinelli: I'm not sure about that. I think at least one element will always pass the key on the first run in the for loop just below this line, where `i` is incremented before being used. At the start, `i = -1`, yet it will always be incremented to `0`.

mmwoodfo:
@ashermancinelli Sorry only saw your @ today, I'm glad you figured it out.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,117 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using Sorter;
 
 namespace Sorter.Tests
 {
     [TestClass]
     public class SortUtilityTests
     {
+        [DataTestMethod]
```
breaman: You are missing a test where the order is something besides ascending or descending. The assignment was to have 3 different method types performing 3 different types of sorts.

PR #273, opened 2019-11-18
Assignment 7

crackerJackTech:


DelegateFun/Sorter/SortUtility.cs
```diff
@@ -2,9 +2,37 @@
...
+        public delegate bool SortingTypeFuction(int i, int j);
+
+        public static void InsertionSort(int[] items, SortingTypeFuction compare)
+        {
+            if (items is null)
+            {
+                throw new ArgumentNullException(nameof(items));
+            }
+
+            if(compare is null)
```
skozubenko1: I like that you check if your delegate is null. Everything looks good! :)

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,91 @@
...
+            {
+                return first > second;
+            });
+
+            //Assert
+            CollectionAssert.AreEqual(sortedArray, unsortedArray);
+        }
+
+        [TestMethod]
+        [DataRow(new int[] {1,2,3,4,5,6,7,8,9}, new int[] {9,8,7,6,5,4,3,2,1})]
```
skozubenko1: I like the way you use DataRow to test the data. I found it helpful to understand what I will be testing.

PR #274, opened 2019-11-18
assignment 7

skozubenko1:


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,157 @@
...
+using System;
 
 namespace Sorter.Tests
 {
     [TestClass]
     public class SortUtilityTests
     {
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void passNullTest()
```
breaman: Methods should be PascalCase

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,157 @@
...
+
+
+        }
+
         [TestMethod]
         public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod()
         {
+            SortUtility arr = new SortUtility();
+
+            ComparatorDel ascending = (left, right) => { return left < right; };
```
breaman: this is not an anonymous method, this is a lambda statement

PR #275, opened 2019-11-19
Finalized solution and tests

markdietzler:


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,92 @@
...
+            //Act
+
+            rawArray = util.SelectionSortUtil(rawArray, ((int thisOne, int thatOne) => thisOne > thatOne));
+
+            //Assert
+            CollectionAssert.AreEqual(rawArray, sorted);
+        }
+
+        [TestMethod]
+        public void SortUtilSortAscendingLabdaStatement()
```
breaman: Watch spelling on method names, this should have been Lambda instead of Labda I assume. This third method should have sorted differently from the other two methods you had earlier.

PR #277, opened 2019-11-19
Assignment 7

GamerBah:


DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,82 @@
...
+        [DataRow(new[] {10, 1238, 23, 29, 0, 13, 2}, new[] {0, 2, 10, 13, 23, 29, 1238})]
+        [DataRow(new[] {0, 0, 0, 1, 2, 1, 0, 0, 0}, new[] {0, 0, 0, 0, 0, 0, 1, 1, 2})]
+        public void SortUtility_ShouldSortAscending_UsingAnAnonymousMethod(int[] array, int[] expected)
+        {
+            var sortUtility = new SortUtility();
+
+            Comparison comparison = delegate(int x, int y)
+            {
+                if (x < y) return true;
+                return false;
```
kenny2892: This could just be "return x < y;" but other than that, it all looks good

PR #278, opened 2019-11-19
Assignment7-1 completed

Tpirtle:


DelegateFun/Sorter/SortUtility.cs
```diff
@@ -1,10 +1,35 @@
...
     {
         // Sort method should be implemented here
         // It should accept an int[] and a delegate you define that performs the actual comparison
+        public delegate bool IsLowerOrHigher(int x, int y);
+
+        public int[] SelectionSort(int[] ara, IsLowerOrHigher isLowerOrHigher)
+        {
+            if (ara == null)
+                throw new ArgumentNullException("The Integer Array is Null.");
+            else if (isLowerOrHigher == null)
```
mmwoodfo:
```suggestion
            if (isLowerOrHigher == null)
```
Good call - I forgot to do this on mine. However you don't need and if-else here, you can just use two if's.

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -1,10 +1,35 @@
...
     public class SortUtility
     {
         // Sort method should be implemented here
         // It should accept an int[] and a delegate you define that performs the actual comparison
+        public delegate bool IsLowerOrHigher(int x, int y);
+
+        public int[] SelectionSort(int[] ara, IsLowerOrHigher isLowerOrHigher)
+        {
+            if (ara == null)
+                throw new ArgumentNullException("The Integer Array is Null.");
```
mmwoodfo:
```suggestion
                throw new ArgumentNullException(nameof(ara));
```

DelegateFun/Sorter/SortUtility.cs
```diff
@@ -1,10 +1,35 @@
...
+            else if (isLowerOrHigher == null)
+                throw new ArgumentNullException("The Delegator is Null.");
```
mmwoodfo:
```suggestion
                throw new ArgumentNullException(nameof(isLowerOrHigher));
```

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,78 @@
...
+        public void SortUtility_ShouldSort_UsingAnAnonymousMethod(int[] toBeSorted, int[] expected)
+        {
+            SortUtility sortUtility = new SortUtility();
+            toBeSorted = sortUtility.SelectionSort(toBeSorted, delegate (int x, int y)
+            {
+                if (x < y)
+                    return true;
+                else
+                    return false;
+            });
```
NikolasGraham: This can be shortened to a one line return statement. Since you're sorting ascending you can use `return x < y`.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,78 @@
...
+                else
+                    return x < y;
+            });
+
+            CollectionAssert.AreEqual(expected, toBeSorted);
+        }
+
+        [TestMethod]
+        [ExpectedException(typeof(ArgumentNullException))]
+        public void SortUtility_IntNull()
```
NikolasGraham: Your naming scheme sounds odd here. Since an int cant be null, and what's actually null is your array. So maybe ArrayNull instead of IntNull.

DelegateFun/Sorter.Tests/SortUtilityTests.cs
```diff
@@ -1,13 +1,78 @@
...
+        [DataTestMethod]
+        [DataRow(new int[] { 4, 7, 2, 8, 9 }, new int[] { 2, 4, 7, 8, 9 })]
+        [DataRow(new int[] { 12, 17, 23, 18, 29 }, new int[] { 12, 17, 18, 23, 29 })]
+        [DataRow(new int[] { 4, 37, 56, 13, 1 }, new int[] { 1, 4, 13, 37, 56 })]
+        public void SortUtility_ShouldSort_UsingAnAnonymousMethod(int[] toBeSorted, int[] expected)
+        {
+            SortUtility sortUtility = new SortUtility();
+            toBeSorted = sortUtility.SelectionSort(toBeSorted, delegate (int x, int y)
+            {
+                if (x < y)
```
breaman: could have just been return x < y;

DelegateFun/Sorter.Tests/SortUtilityTests.cs
breaman: Agreed, the naming here isn't quite clear as to what the test is actually doing.

PR #284, opened 2019-11-20
Assignment 8

AaronSauther:
Assignment is finished. Have multiple AD0001 warnings that do not point to any line in any file that is generating the warning. After online research, all reported instances of this warning that I could find were confirmed to be a glitch with Microsoft analyzers. After discovering the AD0001 warning was a bug I tried to suppress the AD0001 warnings with nowarn statements the the csproj files but no dice. Any advice on how to deal with these would be greatly appreciated.


Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,34 @@
...
+        public void CsvRows_MockCsv_ReturnsCorrectContents()
+        {
+            //Arrange
+            SampleData sampleData = new SampleData(Environment.CurrentDirectory + "//MockPeople.csv");
+            string[] expectedData = File.ReadAllLines(Environment.CurrentDirectory + "//MockPeople.csv").Skip(1).ToArray();
+            string[] actualData = sampleData.CsvRows.ToArray();
+            //Act
+            bool ContentsMatch = Enumerable.SequenceEqual<string>(expectedData, actualData);
+            //Assert
+            Assert.IsTrue(ContentsMatch);
```
samrnixon: Looks like a great test method! Nice and simple read through.

Assignment.Tests/AddressTests.cs
```diff
@@ -0,0 +1,106 @@
...
+            Address spongebobsAddress = new Address(csvRow);
+            Address mrKrabsAddress = new Address(csvRowTwo);
+            //Act
+            bool addressesAreEqual = spongebobsAddress.Equals(mrKrabsAddress);
+            //Assert
+            Assert.IsFalse(addressesAreEqual);
+        }
+
+        [TestMethod]
+        public void GetHashCode_SameAddresses_ReturnsSameHashcode()
```
twofingerrightclick: The assignment doesn't require Address to have an equals method to accomplish any of the tasks. But you definitely did a thorough job implementing and testing it. It makes sense to test the Address and Person classes if they had logic within, but since they are just a list of properties and have no functions that are ever used in the SampleData.App, it could be argued that they do not need to be tested independently of the SampleData.App

Assignment.Tests/PersonTests.cs
```diff
@@ -0,0 +1,107 @@
...
+            Person spongebob = new Person(csvRow);
+            //Assert
+            Assert.AreEqual<string>("Spongebob", spongebob.FirstName);
+        }
+
+        [TestMethod]
+        public void Constructor_CsvRow_CorrectLastName()
+        {
+            //Arrange
+            string csvRow = "1,Spongebob,Squarepants,jellyfishgod@gmail.com,1391 Mudlick Road,Spokane,WA,99202";
```
twofingerrightclick: You use this same string in many places. Consider making it a field in the class. Always convenient to do something that makes commonly used values accessible via a variable in some way.

Assignment.Tests/AddressTests.cs
twofingerrightclick: More like make tests to check if the SampleData class uses the Address and Person classes correctly.

Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,158 @@
...
+        {
+            //Arrange
+            SampleData sampleData = new SampleData(Environment.CurrentDirectory + "\\MockPeople.csv");
+            //Act
+            //Assert
+            Assert.AreEqual<int>(sampleData.CsvRows.Count(), sampleData.CsvRows.Count());
+        }
+
+        [TestMethod]
+        public void CsvRows_MockCsv_ReturnsCorrectContents()
```
twofingerrightclick: This is a good test. Going to implement mine the same way

twofingerrightclick: And in the future you don't need to convert the queries to arrays. Here is what I have:
`SampleData sampleData = new SampleData(_TestFilePath);

            IEnumerable<string> actualData = sampleData.CsvRows;
            IEnumerable<string> expectedData = File.ReadAllLines(_TestFilePath).Skip(1);

            bool contentsMatch = Enumerable.SequenceEqual<string>(expectedData, actualData);

            Assert.IsTrue(contentsMatch);`

Assignment.Tests/AddressTests.cs
AaronSauther: Thanks for the review! Iwould agree that equals and hashcode are not required, but I went through and added those methods to person and address so I could use "assert.equals()" or "Enumerable.SequenceEqual()" dynamically without having to write any special code when validating addresses or people in tests. GetHashcode() was in no way needed, but I wrote it on the principle that when equals is implemented hashcode must also be implemented.

Assignment.Tests/PersonTests.cs
AaronSauther: Excellent point, I will refactor my code accordingly. Thanks!

Assignment.Tests/SampleDataTests.cs
AaronSauther: Thanks! I know just what you mean, as I grappled with converting the queries to arrays or just leaving them as IEnumerable. I chose to convert to arrays because I wanted better readability for my tests in terms of debugging since you can only see "current" property of IEnumerator in the debugger for data types of IEnumerable (or at least so i thought).

Assignment.Tests/SampleDataTests.cs
AaronSauther: Thanks!

twofingerrightclick:
Gotcha on the Equals method for It's use in Enumerable sequence... That's a good way to do that.

Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,172 @@
...
+            bool isSorted = Enumerable.SequenceEqual(actualData, actualData.OrderBy(state => state));
+            //Assert
+            Assert.IsTrue(isSorted);
+        }
+
+        [TestMethod]
+        public void GetAggregateSortedListOfStatesUsingCsvRows_MockCsv_ReturnsCorrectContents()
+        {
+            //Arrange
+            SampleData sampleData = new SampleData(Environment.CurrentDirectory + "//MockPeople.csv");
```
MarkMichaelis: This doesn't work if your run the test from a different directory.  Use something like AppDomain.CurrentAppDomain.BaseDirectory

MarkMichaelis:
##  (AaronSauther)

## Assignment8

| The assignment contains only a few required elements or has significant problems. | **5**


| Assignment  | **20** |
| 1. Implement the `ISampleData.CsvRows` property, loading the data from the `People.csv` file and returning each line as a single string.✔ <br/>- Change the "Copy to" property on People.csv to "Copy if newer" so that the file is deployed along with your test project.✔ <br/> - Using LINQ, skip the first row in the `People.csv`.✔ <br/> - Be sure to appropriately handle resource (`IDisposable`) items correctly if applicable (and it may not be depending on how you implement it). ✔  | ✔ |
| 2. Implement `IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()` to return a **sorted**, **unique** list of states.✔ <br/> Use `ISampleData.CsvRows` for your data source.✔ <br/>- Don't forget the list should be unique.✔ <br/>- Sort the list alphabetically ✔ <br/>- Include a test that leverages a hard coded list of Spokane based addresses. ✔ <br/>- Include a test that uses LINQ to verify the data is sorted correctly (do not use a hard coded list). ✔  | ✔ |
| 3. Implement `ISampleData.GetAggregateSortedListOfStatesUsingCsvRows()` to return a `string` that contains a **unique**, comma separated list of states.✔ <br/>- Use `ISampleData.GetUniqueSortedListOfStatesGivenCsvRows` for your data source.✔ <br/>- Given the array, consider using `string.Join` to combine the list into a single string.  | ✔ |
| 4.Implement the `ISampleData.People` property to return all the items in `People.csv` as `Person` objects ✔ <br/>- Use `ISampleData.CsvRows` as the source of the data. ✔ <br/>- Sort the list by State, City, Zip. (Sort the addresses first then select). ✔ <br/>- Be sure that `Person.Address` is also populated. ✔ <br/>- Adding null validation to all the `Person` and `Address` properties is **optional**. <br/>- Consider using `ISampleData.CsvRows` in your test to verify your results.  | ✔ |
| 5. Implement `ISampleDate.FilterByEmailAddress(Predicate<string> filter)` to return a list of names where the email address matches the `filter`. ✔ <br/>- Use `ISampleData.People` for your data source. ✔  | ✔ |
| 6. Implement `ISampleData.GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people)` to return a `string` that contains a **unique**, comma separated list of states. ✔ <br/>- Use the `people` parameter from `ISampleData.GetUniqueListOfStates` for your data source. ✔ <br/>- At a minimum, use `System.Linq.Enumerable.Aggregate` LINQ method to create your result. ✔ <br/>- Don't forget the list should be unique. ✔ <br/>- It is recommended that at a minimum you use `ISampleData.GetUniqueSortedListOfStatesGivenCsvRows` to validate your result.  | ✔ |


| Project Analysis |   |
| -----------------| - |
| Disabled Warnings **** | **✔**
| Nullable Enabled: | **✔** |
| Appropriately handle all warnings | **✔** |


| Assignment | Enable | Assignment8.sln |  |  | 4 |
| Assignment.Interfaces.Tests | Enable |  |  |  | 4 |
| Assignment.Tests | Enable | Assignment8.sln |  |  | 4 |


| Testing | **20** |
| Test Run Succeeded: Total tests=30, Passed=30, Failed=0, Skipped=0 | **100%** |


| Extra Credit | **0** |
| Write awesome, well tested code that follows all the guidelines outlined in the book. | ✔ |
| Add support for an `IReadOnlyArray` interface (for you to define) that provides readonly access to the count and each element in the array.  Be sure to appropriately decorate it for covariance and contravariance. | ✔ |
| Implement all homework using async/await and multi-threading by defining a new `SampleDataAsync` class that implements `ISampleData`). Refactor both your SampleData and SampleDataAsync classes so there is minimal duplication.  This includes refactoring your tests so a significant amount of test code can be re-used.| ✔ |

PR #288, opened 2019-11-23
Assignment 8

AustinHall7798:


Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        [TestCleanup]
+        public void TestCleanup()
+        {
+            if (File.Exists(_FileName))
+                File.Delete(_FileName);
+        }
+
+        [TestMethod]
+        public void CsvRows_ReadsAllButFirstLine_ReturnsLines()
```
GamerBah: Not sure a test like this is necessary, as all it's showing is that `File.ReadAllLines(path).Skip(1)` works.

PR #293, opened 2019-11-24
Assignment 8

Fervidusletum:
Finished


Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,87 @@
...
+        }
+
+        [TestMethod]
+        public void GetUniqueSortedListOfStatesGivenCsvRowsTest_ReturnsSortedListOfStates_ListInAscendingOrder()
+        {
+            List<string> sutData = (List<string>) SUT.GetUniqueSortedListOfStatesGivenCsvRows();
+            IEnumerable<string> sorted = new List<string>(sutData) // copy data
+                .OrderBy(state => state).ToList();
+
+            Assert.IsTrue(Enumerable.SequenceEqual(sorted, sutData), message: "Not Sorted.");
```
Cybotive: I don't know if using OrderBy in your test is a good way to test the OrderBy in SampleData. What if both are wrong?

Assignment/Address.cs
```diff
@@ -2,9 +2,9 @@
 {
     public class Address : IAddress
     {
-        public string StreetAddress { get; set; }
-        public string City { get; set; }
-        public string State { get; set; }
-        public string Zip { get; set; }
+        public string StreetAddress { get; set; } = "";
```
Cybotive: I'm not sure what the point of blank string is. But a sidenote is that Mark pushed an updated version of the interfaces so they're not as redundant, I'm not sure if they're going to require you use the updated version though.

Fervidusletum: I rebased onto that commit (896ede4c164489524dad26f45b16c2835359a754) before I really got started so I'm fine there, but I appreciate that you brought it up!

Assignment.Tests/SampleDataTests.cs
Fervidusletum: Agreed, they were pretty poorly written tests.

Assignment/SampleData.cs
```diff
@@ -11,32 +11,32 @@ public enum CsvColumn { Id, FirstName, LastName, Email, StreetAddress, City, Sta
...
-            // just some notes to future me here, don't mind them
-            List<string> states = (
-                from line in CsvRows
-                select line.Split(',')[(int)CsvColumn.State]) // transform the line into split strings
+            List<string> states =
+                // transform the line into tuple of states extracted from split strings
+                ( from line in CsvRows select line.Split(',')[(int) CsvColumn.State] )
                 .Distinct() // remove duplicates
                 .OrderBy((string state) => state) // compare by state string
                 .ToList(); // force query to execute
```
ashermancinelli: Might be easier to understand if you consistently use one version of the query, eg
```c#
            List<string> states = CsvRows
                .Select(line => line.Split(',')[(int) CsvColumn.State])
                .Distinct()
                .OrderBy((string state) => state)
                .ToList();
```

ashermancinelli: Might be easier to understand if you consistently use one version of the query, eg
```suggestion
            List<string> states = 
                // transform the line into tuple of states extracted from split strings
                CsvRows
                .Select(line => line.Split(',')[(int) CsvColumn.State])
                .Distinct() // remove duplicates
                .OrderBy((string state) => state) // compare by state string
                .ToList(); // force query to execute
```

Assignment/SampleData.cs
```diff
@@ -11,32 +11,32 @@ public enum CsvColumn { Id, FirstName, LastName, Email, StreetAddress, City, Sta
         private const string _CSV_FILENAME = "People.csv";
```
ashermancinelli: I think private variables are supposed to be in pascal case, eg 
```suggestion
         private const string _CsvFileName = "People.csv";
```

Assignment.Tests/SampleDataTests.cs
```diff
@@ -50,7 +53,7 @@ public void GetUniqueSortedListOfStatesGivenCsvRows_ReturnsSortedListOfStates_Li
             bool hasDescending = sutData
                 .Where((string state, int i) => i < sutData.Count - 1) // ignore last element which would cause index OoB
                 .Select((string state, int i) => (state, sutData[i + 1])) // pair up elements sequentially
-                .Select(((string First, string Second) pair) => // compare pairs of elements to check ascending order
+                .Select(((string First, string Second) pair) => // compare elements of pair to check ascending order
                         string.CompareOrdinal(pair.First, pair.Second) < 0
                 ).Any(isAscending => isAscending == false); // find any result that indicates descending order
```
ashermancinelli:
```suggestion
                ).Any(isAscending => !isAscending); // find any result that indicates descending order
```

Assignment/Address.cs
```diff
@@ -1,10 +1,78 @@
-﻿namespace Assignment
+﻿using System;
+
+namespace Assignment
 {
-    public class Address : IAddress
+    public class Address : IAddress, IEquatable<Address>, IComparable<Address>
```
MarkMichaelis: Wow... above and beyond. :)

Assignment/SampleData.cs
```diff
@@ -1,30 +1,112 @@
...
 namespace Assignment
 {
     public class SampleData : ISampleData
     {
+        public enum CsvColumn { Id, FirstName, LastName, EmailAddress, StreetAddress, City, State, Zip }
+        private const string _Csv_Filename = "People.csv";
+
         // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        private static IEnumerable<string> CsvDefault { get; } = File.ReadAllLines(_Csv_Filename).Skip(1).ToList();
```
MarkMichaelis: Why convert to ToList()?

PR #294, opened 2019-11-24
Assignment 8 Hw8 solution

dsergio:


Assignment/SampleData.cs
```diff
@@ -1,30 +1,79 @@
...
         public IEnumerable<(string FirstName, string LastName)> FilterByEmailAddress(
-            Predicate<string> filter) => throw new NotImplementedException();
+            Predicate<string> filter)
+        {
+            var ret =
+                from person in People
+                where filter(person.EmailAddress)
+                select (person.FirstName, person.LastName);
+            return ret;
+        }
```
ashermancinelli: This seems a little clearer to me:
```suggestion
            Predicate<string> filter) =>
                from person in People
                where filter(person.EmailAddress)
                select (person.FirstName, person.LastName);
```

Assignment/SampleData.cs
```diff
@@ -1,30 +1,79 @@
...
+            IEnumerable<IPerson> people)
+        {
+            IEnumerable<string> ppl =
+               from person in people
+               select person.Address.State;
+
+            string s = ppl.Distinct<string>().Aggregate((i, j) => i + "," + j);
+
+            return s;
+        }
```
ashermancinelli: I think this could be condensed a bit:
```suggestion
            IEnumerable<IPerson> people) =>
                ppl
                    .Select(p => p.Address.State)
                    .Distinct<string>()
                    .Aggregate((i, j) => i + "," + j);
```

PR #296, opened 2019-11-25
Assignment 8

CDobbins1:


Assignment/SampleData.cs
```diff
@@ -1,30 +1,60 @@
...
+                        {
+                            StreetAddress = line[(int)Column.StreetAddress],
+                            City = line[(int)Column.City],
+                            State = line[(int)Column.State],
+                            Zip = line[(int)Column.Zip]
+                        }
+                    })
+                    .OrderBy(person => person.Address!.State)
+                    .ThenBy(person => person.Address!.City)
+                    .ThenBy(person => person.Address!.Zip);
```
Fervidusletum: Might be a good idea to return a cache of the query call from functions, using `.ToList()` or any of the other `.To` functions as the last step in the query chain.  Could be a 'gotcha' if the function says it's returning an `IEnumerable<IPerson>` but is actually returning a deferred query for it.

Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,93 @@
...
+        {
+            var sut = new SampleData();
+
+            IEnumerable<string> query = sut.CsvRows;
+
+            Assert.AreEqual<int>(File.ReadAllLines("People.csv").Length - 1, query.Count());
+        }
+
+        [TestMethod]
+        public void GetUniqueSortedListOfStatesGivenCsvRows_ReturnsSortedList()
```
Fervidusletum: Assignment asked for an additional test of `GetUniqueSortedListOfStatesGivenCsvRows()` that `leverages a hard coded list of Spokane based addresses.`

PR #297, opened 2019-11-25
Assignment 8

dbosco051394:
first three items are done. Missing one test case for hardcoded list of spokane addresses. Not sure how to do that.

### DIFF
--- a/Assignment/SampleData.cs
+++ b/Assignment/SampleData.cs
@@ -7,13 +7,16 @@ namespace Assignment
 {
     public class SampleData : ISampleData
     {
+        public string? FileName { get; set; }
+
         // 1.
         public IEnumerable<string> CsvRows
         {
             get
             {
+                if (FileName is null || FileName.Length <= 0) FileName = @"People.csv";
                 IEnumerable<string> rows;
-                rows = from line in File.ReadAllLines(@"People.csv").Skip(1)
+                rows = from line in File.ReadAllLines(FileName).Skip(1)
                        select line;
                 return rows;
             }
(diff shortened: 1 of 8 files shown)


AssignmentTests/ISampleDataTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        [TestMethod]
+        public void GetAggregateSortedList()
+        {
+            SampleData sd = new SampleData();
+            string list = sd.GetAggregateSortedListOfStatesUsingCsvRows();
+            string expectedList = "AL,AZ,CA,DC,FL,GA,IN,KS,LA,MD,MN,MO,MT,NC,NE,NH,NV,NY,OR,PA,SC,TN,TX,UT,VA,WA,WV";
+
+            Assert.AreEqual(expectedList, list);
+        }
+    }
```
TatyanaHubbard: For the HardCoded list, you can just create a constructor and keep a private file name to test you're own file

Assignment/SampleData.cs
```diff
@@ -1,20 +1,50 @@
...
+using System.IO;
+using System.Linq;
 
 namespace Assignment
 {
     public class SampleData : ISampleData
     {
         // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        public IEnumerable<string> CsvRows
```
JerettLatimer: Your implementation looks good! I like how you used the other syntax.

AssignmentTests/ISampleDataTests.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void GetUniqueSortedListOfStates()
+        {
+            SampleData sd = new SampleData();
+            IEnumerable<string> rows = sd.CsvRows;
+            IEnumerable<string> states = sd.GetUniqueSortedListOfStatesGivenCsvRows();
+
+            
+            IEnumerable<string> expectedRows = from s in rows
+                                               let line = s.Split(',')
+                                               let state = line[6]
```
JerettLatimer: Try to avoid using "magic" numbers. I used an enum as the index so that you can clearly tell what data you are pulling from the array.

Assignment/SampleData.cs
```diff
@@ -1,20 +1,50 @@
...
+        {
+            get
+            {
+                IEnumerable<string> rows;
+                rows = from line in File.ReadAllLines(@"People.csv").Skip(1)
+                       select line;
```
pnotequalnp:
```cs
from item in collection
select item;
```
Code of this form never has any effect.

Assignment/SampleData.cs
```diff
@@ -1,20 +1,50 @@
...
-        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+        public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows()
+        {
+            IEnumerable<string> rows = CsvRows;
+            IEnumerable<string> sortedStates;
+            sortedStates = from line in rows
+                           let s = line.Split(',')
+                           orderby s[6]
+                           select s[6];
```
pnotequalnp: Since a projection (select) generally reduces the size of a collection, it is generally more performant to sort afterwards.

Assignment/SampleData.cs
```diff
@@ -1,20 +1,50 @@
...
         public string GetAggregateSortedListOfStatesUsingCsvRows()
-            => throw new NotImplementedException();
+        {
+            IEnumerable<string> rows = CsvRows;
+            IEnumerable<string> sortedStates = GetUniqueSortedListOfStatesGivenCsvRows();
+            List<string> s = sortedStates.ToList();
+            string aggregateSortedList = "";
+            aggregateSortedList = String.Join(',', s);
+
+            return aggregateSortedList;
```
pnotequalnp: What is the point of initializing this to the empty string before assigning it the intended value, then immediately returning it?

```cs
return String.Join(',', s);
```

AssignmentTests/AssignmentTests.csproj
```diff
@@ -0,0 +1,34 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
+  </PropertyGroup>
```
pnotequalnp: Don't forget to enable nullable reference types in your tests.

pnotequalnp:
The pull request title should start with `Assignment 8` otherwise it won't show up to be graded.

PR #298, opened 2019-11-25
Assignment 8

TatyanaHubbard:
Started fixing Assignment 8, Something weird happened to my Tests, so I will have to re-create it.


Assignment/SampleData.cs
```diff
@@ -1,30 +1,105 @@
...
+                City = person[(int)Column.City],
+                State = person[(int)Column.State],
+                Zip = person[(int)Column.Zip],
+                Address = new Address
+                {
+                    StreetAddress = person[(int)Column.StreetAddress],
+                    City = person[(int)Column.City],
+                    State = person[(int)Column.State],
+                    Zip = person[(int)Column.Zip]
+                }
```
dnisyndrum: Could maybe move this into a separate function for conciseness.

Assignment.Interfaces/IPerson.cs
```diff
@@ -6,11 +6,12 @@ namespace Assignment
...
-        string City { get; }
-        string State { get; }
-        string Zip { get; }
+        string? FirstName { get; }
+        string? LastName { get; }
+        string? Email { get; }
+        string? StreetAddress { get; }
+        string? City { get; }
+        string? State { get; }
+        string? Zip { get; }
```
mmwoodfo: I would suggest removing street address, city, state and zip as this is all taken care of in the address object Person has.

PR #299, opened 2019-11-26
Assignment8

jake-mcg:


Assignment/SampleData.cs
```diff
@@ -1,30 +1,63 @@
...
     public class SampleData : ISampleData
     {
         // 1.
-        public IEnumerable<string> CsvRows => throw new NotImplementedException();
+        public IEnumerable<string> CsvRows => File.ReadAllLines("People.csv").Skip(1).ToList();
 
         // 2.
         public IEnumerable<string> GetUniqueSortedListOfStatesGivenCsvRows() 
-            => throw new NotImplementedException();
+            => CsvRows.Select(ParsePersonText).Select(person => person.Address.State).Distinct().OrderBy(s => s);
```
crackerJackTech: For this return statement, it's easier to read if each call is separated per line.
"CsvRows.Select(ParsePersonText)
 .Select(person => person.Address.State)
 .Distinct()
 .OrderBy(s => s);"

PR #300, opened 2019-11-26
Assignment 8

kenny2892:


Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,180 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Assignment.Tests
+{
+    [TestClass]
+    public class SampleDataTests
+    {
```
NikolasGraham: Not necessary, but it's really useful to have a [testinitalize] and [testcleanup] for projects like this, especially since you're constantly creating files for testing, it's an easy way to just have one setup for you.

Assignment/SampleData.cs
```diff
@@ -1,30 +1,96 @@
...
+                        select item.Split(",")
+
+                        into line
+                        where line.Length == 8
+                        orderby
+                        line[6],
+                        line[5],
+                        line[7]
+
+                        select new Person()
```
NikolasGraham: Making a new person Object might be best to do in a separate method.

Assignment/SampleData.cs
```diff
@@ -1,30 +1,96 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
```
NikolasGraham: Your code overall looks really good. I don't have a lot to add other than it's nice to see .select/.orderby/etc, on different lines just because I think it adds more readability. I also might have made a new function for creating a Person for your 'people'.

PR #301, opened 2019-11-26
Finished Assignment 8

markdietzler:
Finished (I hope) the assignment. It's late, I'm frustrated, and I'm going to bed, I will review tomorrow.

### DIFF
--- a/Assignment/Assignment.csproj
+++ b/Assignment/Assignment.csproj
@@ -3,7 +3,13 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
     <NoWarn>CA1065</NoWarn>
+    <nullable>enable</nullable>
+    <NoWarn>AD0001</NoWarn>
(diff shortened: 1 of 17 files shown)


Assignment/Assignment.csproj
```diff
@@ -3,8 +3,14 @@
   <PropertyGroup>
     <TargetFramework>netstandard2.1</TargetFramework>
     <NoWarn>CA1065</NoWarn>
+    <nullable>enable</nullable>
+    <NoWarn>AD0001</NoWarn>
   </PropertyGroup>
 
+  <ItemGroup>
```
AustinHall7798: I think your checks may be failing because your missing this

PreserveNewest

from your assignment.csproj file. To add this (on mac, I'm not sure about windows) you right-click the people.csv file in the solution explorer, go to properties, and then under the "output to directory" property, change the selection to "copy if newer"

PR #303, opened 2019-11-26
Assignment8

NikolasGraham:
Finished, no extra credit attempted.


Assignment.Tests/SampleDataTests.cs
```diff
@@ -0,0 +1,284 @@
...
+            "4,Molly,Jeannot,mjeannotp@google.ca,00546 International Alley,Spokane,WA,99021",
+            "5,Maria,Rawsthorne,mrawsthorneq@slate.com,609 Kedzie Alley,Spokane,WA,99021",
+            "6,Celestyna,Robken,crobken12@t.co,27 Moland Parkway,Spokane,WA,99021",
+            "7,Selia,Bowe,sbowe13@360.cn,42 Lerdahl Plaza,Spokane,WA,99021",
+            "8,Arthur,Myles,amyles1c@miibeian.gov.cn,4718 Thackeray Pass,Spokane,WA,99021",
+            "9,Claudell,Leathe,cleathe1d@columbia.edu,30262 Steensland Way,Spokane,WA,99021"
+            };
+            using (StreamWriter SW = new StreamWriter(TempFilePathSpokane))
+            {
+                SW.WriteLine("Id,FirstName,LastName,Email,StreetAddress,City,State,Zip");
```
Tpirtle: SW.WriteLine("Id,FirstName,LastName,Email,StreetAddress,City,State,Zip"); I dont think this is needed, you are already have this line in the string[spokane].

Assignment/SampleData.cs
```diff
@@ -1,30 +1,111 @@
...
+            }
+            FileName = fileName;
+        }
+
+        // 1.   
+        public IEnumerable<string> CsvRows
+        {
+            get
+            {
+                foreach (var line in File.ReadAllLines(FileName).Skip(1).ToArray())
```
RSXEnthusiast: Don't think you need the foreach here. I think you could just do CsvRows => File.ReadAllLines(FileName).Skip(1)

PR #307, opened 2019-11-26
Assignmet8 1

Tpirtle:
I believe I have completed everything


Assignment8.Tests/GlobalSuppressions.cs
```diff
@@ -6,3 +6,4 @@
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>", Scope = "member", Target = "~M:Assignment8.Tests.SampleDataTests.GetUniqueSortedListOfStatesGivenCsvRows_HardCodedSpokane")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>", Scope = "member", Target = "~M:Assignment8.Tests.SampleDataTests.GetUniqueSortedListOfStatesGivenCsvRows_AllDistinctUingLINQ")]
 [assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>", Scope = "member", Target = "~M:Assignment8.Tests.SampleDataTests.GetUniqueSortedListOfStatesGxivenCsvRows_AllDistinctUingLINQ")]
+[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>", Scope = "member", Target = "~M:Assignment8.Tests.SampleDataTests.SampleDataConstructor_Null")]
```
dougdoner: You can replace duplicate suppressions with a single suppression rule by removing the Scope and Target properties from the rule, i.e.:
`[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Naming", "CA1707:Identifiers should not contain underscores", Justification = "<Pending>"]`

dougdoner: You will also want to put in a justification for why you put in the rule.

Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
+        }
 
         // 4.
-        public IEnumerable<IPerson> People => throw new NotImplementedException();
+        public IEnumerable<IPerson> People
+            => CsvRows.Where(myList
+                => myList.Split(",").Length == 8
+            ).OrderBy(
+                state => state.Split(",")[6]
+            ).ThenBy(
```
dougdoner: Good use of ThenBy after OrderBy

Assignment/SampleData.cs
```diff
@@ -1,30 +1,100 @@
...
 
         // 6.
-        public string GetAggregateListOfStatesGivenPeopleCollection(
-            IEnumerable<IPerson> people) => throw new NotImplementedException();
+        public string GetAggregateListOfStatesGivenPeopleCollection(IEnumerable<IPerson> people)
+        {
+            if (people == null)
+                throw new ArgumentNullException("People is null");
+
+            string[] stateAra = people.Select(person => person.Address.State).Distinct().ToArray();
```
dougdoner: You don't need to necessarily use the ToArray() method here, you can probably just return the Aggregate of people.Select anyway, i.e.:
`return people.Select(person => person.Address.State).Distinct().Aggregate((state, nextState) => $"{state},{nextState}");`

PR #308, opened 2019-11-27
Assignment 9 - mmwoodfo

mmwoodfo:


ShoppingList/ShoppingList/ItemStrikeThroughConverter.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace ShoppingList
+{
+    public class ItemStrikeThroughConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if (value is true)
+                return TextDecorations.Strikethrough;
+            else
+                return null!;
```
pnotequalnp: If your plan is to return null, you should allow null as a return value instead of overriding your own type signature.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,172 @@
...
+            MoveItemDownCommand = new RelayCommand(OnMoveItemDown);
+            CrossOffCommand = new RelayCommand(OnCrossOff);
+            ShowHelpCommand = new RelayCommand(OnShowHelp);
+            ExportCommand = new RelayCommand(OnExport);
+        }
+
+        //---------- BINDING FUNCTIONS -----------//
+        private void OnAddItem()
+        {
+            if (!string.IsNullOrEmpty(TextToAddToList))
```
pnotequalnp: Perhaps also exclude strings consisting only of whitespace.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,172 @@
...
+            }
+        }
+
+        private void OnShowHelp()
+        {
+            if (ShowPopUp)
+                ShowPopUp = false;
+            else
+                ShowPopUp = true;
+        }
```
pnotequalnp:
```cs
private void OnShowHelp() =>
    ShowPopUp = !ShowPopUp;
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
mmwoodfo: That is probably a good call - Always expect the worst from users. :)

ShoppingList.Tests/ItemStrikeThroughConverterTests.cs
```diff
@@ -0,0 +1,39 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System.Windows;
+
```
NikolasGraham: I like your implementation for crossing off items. I ended doing the same thing as I didn't like my previous version.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,374 @@
...
+
+//Creating mock objects (for my future reference)
+//https://intellitect.com/unit-testing-with-mocks/
+
+namespace ShoppingList.Tests
+{
+    [TestClass()]
+    public class MainWindowViewModelTests
+    {
+
```
NikolasGraham: Your formatting is really nice with the way your using comments to separate sections.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+
+        private void OnDeleteItem()
+        {
+            if (SelectedItem != null)
+            {
+                ShoppingList.Remove(SelectedItem);
+                SelectedItem = null;
+            }
+        }
+
```
NikolasGraham: I had a sort that just did so alphabetically, but I really like your implementation to just move each individual item around wherever you want.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+            ShowHelpCommand = new RelayCommand(OnShowHelp);
+            ExportCommand = new RelayCommand(OnExport);
+
+            _DialogService = dialogService;
+        }
+
+        //---------- BINDING FUNCTIONS -----------//
+        private void OnAddItem()
+        {
+            if (!string.IsNullOrWhiteSpace(TextToAddToList))
```
NikolasGraham: I assume you don't care if the user has lots of white space before any text?

mmwoodfo: No, I figured if a user for whatever reason wants to add whitespace in their item name then so be it. They have the freedom to edit/delete it if they change their mind. As long as an empty item name isn't added to the list.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -1,12 +1,171 @@
...
+        MinHeight="350"
+        MinWidth="650"
+        TextElement.FontSize="20"
+        x:Name="MainWindowSizing"
+        WindowStartupLocation="CenterScreen"
+        md:DialogServiceViews.IsRegistered="True"
+        DataContext="{Binding Source={StaticResource Locator}, Path=MainWindow}"
+        Title="Shopping List" Height="450" Width="800">
+
+    <Window.Resources>
```
twofingerrightclick: This app is awesome! Nice work!

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -1,12 +1,171 @@
...
+                Command="{Binding ExportCommand}" 
+                Grid.Column="1" Grid.Row="4"
+                VerticalAlignment="Bottom"/>
+
+        <!--R1-->
+        <TextBox Text="{Binding SelectedItem.Name, UpdateSourceTrigger=PropertyChanged}" 
+                 Grid.Column="1" Grid.Row="1" Grid.ColumnSpan="{Binding SpanAll}"
+                 Visibility="{Binding SelectedItem, Converter={StaticResource ItemVisibilityConverter}}"/>
+        
+        <!--R2-->
```
twofingerrightclick: I like the coordinate comments.

ShoppingList/ShoppingList/Item.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    public class Item
+    {
+        public string Name { get; set; }
+        public bool CheckedOff { get; set; }
+
+        public Item(string name)
+        {
+            Name = name ?? throw new ArgumentNullException(nameof(name));
+            CheckedOff = false;
```
Keboo: This is not needed as the default of a `bool` is `false`

ShoppingList/ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if (value is Item)
+                return "Visible";
```
Keboo: You probably wanted to use the `Visibility` enum rather than using strings. The type converters are what are saving this and making it work.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -1,12 +1,171 @@
...
+        md:DialogServiceViews.IsRegistered="True"
+        DataContext="{Binding Source={StaticResource Locator}, Path=MainWindow}"
+        Title="Shopping List" Height="450" Width="800">
+
+    <Window.Resources>
+        <local:ItemVisibilityConverter x:Key="ItemVisibilityConverter"/>
+        <local:ItemStrikeThroughConverter x:Key="ItemStrikeThroughConverter"/>
+
+        <!-- ListBox Style -->
+        <Style TargetType="ListBox">
```
Keboo: This striped off the base style, is that intentional?

ShoppingList/ShoppingList/MainWindow.xaml.cs
```diff
@@ -7,9 +7,17 @@ namespace ShoppingList
     /// </summary>
     public partial class MainWindow : Window
     {
+        public ViewModelLocator ViewModel { get; }
```
Keboo: This name is pretty misleading since it is not a view model but rather a view model locator.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+        public ObservableCollection<Item> ShoppingList { get; } = new ObservableCollection<Item>();
+
+        private string _TextToAddToList = "";
+        public string TextToAddToList
+        {
+            get => _TextToAddToList;
+            set => Set(ref _TextToAddToList, value);
+        }
+
+        private Item? _SelectedItem = null;
```
Keboo: Default value is not needed here.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,170 @@
...
+        public Item? SelectedItem
+        {
+            get => _SelectedItem!;
+            set => Set(ref _SelectedItem, value);
+        }
+
+        private bool _ShowPopUp = false;
```
Keboo: Default value is not needed.

PR #312, opened 2019-12-01
as9

ashermancinelli:
WIP


ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,73 @@
...
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition Width="*"/>
+            <ColumnDefinition Width="0.5*"/>
+            <ColumnDefinition Width="0.5*"/>
+        </Grid.ColumnDefinitions>
+
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="Auto"/>
```
mmwoodfo:
```suggestion
            <RowDefinition Height="*"/>
```
Style suggestion - this will expand your listbox to be full page length

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,73 @@
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
         mc:Ignorable="d"
-        Title="MainWindow" Height="450" Width="800">
+        d:DataContext="{d:DesignInstance Type=local:ListModel}"
```
mmwoodfo:
```suggestion
        d:DataContext="{d:DesignInstance Type=local:ListModel}"
        MinHeight="350"
        MinWidth="650"
```
Prevent the user from making the screen too small - again, style suggestion. Looks great otherwise. :)

ShoppingList/ShoppingList/ViewModelLocator.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace ShoppingList
+{
+    public class ViewModelLocator
+    {
+        public ViewModelLocator()
+        {
+            SimpleIoc.Default.Register<MainWindowViewModel>();
+        }
+
+        public MainWindowViewModel mainWindow => SimpleIoc.Default.GetInstance<MainWindowViewModel>();
```
Keboo: Properties should be PascalCase
Member can be made static

ShoppingList/ShoppingList/Item.cs
```diff
@@ -0,0 +1,17 @@
+﻿namespace ShoppingList
+{
+    public class Item
+    {
+        public string Name { get; set; } = "";
```
Keboo: Default value here is not used in any execution path. Both constructors override it.

ShoppingList/ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,20 @@
...
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+		{
+            if (value is Item)
+				return "Visible";
```
Keboo: There is a `Visibility` enum that would be better to return here. The WPF type converters are the only reason these strings work.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,65 @@
...
+                 ItemsSource="{Binding ShoppingList}" SelectedItem="{Binding SelectedItem}"
+                 ItemTemplate="{StaticResource ItemTemplate}"
+                 HorizontalContentAlignment="Stretch"
+                 Grid.Row="0" Grid.Column="0"
+                 Grid.ColumnSpan="2"/>
+        
+        <TextBox Text="{Binding NameToAppend}"
+                 Grid.Row="1" Foreground="Black" Grid.Column="0">
+            <TextBox.InputBindings>
+                <KeyBinding Key="Enter" Command="{Binding AddItem}"
```
Keboo: This doesn't quite work here, because the TextBox is consuming all of the key presses.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,65 @@
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
         mc:Ignorable="d"
-        Title="MainWindow" Height="450" Width="800">
+        Title="Shopping List" Height="600" Width="800"
+        TextElement.Foreground="{DynamicResource MaterialDesignBody}"
+        Topmost="True">
```
Keboo: In general, you don't want to set `Topmost="True"`. I only did this in class to make it easier to show the app alongside of the code.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,65 @@
...
+        </TextBox>
+        
+        <Button Content="Add Item"
+                Command="{Binding AddItem}" CommandParameter="{Binding Text, ElementName=NewItem}"
+                Grid.Row="1" Grid.Column="1"/>
 
+        <DockPanel Visibility="{Binding SelectedItem, Converter={StaticResource ItemVisibilityConverter}}"
+                   Grid.Row="3" Grid.Column="0"
+                   Grid.ColumnSpan="2">
+            <TextBlock Text="Name:" Foreground="Black" Margin="10,5,10,5"/>
```
Keboo: When you put items inside of a DockPanel, you arrange them using the `DockPanel.Dock` attached property. The last item in the dock panel will fill the remaining space in the center. 
```suggestion
            <TextBlock Text="Name:" Foreground="Black" Margin="10,5,10,5" DockPanel.Dock="Left"/>
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,58 @@
...
+                var newItem = new Item { Name = NameToAppend };
+                ShoppingList.Add(newItem);
+                SelectedItem = newItem;
+                NameToAppend = "";
+            }
+        }
+
+        private void OnRemoveItem(Item item)
+        {
+            if (object.ReferenceEquals(SelectedItem, item))
```
Keboo: You technically don't need the explicit type here since all classes derive from `object`
```suggestion
            if (ReferenceEquals(SelectedItem, item))
```

PR #313, opened 2019-12-03
Assignment 9

Fervidusletum:
Finished

### DIFF
--- a/ShoppingList/ShoppingList/MainWindowViewModel.cs
+++ b/ShoppingList/ShoppingList/MainWindowViewModel.cs
@@ -7,6 +7,7 @@ namespace ShoppingList
     {
         public ObservableCollection<Item> ShoppingItems { get; } = new ObservableCollection<Item>();
 
+        private int SelectedIndex { get; set; }
         private Item? _SelectedListItem;
         public Item? SelectedListItem
         {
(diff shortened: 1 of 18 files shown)


ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+
+            SelectedListItem.Text = "";
+            OnDeselect();
+        }
+
+        // moves toward start of list
+        public ICommand MoveUpCommand { get; }
+        private void OnMoveUp()
+        {
+            if (SelectedListItem is null || SelectedIndex <= 0) return;
```
mmwoodfo: Wow this is a much better way to do it than what I was doing! I think I may refactor & credit you in my code because it's so much simpler! Great job. :)

mmwoodfo:
```suggestion
            if (!(SelectedItem is null) && index < ShoppingList.Count-1)
```
With that said, why not get rid of the return statement & just have the if statement like this?

Fervidusletum: Thanks!  I'm happy you found it helpful.

PR #314, opened 2019-12-03
Assignment 9

pnotequalnp:
WIP


pnotequalnp:
Seems done.

ShoppingList/Item.cs
```diff
@@ -3,11 +3,5 @@
     public class Item
     {
         public string Name { get; set; } = "New Item";
-
-        public override bool Equals(object? obj) =>
-            obj is Item item && Name == item.Name;
-
-        public override int GetHashCode() =>
-            Name.GetHashCode();
```
ashermancinelli: Overriding method does stop some compiler errors for me, but yeah probably better without unneeded code. Looks good!

ShoppingList/MainWindow.xaml
```diff
@@ -0,0 +1,63 @@
...
+            <RowDefinition />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition />
+            <ColumnDefinition Width="Auto" />
+        </Grid.ColumnDefinitions>
+        
+        <ListBox x:Name="ItemList"
+                 ItemsSource="{Binding Items}" SelectedItem="{Binding SelectedItem}"
```
twofingerrightclick: The KeyWord SelectedItem and the property SelectedItem seems confusing.

twofingerrightclick: Or is it convenient?

ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,15 @@
+﻿using System;
+using System.Globalization;
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null) =>
```
Keboo: There is a `Visibility` enum that would have been better here. The default WPF type converters are the reason why passing a string works. The actual Visibility property uses the enum.

ShoppingList/MainWindow.xaml
```diff
@@ -0,0 +1,63 @@
...
+                 Grid.Row="1" Grid.Column="0" />
+        
+        <Button Content="Add Item"
+                Command="{Binding AddItem}" CommandParameter="{Binding Text, ElementName=NewItem}"
+                Grid.Row="1" Grid.Column="1"/>
+
+        <DockPanel Visibility="{Binding SelectedItem, Converter={StaticResource ItemVisibilityConverter}}"
+                   Grid.Row="3" Grid.Column="0"
+                   Grid.ColumnSpan="2">
+            <TextBlock Text="Name:" Margin="10,5,10,5"/>
```
Keboo: When you put items inside of a DockPanel, you arrange them using the `DockPanel.Dock` attached property. The last item in the dock panel will fill the remaining space in the center. 
```suggestion
            <TextBlock Text="Name:" Margin="10,5,10,5" DockPanel.Dock="Left"/>
```

ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,57 @@
...
+                var newItem = new Item { Name = NewItemName };
+                Items.Add(newItem);
+                SelectedItem = newItem;
+                NewItemName = "";
+            }
+        }
+
+        private void OnRemoveItem(Item item)
+        {
+            if (object.ReferenceEquals(SelectedItem, item))
```
Keboo: You technically don't need the explicit type here since all classes derive from `object`
```suggestion
            if (ReferenceEquals(SelectedItem, item))
```

PR #315, opened 2019-12-05
Assignment 9

dnisyndrum:


ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,58 @@
...
+
+        <ListBox x:Name="ItemList"
+             ItemsSource="{Binding Items}" SelectedItem="{Binding SelectedItem}"
+             ItemTemplate="{StaticResource ItemTemplate}"
+             HorizontalContentAlignment="Stretch"
+             Grid.Row="1" 
+             Grid.Column="0"
+             Grid.ColumnSpan="2"
+             Background="LightCyan"
+        />
```
CDobbins1: I like the way this is organized. Very readable.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,58 @@
...
+    
+    
+    
+    <Window.Resources>
+        <local:VisibilityConverter x:Key="VisibilityConverter"/>
+        <DataTemplate DataType="{x:Type local:Item}" x:Key="ItemTemplate">
+            <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*" />
+                </Grid.ColumnDefinitions>
```
Keboo: This is the default for `ColumnDefinitions` so this could be omitted.

ShoppingList/ShoppingList/MainWindowsModelView.cs
```diff
@@ -0,0 +1,37 @@
...
+        private Item? _SelectedItem;
+        private string _NewItemName;
+        public RelayCommand? AddItem { get; }
+        public Item? SelectedItem { get => _SelectedItem; set => Set(ref _SelectedItem, value); }
+        public string NewItemName { get => _NewItemName; set => Set(ref _NewItemName, value); }
+
+        public MainWindowViewModel()
+        {
+            AddItem = new RelayCommand(OnAddItem);
+            _NewItemName = "";
```
Keboo: Since you check the property with `string.IsNullOrWhiteSpace` this initialization is not actually needed.

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if(value is Item)
+            {
+                return "Visible";
```

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,58 @@
...
+                <TextBox Text="{Binding ItemName, UpdateSourceTrigger=PropertyChanged, Mode=TwoWay}"
```
Keboo: This means that the "edit" of an items is always visible. The intent was when no item is selected the view elements for editing an item should be hidden.

PR #318, opened 2019-12-06
Assignment9

AaronSauther:
Assignment finished, although I am working with the AD0001 Microsoft analyzer warning glitch.


ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,51 @@
...
+        {
+            Items = new ObservableCollection<ShoppingItem>();
+            Items.Add(new ShoppingItem("test1"));
+            Items.Add(new ShoppingItem("test1"));
+            AddItemCommand = new Command(OnAddItem);
+        }
+
+        public void OnAddItem()
+        {
+            Items.Add(new ShoppingItem("New Item"));
```
samrnixon: Looks good, if you don't want a white-space "item" to be added to the list though I would add an if statement checking for that before you add to the list.

ShoppingList/ShoppingList/ShoppingItem.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
+namespace ShoppingList
+{
+    public class ShoppingItem : INotifyPropertyChanged
+    {
+        public event PropertyChangedEventHandler? PropertyChanged;
+
+        private bool SetProperty<T>(ref T field, T newValue, [CallerMemberName]string? propertyName = null)
```
Keboo: Since this method is duplicated, it would have been good to create a common base class to avoid duplication.

PR #319, opened 2019-12-08
Assignment 9

JerettLatimer:


ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,46 @@
...
+            sut.NewName = testString;
+
+            // Act
+            sut.AddItemCommand.Execute(testString);
+
+            // Assert
+            Assert.AreEqual(0, sut.ShoppingList.Count);
+        }
+    }
+}
```
crackerJackTech: This is a great use of the DataRow.

ShoppingList/ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if(value is Item)
+            {
+                return "Visible";
+            }
+            return "Hidden";
```
crackerJackTech: I didn't think of using the string "Visible" or "Hidden" for the return values. Nice!

ShoppingList/ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ShoppingList
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if(value is Item)
+            {
+                return "Visible";
```
Keboo: There is a `Visibility` enum that would have been better here. The default WPF type converters are the reason why passing a string works. The actual Visibility property uses the enum.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,67 @@
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
-        mc:Ignorable="d"
+        mc:Ignorable="d" Topmost="True"
```

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,67 @@
...
+                </Grid.ColumnDefinitions>
+                <TextBlock Text="{Binding Name}" Grid.Column="0"/>
+            </Grid>
+        </DataTemplate>
+    </Window.Resources>
+
     <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="2*"/>
```
Keboo: Because the star sizing is just weights, and because the other rows are all Auto. This row will just fill the remaining space. So it is functionally equal to `Height="1*" which is also the default.
```suggestion
            <RowDefinition />
```

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,12 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWPF>true</UseWPF>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="IntelliTect.Analyzers" Version="0.1.3" />
+    <PackageReference Include="MvvmLightLibs" Version="5.4.1.1" />
```
Keboo: This is close, but this nuget package will cause the NU1701 warning because it does not support .NET Core. To make it work you wanted to use MvvmLightLibsStd10

PR #320, opened 2019-12-08
Assignment9

Cybotive:
image


ShoppingList/ShoppingList/Person.cs
```diff
@@ -0,0 +1,19 @@
+﻿using System;
+
+namespace ShoppingList
+{
+    public class Person
```
AustinHall7798: You might want to change the people/person names to something more similar to shoppingList/listItems

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,71 @@
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
-        mc:Ignorable="d"
-        Title="MainWindow" Height="450" Width="800">
-    <Grid>
+        mc:Ignorable="d" Topmost="True"
```

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,71 @@
...
+
+        <Style TargetType="Button" BasedOn="{StaticResource MyButtonStyle}">
+            <Style.Triggers>
+                <Trigger Property="IsPressed" Value="True">
+                    <Setter Property="Foreground" Value="DarkGray" />
+                </Trigger>
+            </Style.Triggers>
+        </Style>
+
+        <Style TargetType="ToggleButton" BasedOn="{StaticResource MyButtonStyle}">
```
Keboo: This style isn't being used since you don't have a ToggleButton

ShoppingList/ShoppingList/Person.cs
Keboo: Agreed

PR #321, opened 2019-12-08
assignment 9

dsergio:


ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+        }
+
+        public ObservableCollection<ShoppingItem> ShoppingItemList { get; } = new ObservableCollection<ShoppingItem>();
+
+        public Command AddItemCommand { get; }
+        public Command DeleteItemCommand { get; }
+
+        private bool CanExecute { get; set; }
+
+        private ShoppingItem _SelectedShoppingItem;
```
Fervidusletum:
```suggestion
        private ShoppingItem _SelectedShoppingItem = new ShoppingItem("Item", true, 1, 1, UnitType.Each);
```

Just a throw-away item. Safer than using null! or leaving it uninitialized, in case you don't initialize it and the list in the constructor.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+            ShoppingItemList.Add(new ShoppingItem("Coffee", true, 1, 1, UnitType.Pound));
+            ShoppingItemList.Add(new ShoppingItem("Bananas", true, 5, 1, UnitType.Each));
+            SelectedShoppingItem = ShoppingItemList.First();
+            CanExecute = true;
+            AddItemCommand = new Command(OnAddItem, () => CanExecute);
+            DeleteItemCommand = new Command(OnDeleteItem, () => CanExecute);
+        }
+
+        private void OnAddItem()
+        {
```
Fervidusletum:
```suggestion
        {
            if (SelectedShoppingItem is null) return;
            
```
If you delete an item, and then click add without filling in the boxes, the program throws `NullReferenceException`.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,77 @@
...
+
+        <StackPanel Grid.Column="0" Grid.Row="0">
+            <TextBlock Text="Name" />
+            <TextBox Text="{Binding SelectedShoppingItem.Name}"/>
+            <TextBlock Text="Quantity" />
+            <TextBox Text="{Binding SelectedShoppingItem.Quantity}"/>
+            <TextBlock Text="Size" />
+            <TextBox Text="{Binding SelectedShoppingItem.Size}"/>
+            <TextBlock Text="Unit" />
+            <TextBox Text="{Binding SelectedShoppingItem.Unit}"/>
```
Fervidusletum: A pull-down menu (combo box) would be helpful for selecting units.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,73 @@
...
+
+        public enum UnitType
+        {
+            Gallon,
+            FluidOunce,
+            Liter,
+            Gram,
+            Pound,
+            Each
+        }
```
dnisyndrum: Interesting to add different units for more specificity. I like the idea, adds some variety.

ShoppingList/ShoppingList/ShoppingItem.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ShoppingList
+{
+    public class ShoppingItem
+    {
+        public string Name { get; set; }
+        public bool Active { get; set; }
+        public int Quantity { get; set; }
+        public double Size { get; set; }
+        public UnitType Unit { get; set; }
```
dnisyndrum: Nice, I like all the options given to the user to make a very specific shopping list.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,102 @@
...
+                </DataTemplate>
+            </ListBox.ItemTemplate>
+
+        </ListBox>
+        <StackPanel Grid.Row="4" Grid.Column="0">
+            <Button Command="{Binding CreateNewItemCommand}" HorizontalContentAlignment="Center"
+                    >
+                <Button.Content>
+                    <Border>
+                        <TextBlock Text="Create New Shopping Item" />
```
Keboo: Because you are not adding anything with the Border or the TextBlock this string could be inlined as the content of the button directly.

ShoppingList/ShoppingList/ShoppingItemVisibleConverter.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace ShoppingList
+{
+    public class ShoppingItemVisibleConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if (value is ShoppingItem)
+            {
+                return "Visible";
```
Keboo: There is a `Visibility` enum that would have been better here. The default WPF type converters are the reason why passing a string works. The actual Visibility property uses the enum.

PR #322, opened 2019-12-08
Assignment9

dbosco051394:
Will have more changes by Sunday at midnight.

### DIFF
--- a/ShoppingList.Tests/MainWindowViewModelTests.cs
+++ b/ShoppingList.Tests/MainWindowViewModelTests.cs
@@ -1,0 +1,57 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace ShoppingList.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+
+        [TestMethod]
+        public void Constructor_SetsDefaultSelectedItem()
+        {
+            var viewmodel = new MainWindowViewModel();
+
+            Assert.AreEqual(viewmodel.ShoppingList.First(), viewmodel.SelectedItem);
+        }
+
+        [TestMethod]
+        public void AddItemCommand_AddsItem()
+        {
+            var viewmodel = new MainWindowViewModel();
+            int expectedCount = viewmodel.ShoppingList.Count;
+            viewmodel.Text = "TestItem";
+
+            viewmodel.AddNewItemCommand.Execute(null);
+
+            Assert.AreEqual(expectedCount + 1, viewmodel.ShoppingList.Count);
+        }
+
+        [TestMethod]
+        public void AddItemCommand_AddsItemWithCorrectName()
+        {
+            var viewmodel = new MainWindowViewModel();
+            string expectedText = "TestItem";
+            viewmodel.Text = expectedText;
+
+            viewmodel.AddNewItemCommand.Execute(null);
+
+            Assert.AreEqual(expectedText, viewmodel.ShoppingList.Last().Name);
+        }
+
+        [TestMethod]
+        public void EditsSelectedItem_ChangesItemInList()
+        {
+            var viewmodel = new MainWindowViewModel();
+            string expectedNewName = "NewName";
+            viewmodel.SelectedItem = viewmodel.ShoppingList.Last();
+
+            viewmodel.SelectedItem.Name = expectedNewName;
+
+            Assert.AreEqual(expectedNewName, viewmodel.ShoppingList.Last().Name);
+        }
+    }
+}
--- a/ShoppingList/ShoppingList/MainWindow.xaml
+++ b/ShoppingList/ShoppingList/MainWindow.xaml
@@ -5,8 +5,36 @@
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
         mc:Ignorable="d"
-        Title="MainWindow" Height="450" Width="800">
+        TextElement.FontSize="25"
+        d:DataContext="{d:DesignInstance Type=local:MainWindowViewModel}"
+        Title="MainWindow" MinHeight="450" MinWidth="800">
     <Grid>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition></ColumnDefinition>
+            <ColumnDefinition></ColumnDefinition>
+        </Grid.ColumnDefinitions>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition/>
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+        <StackPanel Grid.Row="1" VerticalAlignment="Center">
+            <TextBox Background="BurlyWood" Text="{Binding Text}" TextAlignment="Center"/>
+            <Button Content="Add New Item" Command="{Binding AddNewItemCommand}"/>
+        </StackPanel>
+        <StackPanel Grid.Row="0">
+            <TextBlock TextAlignment="Center" Text="Edit Item"/>
+            <TextBox Background="BurlyWood" TextAlignment="Center" Text="{Binding SelectedItem.Name, UpdateSourceTrigger=PropertyChanged}"/>
+        </StackPanel>
+        <ListBox Background="BurlyWood" Grid.Column="1" Grid.RowSpan="3" ItemsSource="{Binding ShoppingList}" SelectedItem="{Binding SelectedItem}">
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:ShoppingItem}">
+                    <StackPanel >
+                        <TextBlock Text="{Binding Name}"/>
+                    </StackPanel>
+                </DataTemplate>
+            </ListBox.ItemTemplate>
+        </ListBox>
 
     </Grid>
 </Window>
--- a/ShoppingList/ShoppingList/MainWindowViewModel.cs
+++ b/ShoppingList/ShoppingList/MainWindowViewModel.cs
@@ -29,8 +29,8 @@ namespace ShoppingList
             set => SetProperty(ref _Text, value);
         }
 
-        private ShoppingItem _SelectedItem;
-        public ShoppingItem SelectedItem
+        private ShoppingItem? _SelectedItem;
+        public ShoppingItem? SelectedItem
         {
             get => _SelectedItem;
             set => SetProperty(ref _SelectedItem, value);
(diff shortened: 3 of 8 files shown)


ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+            {
+                field = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+            }
+        }
+
+        private string _Text;
+        public string Text 
+        { 
+            get => _Text;
```
mmwoodfo:
```suggestion
            get => _Text = "";
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        public ICommand AddNewItemCommand { get; }
+
+        public ObservableCollection<ShoppingItem> ShoppingList { get; } = new ObservableCollection<ShoppingItem>();
+
+        public MainWindowViewModel()
+        {
+            AddNewItemCommand = new Command(OnAddNewItem);
+            Text = "";
```
mmwoodfo:
```suggestion

```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+            SelectedItem = ShoppingList.First();
+        }
+
+        private void OnAddNewItem()
+        {
+            if(!string.IsNullOrWhiteSpace(Text))
+            {
+                ShoppingList.Add(new ShoppingItem(Text));
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(ShoppingList)));
+                Text = "";
```
mmwoodfo:
```suggestion

```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+            }
```
mmwoodfo:
```suggestion
            }
            Text = "";
```
I suggest moving the clearing text option outside of the if statement because if a user puts whitespace into the textbox this will clear that whitespace whenever they try and add a new item (since whitespace isn't allowed it would be best to clear it for the user).

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        }
+
+        private string _Text;
+        public string Text 
+        { 
+            get => _Text;
+            set => SetProperty(ref _Text, value);
+        }
+
+        private ShoppingItem _SelectedItem;
```
mmwoodfo:
```suggestion
        private ShoppingItem? _SelectedItem;
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+        public ShoppingItem SelectedItem
```
mmwoodfo:
```suggestion
        public ShoppingItem? SelectedItem
```

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,23 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWPF>true</UseWPF>
+    <Nullable>enable</Nullable>
```
mmwoodfo:
```suggestion
    <Nullable>enable</Nullable>
    <NoWarn>AD0001</NoWarn>
```
I would suggest adding this to your tests project too.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,36 @@
...
+            <RowDefinition Height="Auto"/>
+            <RowDefinition/>
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+        <StackPanel Grid.Row="1" VerticalAlignment="Center">
+            <TextBox Background="BurlyWood" Text="{Binding Text}" TextAlignment="Center"/>
+            <Button Content="Add New Item" Command="{Binding AddNewItemCommand}"/>
+        </StackPanel>
+        <StackPanel Grid.Row="0">
+            <TextBlock TextAlignment="Center" Text="Edit Item"/>
```
Cybotive: Does this actually do anything? Or is it just text?

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,23 @@
...
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="IntelliTect.Analyzers" Version="0.1.3" />
+    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.9.8">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.NetCore.Analyzers" Version="2.9.8">
```
mmwoodfo:
```suggestion
    <PackageReference Include="Microsoft.NetCore.Analyzers" Version="2.9.6">
```

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,23 @@
...
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.NetFramework.Analyzers" Version="2.9.8">
```
mmwoodfo:
```suggestion
    <PackageReference Include="Microsoft.NetFramework.Analyzers" Version="2.9.6">
```

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,23 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWPF>true</UseWPF>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="IntelliTect.Analyzers" Version="0.1.3" />
+    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.9.8">
```
mmwoodfo:
```suggestion
    <PackageReference Include="Microsoft.CodeQuality.Analyzers" Version="2.9.6">
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+
+        public ICommand AddNewItemCommand { get; }
+
+        public ObservableCollection<ShoppingItem> ShoppingList { get; } = new ObservableCollection<ShoppingItem>();
+
+        public MainWindowViewModel()
+        {
+            AddNewItemCommand = new Command(OnAddNewItem);
+            Text = "";
+            ShoppingList.Add(new ShoppingItem("Bologna"));
```
Cybotive: These should probably not be here anymore.

ShoppingList/ShoppingList/ShoppingItem.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace ShoppingList
+{
+    public class ShoppingItem
+    {
+
+        public ShoppingItem(string v1)
+        {
+            this.Name = v1;
```
mmwoodfo:
```suggestion
            Name = name;
```
the 'this' keyword is used to reference the current object to distinguish between an object's data members and data members being passed into that object. Since your property name does not match your param name you do not need to use the 'this' keyword as there is no ambiguity. 

I would also suggest renaming your string parameter to 'name' to make your code more readable as it isn't clear what v1 stands for.

ShoppingList/ShoppingList/ShoppingItem.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace ShoppingList
+{
+    public class ShoppingItem
+    {
+
+        public ShoppingItem(string v1)
```
mmwoodfo:
```suggestion
        public ShoppingItem(string name)
```
Unclear what v1 means.

ShoppingList/ShoppingList/MainWindow.xaml
dbosco051394: It's just text for the edit box to let you know it's the box that edits the field. Could have also used a label, but stick with TextBlock for familiarity's sake.

ShoppingList/ShoppingList/MainWindowViewModel.cs
dbosco051394: I set Text to "" here so that the field is cleared after you add a new item to the shopping list.

ShoppingList/ShoppingList/ShoppingItem.cs
dbosco051394: It was an auto generated field I forgot to change. Good find!

ShoppingList/ShoppingList/ShoppingList.csproj
dbosco051394: Thanks, I forgot about this.

ShoppingList/ShoppingList/ShoppingItem.cs
mmwoodfo: Yeah, I figured! Visual Studio is very good at naming things haha. :)

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,57 @@
...
+
+            Assert.AreEqual(viewmodel.ShoppingList.First(), viewmodel.SelectedItem);
+        }
+
+        [TestMethod]
+        public void AddItemCommand_AddsItem()
+        {
+            var viewmodel = new MainWindowViewModel();
+            int expectedCount = viewmodel.ShoppingList.Count;
+            viewmodel.Text = "TestItem";
```
kenny2892: You could make this a DataMethod and use DataRows to try out different text inputs.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            SelectedItem = ShoppingList.First();
+        }
+
+        private void OnAddNewItem()
+        {
+            if(!string.IsNullOrWhiteSpace(Text))
+            {
+                ShoppingList.Add(new ShoppingItem(Text));
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(ShoppingList)));
```
Keboo: Because `ShoppingList` is an `ObservableCollection` you don't need to raise this event. It will handle raising the `INotifyCollectionChanged` event for you.

PR #323, opened 2019-12-08
Assignment 9

TatyanaHubbard:


ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -6,4 +6,15 @@
     <UseWPF>true</UseWPF>
```
mmwoodfo:
```suggestion
    <UseWPF>true</UseWPF>
    <NoWarn>AD0001</NoWarn>
```
You seem to have warnings set to be treated errors - I believe you can ignore this warning & your program should run. Alternatively, you can set treating warnings as errors to false.

TatyanaHubbard: Thank you! I wasn't sure if we were supposed to or not

dbosco051394:
On line 31 of your XAML        
<Button Content="Add Item" Command="{Binding AddItem}" CommandParameter="{Binding Text, ElementName=NewItem}" Grid.Row="2" Background="LightSlateGray"/>

You have ElementName set to 'NewItem' but there is no context for NewItem in your view model. Changing this to 'NewName' could help your issue with adding not working potentially.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,45 @@
...
+
+        public RelayCommand? AddItem { get; }
+
+        public MainWindowViewModel()
+        {
+            AddItem = new RelayCommand(OnAddItem);
+            _NewName = "";
+        }
+
+        private void OnAddItem()
```
matHopg: Your add item seems to only add items to the code behind but does not seem to update the list in your xaml am I mistaken here?

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,62 @@
...
+        
+        <ListBox x:Name="ShoppingListBox" 
+                 Grid.Row="0"
+                 Background="WhiteSmoke"
+                 ItemsSource="{Binding MyShoppingList}" 
+                 SelectedItem="{Binding SelectedShopItem}">
+            
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:ShopItem}">
+                    <StackPanel>
```
Keboo: `StackPanel` is not adding anything here. You can just have the TextBlock be there directly.

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,21 @@
     <OutputType>WinExe</OutputType>
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWPF>true</UseWPF>
+    <Nullable>enable</Nullable>
+    <NoWarn>AD0001</NoWarn>
   </PropertyGroup>
 
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|AnyCPU'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
```
Keboo: In general you don't want these in a `PropertyGroup` with a condition, and instead just put them inside of the default property group above. This type of thing usually happens when you turn these on from the project properties page inside of VS.

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Windows.Data;
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            if (value is ShopItem)
+            {
+                return "Visible";
```

PR #324, opened 2019-12-09
Assignment 9

CDobbins1:


ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,28 @@
...
+            <ColumnDefinition Width="125"/>
+        </Grid.ColumnDefinitions>
+        <Grid.RowDefinitions>
+            <RowDefinition />
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+        <ListBox x:Name="MyListBox" ItemsSource="{Binding Items}" Grid.Row="0" Grid.ColumnSpan="2" HorizontalContentAlignment="Stretch" Background="Gray">
+            <ListBox.ItemTemplate>
+                <DataTemplate>
+                    <StackPanel>
```
Keboo: The StackPanel is not adding anything and could be removed.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,27 @@
...
+
+namespace ShoppingList
+{
+    public class MainWindowViewModel : INotifyPropertyChanged
+    {
+        public event PropertyChangedEventHandler PropertyChanged;
+
+        public ObservableCollection<Item> Items { get; } = new ObservableCollection<Item>();
+
+        public string ItemName { get; set; } = "New Item";
```
Keboo: This property should raise the `PropertyChanged` event when it changes. It appears to "work" in this case only because this property is set before it is set as the MainWindow's DataContext.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,20 @@
...
+        public void AddItemCommand_AddsItem()
+        {
+            MainWindowViewModel sut = new MainWindowViewModel();
+            int initialCount = sut.Items.Count;
+
+            sut.AddItemCommand.Execute(null);
+
+            Assert.AreEqual<int>(initialCount + 1, sut.Items.Count);
+            Assert.AreEqual<string>("New Item", sut.Items[0].Name);
+        }
```
Keboo: This test is a bit confusing. It looks like it is setup to assert that new items are always inserted at the beginning of the collection (due to the `sut.Items[0]`). However that is not how the code works (which indicates that not enough cases were actually testes).

PR #325, opened 2019-12-10
Assignment9 snixon

samrnixon:


ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,59 @@
...
+using System.Collections.Generic;
+using System.Text;
+
+namespace ShoppingList.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void MainWindowViewModel_GoodShoppingItemAdded_CountSuccess()
```
Keboo: This test doesn't add much value since you are really just asserting that the `ObservableCollection` works. You don't need to test framework code.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -3,10 +3,70 @@
         xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+        
         xmlns:local="clr-namespace:ShoppingList"
-        mc:Ignorable="d"
-        Title="MainWindow" Height="450" Width="800">
-    <Grid>
+        mc:Ignorable="d" Topmost="True"
```

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -3,10 +3,70 @@
...
+        Title="MainWindow" Height="400" Width="400">
+    <Window.Resources>
+        <DataTemplate x:Key="ListTemplate">
+            <Grid>
+                <Grid.ColumnDefinitions>
+                    <ColumnDefinition Width="*"/>
+                </Grid.ColumnDefinitions>
+                <TextBlock Text="{Binding Path=ItemName}" Grid.Column="0"/>
+                <Button Command="{Binding ElementName=ShoppingItems, Path=DataContext.RemoveItemCommand}"
+                        CommandParameter="{Binding}"/>
```
Keboo: Nice use of command parameters!

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -3,10 +3,70 @@
...
+                <TextBox Text="{Binding SelectedItem.ItemName, UpdateSourceTrigger=PropertyChanged}"/>
+            </DockPanel>
+        </StackPanel>
+
+        <ListBox x:Name="ShoppingItemsList" ItemsSource="{Binding ShoppingItems}"
+            SelectedItem="{Binding SelectedItem}"
+                 Grid.Column="1" Grid.RowSpan="2" Background="Pink">
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:ShoppingListItem}">
+                    <StackPanel>
```
Keboo: The StackPanel is not adding much value here, and could be omitted.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,72 @@
...
+            AddItemCommand = new Command(OnAddItem);
+
+            ShoppingItems.Add(new ShoppingListItem("Milk"));
+            ShoppingItems.Add(new ShoppingListItem("Bread"));
+            ShoppingItems.Add(new ShoppingListItem("Eggs"));
+        }
+
+        private void OnAddItem()
+        {
+            if (!(_ItemAdd.Length == 0))
```
Keboo:
```suggestion
            if (_ItemAdd.Length != 0)
```

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if (value is ShoppingListItem)
+            {
+                return "Visible";
```

PR #326, opened 2019-12-10
Assignment 9 solution

markdietzler:
Editing list doesn't work, I will try to add it tomorrow.


ShoppingList/ShoppingList/ViewModel/VisibilityConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Text;
+using System.Windows.Data;
+
+namespace ShoppingList.ViewModel
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType, object? parameter, CultureInfo? culture)
+        {
+            return value is ShoppingListItemModel ? "Visible" : "Hidden";
```
Keboo: There is a `Visibility` enum that would have been better here. The default WPF type converters are the reason why passing a string works. The actual Visibility property uses the enum.

ShoppingList/ShoppingList/Views/MainWindow.xaml
```diff
@@ -0,0 +1,65 @@
...
+            <RowDefinition Height="20"/>
+        </Grid.RowDefinitions>
+        <StackPanel Grid.Row="2" Grid.Column="1">            
+            <TextBox Text="{Binding Item}" FontSize="16" TextAlignment="Center" Height="35"/>
+        </StackPanel>
+
+        <StackPanel Grid.Row="3" Grid.Column="1" Visibility="{Binding Model, Converter={StaticResource converter}}">
+            <TextBlock Text="Selected Item" FontSize="20" TextAlignment="Center" VerticalAlignment="Stretch" HorizontalAlignment="Stretch" Margin="5">
+                <TextBlock.TextDecorations>
+                    <TextDecoration/>
```
Keboo: Empty TextDecoration?

ShoppingList/ShoppingList/ViewModel/MainWindowViewModel.cs
```diff
@@ -0,0 +1,97 @@
...
+            if (!EqualityComparer<T>.Default.Equals(field, value)) //if using custom classes need to implement equals
+            {
+                field = value;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
+            }
+        }
+
+        private bool HasItemBoxChanged()
+        {
+            if (string.IsNullOrEmpty(Item)) {
```
Keboo: If you are going to check `string.IsNullOrWhiteSpace` below, this check doesn't add anything.

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -4,6 +4,14 @@
...
     <TargetFramework>netcoreapp3.0</TargetFramework>
     <UseWPF>true</UseWPF>
+    <Nullable>enable</Nullable>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="IntelliTect.Analyzers" Version="0.1.3" />
+    <PackageReference Include="MaterialDesignColors" Version="1.2.1" />
+    <PackageReference Include="MaterialDesignThemes" Version="3.0.0" />
+    <PackageReference Include="MvvmLightLibs" Version="5.4.1.1" />
```
Keboo: This is not the library that you want. This only works for .NET Framework, not .NET Core. You wanted MvvmLightLibsStd10

PR #327, opened 2019-12-10
Assignment9

jake-mcg:


ShoppingList/ShoppingList/ItemVisibilityConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Windows.Data;
+using System.Windows.Navigation;
+
+namespace ShoppingList
+{
+    public class ItemVisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type? targetType,  object? parameter, CultureInfo? culture)
+        {
+            return value is Item ? "Visible" : "Hidden";
```
AaronSauther: You will want to return a Visibility recognized by WPF, as well as use "collapsed" instead of "hidden" because hidden will simply make a blank space and not effectively "hide" your editing elements. I know this seems counterintuitive based on the wording of "hidden", had me stumped for awhile. 

```suggestion
            return value is Item ? Visibility.Visible : Visibility.Collapsed;
```

JerettLatimer:
Just an FYI, as of 8:30 AM on 12/10 your unit test check is failing.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        [TestMethod]
+        public void MainWindowViewModel_OnAddItem()
+        {
+            var sut = new MainWindowViewModel();
+
+            sut.AddItemCommand.Execute(new Item());
```
TatyanaHubbard: Your tests are failing because doing AddItemCommand(New Item) means that the Item is technically Null and will never return a count of 1.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+using System.Windows.Input;
+
+namespace ShoppingList
+{
+    public class MainWindowViewModel : INotifyPropertyChanged
+    {
+        private string _newItemText = "";
```
Keboo: Fields should be specified in _PascalCase, or switched to use an auto property

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,65 @@
...
+        private Item _selectedItem;
```
Keboo: Fields should be specified in _PascalCase, or switched to use an auto property

ShoppingList/ShoppingList/ItemVisibilityConverter.cs
Keboo: Agreed

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,47 @@
...
+            <RowDefinition Height="Auto"/>
+            <RowDefinition Height="*" />
+            <RowDefinition Height="Auto" />
+        </Grid.RowDefinitions>
+        <Grid.ColumnDefinitions>
+            <ColumnDefinition />
+            <ColumnDefinition Width="Auto" />
+        </Grid.ColumnDefinitions>
+        <StackPanel Grid.Row="0" Grid.Column="0" Width="200" HorizontalAlignment="Left">
+            <Button Content="Edit Item" Command="{Binding AddItemCommand}">
```
Keboo: It is a little un-usual to have two buttons on the same view, with different names, bound to the same command.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,47 @@
...
+            </Button>
+        </StackPanel>
+        <StackPanel Grid.Row="0" Grid.Column="0">
+            <TextBox Background="LemonChiffon" Width="500" HorizontalAlignment="Right" Text="{Binding SelectedItem.ItemName, UpdateSourceTrigger=PropertyChanged}">
```
Keboo: Setting explicit widths like this mean that it doesn't resize nicely. Try making your window bigger and smaller and you will see the text boxes don't stretch or shrink

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,47 @@
...
+            </TextBox>
+        </StackPanel>
+        <ListBox Background="AntiqueWhite" Margin="200,140,200,107" Grid.Row="1" ItemsSource="{Binding Items}" SelectedItem="{Binding SelectedItem}">
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:Item}">
+                    <TextBlock Text="{Binding ItemName}"/>
+                </DataTemplate>
+            </ListBox.ItemTemplate>
+        </ListBox>
+        <StackPanel Grid.Row="2" Grid.Column="0" HorizontalAlignment="Right">
```
Keboo: This `StackPanel` is not adding any value. You could remove it and set the same attributes on the `TextBox` directly.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,47 @@
...
+            <TextBox Background="LemonChiffon" Width="500" Text="{Binding NewItemText}">
+            </TextBox>
+        </StackPanel>
+        <StackPanel Grid.Row="2" Grid.Column="0" Width="200" HorizontalAlignment="Left">
```
Keboo: This `StackPanel` is not adding any value. You could remove it and set the same attributes on the `Button` directly.

PR #328, opened 2019-12-10
Assignment 9

kenny2892:


ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,44 @@
...
+            int newLength = viewModel.Items.Count;
+
+            Assert.AreNotEqual<int>(initialLength, newLength);
+        }
+
+        [DataTestMethod]
+        [DataRow("  ")]
+        [DataRow("")]
+        [DataRow("\n")]
+        public void AddItem_InvalidInput_Pass(string itemName)
```
dougdoner: Good use of DataRow to test different types of empty strings.

ShoppingList/ShoppingList/Item.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace ShoppingList
+{
+    public class Item
+    {
+        public string Name { get; set; }
```
dougdoner: Someone else mentioned it on my pull request, but in the future you could add something like a quantity property (not required).

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
```
dougdoner: Since you are allowing null parameters, you might want to check for null parameters in the method.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,49 @@
...
+        </StackPanel>
+
+        <StackPanel Grid.Row="1" Visibility="{Binding SelectedItem, Converter={StaticResource converter}}">
+            <TextBlock Text="Selected Item:" FontSize="20" TextAlignment="Center" VerticalAlignment="Stretch" HorizontalAlignment="Stretch">
+                <TextBlock.TextDecorations>
+                    <TextDecoration/>
+                </TextBlock.TextDecorations>
+            </TextBlock>
+            <TextBlock Text="Name:" FontSize="16" Margin="10,0,0,0" />
+            <TextBox Text="{Binding SelectedItem.Name}" FontSize="16" TextAlignment="Center" Height="35"/>
```
mmwoodfo:
```suggestion
            <TextBox Text="{Binding SelectedItem.Name, UpdateSourceTrigger=PropertyChanged}" FontSize="16" TextAlignment="Center" Height="35"/>
```
To update the text live

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,48 @@
...
+        d:DataContext="{d:DesignInstance Type=local:MainWindowViewModel}"
+        TextElement.Foreground="{DynamicResource MaterialDesignBody}"
+        Background="{DynamicResource MaterialDesignPaper}"
         Title="MainWindow" Height="450" Width="800">
+
+    <Window.Resources>
+        <local:VisibilityConverter x:Key="converter"/>
+
+        <DataTemplate DataType="{x:Type local:Item}" x:Key="ItemTemplate">
+            <StackPanel>
```
Keboo: The `StackPanel` is not adding anything here and could be removed.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,44 @@
...
+        public RelayCommand AddItemCommand { get; }
+
+        public MainWindowViewModel()
+        {
+            AddItemCommand = new RelayCommand(OnAddItem);
+        }
+
+        private void OnAddItem()
+        {
+            if (!string.IsNullOrWhiteSpace(NewItemName) && !string.IsNullOrEmpty(NewItemName))
```
Keboo: Since you already checked `string.IsNullOrWhiteSpace` the `string.IsNullOrEmpty` is meaningless and can be removed.
```suggestion
            if (!string.IsNullOrWhiteSpace(NewItemName))
```

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+        {
+            return value is Item ? "Visible" : "Hidden";
```

PR #330, opened 2019-12-10
Assignment 9

crackerJackTech:
…tions have been met except the extra credit.

### DIFF
--- a/ShoppingList.Tests/MainWindowViewModelTests.cs
+++ b/ShoppingList.Tests/MainWindowViewModelTests.cs
@@ -1,0 +1,79 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Text;
+using System.Windows.Input;
+
+namespace ShoppingList.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        private void OnTestMethod() { }
+        
+        
+        [TestMethod]
+        public void Constructor_MainWindowViewModelContainsCommands_Success()
+        {
+            //Arrange
+            var testViewModel = new MainWindowViewModel();
+            //Act
+            Command AddListItemCommand = new Command(OnTestMethod);
+            Command RemoveListItemCommand = new Command(OnTestMethod);
+            ICommand ChangeListItemNameCommand = new Command(OnTestMethod);
+            //Assert
+            Assert.AreEqual<Type>(AddListItemCommand.GetType(), testViewModel.AddListItemCommand.GetType());
+            Assert.AreEqual<Type>(RemoveListItemCommand.GetType(), testViewModel.RemoveListItemCommand.GetType());
+            Assert.AreEqual<Type>(ChangeListItemNameCommand.GetType(), testViewModel.ChangeListItemNameCommand.GetType());
+        }
+
+        [TestMethod]
+        public void ShoppingList_ListContainsItems_Success()
+        {
+            //Arrange
+            var testList = new ObservableCollection<ShoppingListItem>()
+            {
+                new ShoppingListItem("testItem1"),
+                new ShoppingListItem("testItem2")
+            };
+            var testViewModel = new MainWindowViewModel();
+            var viewModelList = testViewModel.ShoppingList;
+            //Act
+
+            //Assert
+            Assert.AreEqual<int>(testList.Count, testViewModel.ShoppingList.Count);
+            Assert.AreEqual<Type>(testList.GetType(), viewModelList.GetType());
+        }
+
+        [TestMethod]
+        public void OnAddListItem_ItemIsAddedToListFromExecute_Fails()
+        {
+            //Arrange
+            var testViewModel = new MainWindowViewModel();
+            var expectedItem = new ShoppingListItem("Item3");
+            var viewModelList = testViewModel.ShoppingList;
+
+            //Act
+            testViewModel.AddListItemCommand.Execute(expectedItem);
+
+            //Assert
+            Assert.IsFalse(viewModelList.Contains(expectedItem));
+        }
+
+        [TestMethod]
+        public void OnRemoveListItem_AttemptRemovalOfFirstItem_Fails()
+        {
+            //Arrange
+            var testViewModel = new MainWindowViewModel();
+            var expectedItem = new ShoppingListItem("Item1");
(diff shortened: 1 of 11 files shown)


ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,79 @@
...
+namespace ShoppingList.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        private void OnTestMethod() { }
+        
+        
+        [TestMethod]
+        public void Constructor_MainWindowViewModelContainsCommands_Success()
```
Keboo: Not sure this test adds much value.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,79 @@
...
+                new ShoppingListItem("testItem1"),
+                new ShoppingListItem("testItem2")
+            };
+            var testViewModel = new MainWindowViewModel();
+            var viewModelList = testViewModel.ShoppingList;
+            //Act
+
+            //Assert
+            Assert.AreEqual<int>(testList.Count, testViewModel.ShoppingList.Count);
+            Assert.AreEqual<Type>(testList.GetType(), viewModelList.GetType());
```
Keboo: Why test for this?

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,61 @@
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
-        mc:Ignorable="d"
+        mc:Ignorable="d" Topmost="True"
```

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,61 @@
...
+        </Grid.ColumnDefinitions>
+        <Grid.RowDefinitions>
+            <RowDefinition Height="Auto"/>
+            <RowDefinition />
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+
+        <StackPanel Grid.Row="1" Visibility="{Binding SelectedListItem, Converter={StaticResource converter}}">
+
+            <TextBox HorizontalAlignment="Left" Width="200" FontSize="20"  Text="{Binding SelectedListItem.ListItemName}"/>
```
Keboo: You don't want to set an explicit Width on these. By doing so it means that they wont scale as the window is resized.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,61 @@
...
+            <Button HorizontalAlignment="Left" Width="200" Height="50" FontSize="20" Content="Edit Selected Item" Command="{Binding ChangeListItemNameCommand}"/>
+
+        </StackPanel>
+
+        <TextBlock Grid.Column="0" Grid.Row="2" TextWrapping="Wrap">
+            <TextBlock.Text>
+                <MultiBinding StringFormat="Hi {0}">
```
Keboo: Since this only has a single binding you don't actually need the multibinding and can just put the StringFormat on the binding directly.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,61 @@
...
+                    <Binding Path="SelectedListItem.ListItemName" />
+                </MultiBinding>
+            </TextBlock.Text>
+        </TextBlock>
+
+        <ListBox ItemsSource="{Binding ShoppingList}" FontSize="20" FontWeight="Bold" Grid.Column="1" Grid.RowSpan="2" x:Name="ShoppingList"
+                 SelectedItem="{Binding SelectedListItem}">
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:ShoppingListItem}">
+                    <StackPanel>
```
Keboo: The StackPanel is not really adding any value and could be removed.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,61 @@
...
+                    </StackPanel>
+                </DataTemplate>
+            </ListBox.ItemTemplate>
+        </ListBox>
+
+
+
+
+        <StackPanel Grid.Row="2" Grid.Column="0">
+            <ToggleButton Height="50" FontSize="20" Content="Remove Selected Item on List" Command="{Binding RemoveListItemCommand}" />
```
Keboo: Is there a reason this is ToggleButton and not just a Button?

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,84 @@
...
+        {
+            get => _SelectedListItem;
+            set => SetProperty(ref _SelectedListItem, value);
+        }
+
+        public Command AddListItemCommand { get; }
+        public Command RemoveListItemCommand { get; }
+        public ICommand ChangeListItemNameCommand { get; }
+
+        private bool CanExecute { get; set; }
```
Keboo: This does not appear to get accessed for anything.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,84 @@
...
+        private void OnAddListItem()
+        {
+            ShoppingList.Add(new ShoppingListItem($"Item{ShoppingList.Count + 1}"));
+            CanExecute = true;
+            AddListItemCommand.InvokeCanExecuteChanged();
+        }
+
+        private void OnRemoveListItem()
+        {
+            if(!(SelectedListItem is null))
```
Keboo:
```suggestion
            if(SelectedListItem is { })
```

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+
+            if(value is null)
+            {
+                value = Visibility.Collapsed;
+            }
+
+            return value;
```
Keboo: You probably want to return a different Visibility value here. This will likely just cause a hidden type conversion failure.

PR #331, opened 2019-12-10
Assignment9

twofingerrightclick:


ShoppingList/ShoppingList/DateAddedConverter.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    class DateAddedConverter : IValueConverter
```
Keboo: DateAddedConverter is an internal class that is apparently never instantiated. If so, remove the code from the assembly. If this class is intended to contain only static members, make it static (Shared in Visual Basic).

ShoppingList/ShoppingList/ShoppingItemInvisibleConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+{
+    public class ShoppingItemInvisibleConverter : BaseConverter, IValueConverter
+
+    {
+
+        public object Convert(object value, Type? targetType =null , object? parameter = null, CultureInfo? culture = null)
+        {
+            if (value is ShoppingItem && value!=null)
+            {
+                return "Visible";
```
Keboo: There is a `Visibility` enum that would have been better here. The default WPF type converters are the reason why passing a string works. The actual Visibility property uses the enum.

ShoppingList/ShoppingList/ShoppingListMainWindowViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace ShoppingList
+{
+    public class ShoppingListMainWindowViewModel : ViewModelBase
+    {
+
+        public string Title { get;} = "Shopping List";
+
+        private string _NewEntryText = "Enter item...";
+        public string NewEntryText { get => _NewEntryText; set{Set(ref _NewEntryText, value);
+                RaisePropertyChanged();
```
Keboo: The Set method will raise the PropertyChanged event. You don't need to do it here.

ShoppingList/ShoppingList/ShoppingListMainWindowViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+}
+
+       
+
+        private ShoppingItem? _SelectedShoppingItem;
+        public ShoppingItem? SelectedShoppingItem
+        {
+            get => _SelectedShoppingItem;
+            set { Set(ref _SelectedShoppingItem, value);
+                RaisePropertyChanged();
```
Keboo: The Set method will raise the PropertyChanged event. You don't need to do it here.

ShoppingList/ShoppingList/ShoppingListMainWindowViewModel.cs
```diff
@@ -0,0 +1,94 @@
...
+            }
+        } 
+
+        public ObservableCollection<ShoppingItem> ShoppingListItems { get; } = new ObservableCollection<ShoppingItem>();
+
+  
+
+        private bool CanExecute { get; set; }
```
Keboo: This doesn't appear to be accessed for anything.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -5,8 +5,82 @@
...
+        <local:ShoppingItemInvisibleConverter x:Key="ItemInvisibleConverter"/>
+
+
+        <DataTemplate DataType="{x:Type local:ShoppingItem}" x:Key="ShoppingItemTemplate">
+            <StackPanel>
+                <TextBlock FontSize="15" >
+                    <TextBlock.Text>
+                        <MultiBinding StringFormat="{}{0} Added: {1}">
+                            <Binding Path="Name" />
+                            <Binding Converter="{StaticResource DateAddedConverter}" />
```
Keboo: You have a `TimeAdded` property on your ShoppingItem but don't use it?

PR #332, opened 2019-12-10
Assignment 9

AustinHall7798:
Assignment 9 finished. Both Extra credit completed


ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -1,12 +1,113 @@
...
+
+        <TextBox Text="{Binding NewItem}"
+                 AcceptsReturn="True"
+                 x:Name="ShoppingListTextBox"
+                 Grid.Row="2"
+                 Grid.Column="0"
+                 Height="30" 
+                 FontSize="17" 
+                 Grid.ColumnSpan="2" Background="White">
+        </TextBox>
```
jacob-berger: Much easier to read a layout object's attributes when they're on separate lines like this.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        public ICommand AddItem { get; }
+        public ICommand DeleteItem { get; }
+        public ICommand EditItem { get; }
+
+        public MainWindowViewModel()
+        {
+            AddItem = new RelayCommand(OnAddItem);
+            DeleteItem = new RelayCommand(OnDeleteItem);
+            EditItem = new RelayCommand(OnEditItem);
```
jacob-berger: You have a command for editing an item, but it seems it's never actually used other than just deselecting the item?

AustinHall7798: That's to hide the editing interface after editing the selected item since the visibility of the editor is bound to the selected item. There might be a better way to do it but I'm new to this XAML thing

ShoppingList/ShoppingList/MainWindow.xaml
GamerBah: Going to give the same praise as above. I can't understand why anything but chop-down property lists is what IDE's ship with. Nice job.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,74 @@
...
+            MainWindowViewModel viewModel = new MainWindowViewModel
+            {
+                NewItem = "apple"
+            };
+            viewModel.AddItem.Execute(null);
+             Assert.AreEqual(viewModel.ItemsList.ElementAt(0).Name, "~apple");
+        }
+
+        [DataRow(" ")]
+        [DataRow(null)]
```
JerettLatimer: Good use of data driven tests!

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -1,12 +1,112 @@
...
+                 Background="Turquoise"
+                 SelectionMode="Single" 
+                 Foreground="#FFC75050" 
+                 FontFamily="Impact" 
+                 Grid.ColumnSpan="2" 
+                 FontSize="20" BorderThickness="2,2,2,2" >
+
+            <ListBox.ItemTemplate>
+                <DataTemplate>
+                    <StackPanel Orientation="Horizontal">
```
Keboo: The StackPanel is not adding much here and could be omitted.

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,21 @@
...
+using System.Text;
+using System.Windows.Data;
+
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object? value, Type? targetType = null, object? parameter = null, CultureInfo? culture = null)
+        {
+            return value is ShoppingItem ? "Visible" : "Hidden";
```

PR #333, opened 2019-12-10
Assignment 9

GamerBah:
While not quite being as easy as UWP, was still enjoyable :)

Something to note is that I wasn't quite sure if my `Remove Item` feature fit with MVVM.
Is it possible to test the actual Views, if we had to? I'm guessing the reason we didn't have to was that it would've been too complicated?

### DIFF
--- a/ShoppingList.Tests/GlobalSuppressions.cs
+++ b/ShoppingList.Tests/GlobalSuppressions.cs
@@ -1,0 +1,58 @@
+﻿// This file is used by Code Analysis to maintain SuppressMessage
+// attributes that are applied to this project.
+// Project-level suppressions either have no target or are given
+// a specific target and scoped to a namespace, type, member, etc.
+
+using System.Diagnostics.CodeAnalysis;
+
+[assembly:
+    SuppressMessage("Naming",
+        "CA1707:Identifiers should not contain underscores",
+        Justification = "<Pending>",
+        Scope         = "member",
+        Target        = "~M:ShoppingList.Tests.MainWindowViewModelTests.AddItem_GivenCorrectInput_AddsToList")]
+[assembly:
+    SuppressMessage("Naming",
+        "CA1707:Identifiers should not contain underscores",
+        Justification = "<Pending>",
+        Scope         = "member",
+        Target =
+            "~M:ShoppingList.Tests.MainWindowViewModelTests.AddItem_GivenInvalidInput_ListRetainsOriginalSize(System.String)")]
+[assembly:
+    SuppressMessage("Naming",
+        "CA1707:Identifiers should not contain underscores",
+        Justification = "<Pending>",
+        Scope         = "member",
+        Target =
+            "~M:ShoppingList.Tests.MainWindowViewModelTests.SelectedItem_CorrectlyUpdatesRelatedProperties")]
+
+[assembly:
+    SuppressMessage("Naming",
+        "CA1707:Identifiers should not contain underscores",
+        Justification = "<Pending>",
+        Scope         = "member",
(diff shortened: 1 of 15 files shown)


ShoppingList.Tests/GlobalSuppressions.cs
```diff
@@ -0,0 +1,58 @@
...
+        Justification = "<Pending>",
+        Scope         = "member",
+        Target        = "~M:ShoppingList.Tests.MainWindowViewModelTests.AddItem_GivenCorrectInput_AddsToList")]
+[assembly:
+    SuppressMessage("Naming",
+        "CA1707:Identifiers should not contain underscores",
+        Justification = "<Pending>",
+        Scope         = "member",
+        Target =
+            "~M:ShoppingList.Tests.MainWindowViewModelTests.AddItem_GivenInvalidInput_ListRetainsOriginalSize(System.String)")]
```
Keboo: Use keyword instead of type name

ShoppingList/ShoppingList/App.xaml
```diff
@@ -4,6 +4,17 @@
...
+            <ResourceDictionary.MergedDictionaries>
+                <ResourceDictionary
+                    Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Dark.xaml" />
+                <ResourceDictionary
+                    Source="pack://application:,,,/MaterialDesignThemes.Wpf;component/Themes/MaterialDesignTheme.Defaults.xaml" />
+                <ResourceDictionary
+                    Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Primary/MaterialDesignColor.DeepOrange.xaml" />
+                <ResourceDictionary
+                    Source="pack://application:,,,/MaterialDesignColors;component/Themes/Recommended/Accent/MaterialDesignColor.Amber.xaml" />
+            </ResourceDictionary.MergedDictionaries>
```
Keboo: This works but in the 3.0.0 release there is an easier way to declare the colors. https://github.com/MaterialDesignInXAML/MaterialDesignInXamlToolkit/wiki/Getting-Started#version-300-and-later

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,145 @@
...
+                                              Panel.ZIndex="1"
+                                              materialDesign:ShadowAssist.ShadowDepth="Depth2">
+                        <Grid>
+                            <Grid.ColumnDefinitions>
+                                <ColumnDefinition Width="8*" />
+                                <ColumnDefinition Width="2*" />
+                                <ColumnDefinition Width="2*" />
+                            </Grid.ColumnDefinitions>
+
+                            <Viewbox Margin="25">
```
Keboo: A Viewbox is probably not the container that you want here. A generic holder is usual a Border. You could also just omit the Viewbox and apply the Margin directly to the TextBox

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        public Item? SelectedItem
+        {
+            get => _SelectedItem;
+            set
+            {
+                _SelectedItem = value;
+                NewName       = _SelectedItem?.Name ?? "";
+                NewAmount     = _SelectedItem?.Amount - 1 ?? 0;
+                PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(nameof(SelectedItem)));
```
Keboo: This works, but will fire PropertyChanged even if the value has not changed.

PR #335, opened 2019-12-10
Assignment9 2

Tpirtle:
I did the Extra credit

### DIFF
--- a/ShoppingList.Tests/MainWindowViewModelTests.cs
+++ b/ShoppingList.Tests/MainWindowViewModelTests.cs
@@ -1,0 +1,66 @@
+﻿using Microsoft.VisualStudio.TestTools.UnitTesting;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+
+namespace ShoppingList.Tests
+{
+    [TestClass]
+    public class MainWindowViewModelTests
+    {
+        private MainWindowViewModel ViewModel { get; set; }
+        [TestInitialize]
+        public void SetUp()
+        {
+            ViewModel = new MainWindowViewModel();
+        }
+        [TestMethod]
+        public void Constructor_SetsDefaultSelectedPerson()
+        {
+            Assert.AreEqual(ViewModel.Items.First(), ViewModel.SelectedItem);
+        }
+        [TestMethod]
+        public void ValueConverter_IsVisible()
+        {
+            Item item = new Item("Apple");
+            var converter = new VisibilityConverter();
+
+            string visibility = (string)converter.Convert(item,null,null,null);
+
+            Assert.AreEqual("Visible",visibility);
+        }
+
+        [TestMethod]
+        public void AddPersonCommand_AddsPerson()
+        {
+            int initialCount = ViewModel.Items.Count;
+            ViewModel.NewItem = "Toast";
+
+            ViewModel.AddItemCommand.Execute(null);
+
+            Assert.AreEqual(initialCount + 1, ViewModel.Items.Count);
+        }
+
+        [TestMethod]
+        public void AddPersonCommand_NotNull()
+        {
+            int initialCount = ViewModel.Items.Count;
+
+            ViewModel.AddItemCommand.Execute("");
+
+            Assert.AreEqual(initialCount, ViewModel.Items.Count);
+        }
+        [TestMethod]
+        public void ChangeNameCommand_NameIsChanged()
+        {
+            string initialName = ViewModel.SelectedItem.Name;
+            ViewModel.SelectedItem.Name = "Chee";
+
+            ViewModel.ChangeNameCommand.Execute(null);
+            ViewModel.SelectedItem = ViewModel.Items.First();
+
+            Assert.AreNotEqual(initialName, ViewModel.SelectedItem.Name);
+        }
+    }
+}
--- a/ShoppingList/ShoppingList/MainWindow.xaml
+++ b/ShoppingList/ShoppingList/MainWindow.xaml
@@ -6,7 +6,59 @@
         xmlns:local="clr-namespace:ShoppingList"
         mc:Ignorable="d"
         Title="MainWindow" Height="450" Width="800">
-    <Grid>
+    <Window.Resources>
+        <local:VisibilityConverter x:Key="VisibilityConverter"/>
 
+        <Style x:Key="EditTextStyle" TargetType="TextBlock">
+            <Setter Property="FontStyle" Value="Italic"/>
+        </Style>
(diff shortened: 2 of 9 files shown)


ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,66 @@
...
+    public class MainWindowViewModelTests
+    {
+        private MainWindowViewModel ViewModel { get; set; }
+        [TestInitialize]
+        public void SetUp()
+        {
+            ViewModel = new MainWindowViewModel();
+        }
+        [TestMethod]
+        public void Constructor_SetsDefaultSelectedPerson()
```
NikolasGraham: Should "Person" be "Item"?

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            Item item = new Item("Apple");
+            var converter = new VisibilityConverter();
+
+            string visibility = (string)converter.Convert(item,null,null,null);
+
+            Assert.AreEqual("Visible",visibility);
+        }
+
+        [TestMethod]
+        public void AddPersonCommand_AddsPerson()
```
NikolasGraham: Name should be `AddItemCommand_AddsItem()`

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,66 @@
...
+            int initialCount = ViewModel.Items.Count;
+            ViewModel.NewItem = "Toast";
+
+            ViewModel.AddItemCommand.Execute(null);
+
+            Assert.AreEqual(initialCount + 1, ViewModel.Items.Count);
+        }
+
+        [TestMethod]
+        public void AddPersonCommand_NotNull()
```
NikolasGraham: Name should be item here again.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+                CanExecute = true;
+                ChangeNameCommand.RaiseCanExecuteChanged();
+            }
+        }
+        private void OnChangeName()
+        {
+            SelectedItem = null;
+        }
+    }
+}
```
NikolasGraham: Technically not in the requirements, but just as a shopping list I feel like it's not complete without a way to cross off items or delete them.

ShoppingList.Tests/MainWindowViewModelTests.cs
```diff
@@ -0,0 +1,66 @@
...
+        {
+            string initialName = ViewModel.SelectedItem.Name;
+            ViewModel.SelectedItem.Name = "Chee";
+
+            ViewModel.ChangeNameCommand.Execute(null);
+            ViewModel.SelectedItem = ViewModel.Items.First();
+
+            Assert.AreNotEqual(initialName, ViewModel.SelectedItem.Name);
+        }
+    }
```
NikolasGraham: No tests for selected item.

ShoppingList.Tests/MainWindowViewModelTests.cs
Tpirtle: Yes, these should be Item, I referenced the Tests that Kevin did in class, and wrote down Person on the names instead of Item.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+            AddItemCommand = new Command(OnAddItem);
+
+            Items.Add(new Item("Cheese"));
+
+            SelectedItem = Items.First();
+        }
+
+        private void OnAddItem()
+        {
+            if (NewItem != "" && NewItem != null)
```
Keboo: Test for empty strings using 'string.Length' property or 'string.IsNullOrEmpty' method instead of an Equality check.

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Collections.ObjectModel;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Windows.Input;
+
+namespace ShoppingList
+{
+    public class MainWindowViewModel : ViewModelBase, INotifyPropertyChanged
```
Keboo: `ViewModelBase` implements `INotifyPropertyChanged` so you don't need it here.
```suggestion
    public class MainWindowViewModel : ViewModelBase
```

ShoppingList/ShoppingList/MainWindowViewModel.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        public event PropertyChangedEventHandler PropertyChanged;
+        private void SetProperty<T>(ref T field, T value, [CallerMemberName]string propertyName=null)
```
Keboo: There is a `Set` method on `ViewModelBase`. You no longer need this implementation.

ShoppingList/ShoppingList/VisibilityConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ShoppingList
+{
+    public class VisibilityConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, CultureInfo culture)
+        {
+            if(value is Item)
+            {
+                return "Visible";
```

ShoppingList/ShoppingList/ShoppingList.csproj
```diff
@@ -6,4 +6,10 @@
     <UseWPF>true</UseWPF>
   </PropertyGroup>
 
+  <ItemGroup>
+    <PackageReference Include="MaterialDesignThemes" Version="3.0.0" />
```
Keboo: This is reference but it is not implemented. Your app doesn't actually have any of the material design theming.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -6,7 +6,59 @@
...
+        <ListBox x:Name="ItemList"
+                 Grid.Column="0"
+                 Background="Lime"
+                 Height="450"
+                 Width="500"
+                 ItemsSource="{Binding Items}"
+                 SelectedItem="{Binding SelectedItem}">
+            <ListBox.ItemTemplate>
+                <DataTemplate>
+                    <StackPanel Orientation="Vertical">
```
Keboo: The StackPanel is not adding much value here and can be omitted.

PR #336, opened 2019-12-10
Assignment9 

skozubenko1:
still working on testing

### DIFF
--- a/ShoppingList/ShoppingList/MainWindow.xaml.cs
+++ b/ShoppingList/ShoppingList/MainWindow.xaml.cs
@@ -1,7 +1,4 @@
-﻿using ShoppingList.Models;
-using ShoppingList.ViewModels;
-using System.Collections.Generic;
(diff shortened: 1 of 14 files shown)


ShoppingList/ShoppingList/MainWindow.xaml.cs
```diff
@@ -7,9 +8,40 @@ namespace ShoppingList
...
+        }
+
+        private void Grid_SelectionChanged(object sender, System.Windows.Controls.SelectionChangedEventArgs e)
+        {
+            buttonEdit.IsEnabled 
+                = buttonRemove.IsEnabled 
+                = grid.SelectedItems.Count > 0;
+        }
+
+        private void buttonRemove_Click(object sender, RoutedEventArgs e)
```
Keboo: This should have been a second Command on your view model rather than using the event and calling into the view model.

PR #337, opened 2019-12-10
assignment 9

whereisdaniru:


ShoppingList/ShoppingList/Item.cs
```diff
@@ -0,0 +1,25 @@
...
+
+namespace ShoppingList
+{
+    public class Item
+    {
+        public string ItemName { get; set; }
+
+        public Item(string itemName)
+        {
+            if (itemName is null) throw new ArgumentNullException(nameof(ItemName));
```
Keboo: The parameter passed to this exception's constructor should be the parameter name; not a message

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,60 @@
...
+            <RowDefinition Height="Auto"/>
+        </Grid.RowDefinitions>
+
+        <StackPanel Grid.Column="0" Grid.Row="2" Visibility="{Binding SelectedItem, Converter={StaticResource converter}}">
+            <TextBox VerticalAlignment="Center" HorizontalAlignment="Center" Width="300" FontSize="20"  Text="{Binding SelectedItem.ItemName}"/>
+            <Button VerticalAlignment="Center" HorizontalAlignment="Center" Height="auto" Width ="300" FontSize="20" Content="Edit Selected Item" Command="{Binding ChangeItemCommand}"/>
+        </StackPanel>
+
+        <StackPanel Grid.Column="0" Grid.Row="4">
+            <ToggleButton Height="auto" FontSize="20" Content="Remove Selected Item From List" Command="{Binding RemoveItemCommand}" />
```
Keboo: Does this need to be a ToggleButton or would a normal Button suffice.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,60 @@
...
+
+        <TextBlock Grid.Column="1" Grid.Row="1" FontSize="20" FontWeight="Heavy" FontStyle="Oblique" TextAlignment="Center" Background="PaleGoldenrod" Width="300">
+            Shopping List
+        </TextBlock>
+
+        <ListBox ItemsSource="{Binding ShoppingList}" FontSize="20" FontStyle="Oblique" HorizontalContentAlignment="Center" Background="PaleGoldenrod" Width="300" Grid.Column="1" Grid.Row="2" x:Name="ShoppingList"
+                 SelectedItem="{Binding SelectedItem}">
+            <ListBox.ItemTemplate>
+                <DataTemplate DataType="{x:Type local:Item}">
+                    <StackPanel>
```
Keboo: The StackPanel is not adding much here and could be removed.

ShoppingList/ShoppingList/MainWindow.xaml
```diff
@@ -4,9 +4,60 @@
         xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
         xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
         xmlns:local="clr-namespace:ShoppingList"
-        mc:Ignorable="d"
+        mc:Ignorable="d" Topmost="True"
```

